Issue #23, opened 2023-07-04
Undo/redo history


chrisvxd:
* Add Undo/Redo buttons to header using react-feather and IconButton component
* Avoid storing entire state
* Include keyboard shortcuts, if possible

chrisvxd:
Self assigning to prevent anyone picking up until #37 is complete.

CC @yuddomack

chrisvxd:
Hey @Yuddomack, #37 is merged so feel free to have another go at this!

chrisvxd:
Available in `0.9.0-canary.222697e` via #111

PR #49, opened 2023-09-07
undo redo history

Yuddomack:
Hi! I came across this project while researching for easy page development and interested in this project.
thank you for creating a good project.

I looked at #23 and made a simple draft. (using use-undo)
I think basic operations are possible. but as you know, It is not advisable to record all 'Data' for every change.

so I think it could be a good idea to define some actions using the command pattern.

```javascript
// pseudo
function Puck(){
  const [currentData, setCurrentData) = useState(initialData)
  const [commandHistory, setCommandHistory = setState([]) // or { [id]: { beforeAction: id, afterAction: id, ... } }

  // commands (it can make the related functions more abstract.)
  const changeField = (id, field) => {
    // todo
  }
  const addBlock = () => {}
  const removeBlock = () => {}
  // ...
  const undo = (id) => {
    commandHistory[id].inverse()
    setSelected(id)
    focus(field)
  }

  // ...

  <InputOrGroup onBlur={changeField} /> // for record once, not every change
}
```

How about your plan?
If the priority is not high and there is no one in charge of this feature, I can try this

### DIFF
--- a/apps/demo/app/Framework.ts
+++ b/apps/demo/app/Framework.ts
@@ -1,2 +1,2 @@
-export const validFrameworks = ["antd", "material-ui", "custom"];
-export type Framework = "antd" | "material-ui" | "custom";
+export const validFrameworks = ["antd", "material-ui", "custom"] as const;
+export type Framework = (typeof validFrameworks)[number];
--- a/packages/core/components/Puck/index.tsx
+++ b/packages/core/components/Puck/index.tsx
@@ -8,6 +8,7 @@ import {
   useState,
 } from "react";
 import { DragDropContext } from "react-beautiful-dnd";
+import useUndo from "use-undo";
 import DroppableStrictMode from "../DroppableStrictMode";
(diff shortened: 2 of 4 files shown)


chrisvxd (commented):
Hi @Yuddomack, thanks for this! The behaviour works really well.

We're currently refactoring how we write state as part of #37, which introduces a reducer. Once this is merged, it might be possible to find a way to handle undo/redo without storing the entire `Data` object in memory. The reducer is currently [here](https://github.com/measuredco/puck/pull/37/files#diff-52f2c548804b8f08ed5f2e129cd332eb261706c0b3ff913d78f0912c58e4f29eR75), but the code is WIP and subject to change.

I also noticed this code doesn't work for undoing the insertion of new components. This could be a React `key` issue rather than a data issue, and might also change with #37.

I think I'd rather wait until we can rewind specific mutations than store the entire object in memory. Would you be up for giving this another go once #37 lands?

PR #53, opened 2023-09-08
docs: replace quotes API in demo with hardcoded quotes

chrisvxd:
The [quotes API](https://github.com/lukePeavey/quotable/tree/master) we were using in the demo went down. Unsure if related to our sudden surge of traffic. If it is, apologies to maintainer @lukePeavey.

This PR replaces the API call with some hardcoded quotes.


monospaced (approved):
LGTM

PR #58, opened 2023-09-09
Add support for custom fields

### LINKED ISSUE #57: Custom field type  and custom field input
How can I create a new field type and it's editor?

The current plugin system doesn't seem to support this, only allowing us to reimplement the whole sidebar editing?

chrisvxd:
This PR proposes an implementation for custom field support, allowing users to extend the included field types.

* Adds new `custom` option to field `type`
* Adds new `render` param, accepting the following options:
  - **field** (`Field`): Field configuration
  - **name** (`string`): Name of the field
  - **value** (`any`): Value for the field
  - **onChange** (`(value: any) => void`): Callback to change the value
  - **readOnly** (`boolean` | `undefined`): Whether or not the field should be in readOnly mode
* Exposes `<FieldLabel>` component which accepts the `children`, `label` and `icon` props (currently undocumented)

If we like this direction, suggest we mark as Ready for Review.

Closes #57 

## Example

```tsx
import { FieldLabel } from "@measured/puck";

export const MyComponent: ComponentConfig = {
  fields: {
    myField: {
      type: "custom",
      render: ({ field, name, onChange, value }) => {
        return (
          <FieldLabel label={field.label || name}>
            <input
              placeholder="Enter text..."
              type="text"
              name={name}
              defaultValue={value}
            ></input>
          </FieldLabel>
        );
      },
    },
  },
};
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -110,6 +110,37 @@ const myPlugin = {
 };
 ```
 
+## Custom fields
+
+Puck supports custom fields using the `custom` field type and `render` method.
+
+In this example, we optionally add the `<FieldLabel>` component to add a label:
+
+```tsx
+import { FieldLabel } from "@measured/puck";
+
+export const MyComponent: ComponentConfig = {
+  fields: {
+    myField: {
+      type: "custom",
+      render: ({ field, name, onChange, value }) => {
+        return (
+          <FieldLabel label={field.label || name}>
+            <input
+              placeholder="Enter text..."
+              type="text"
+              name={name}
+              defaultValue={value}
+              onChange={(e) => onChange(e.currentTarget.value)}
+            ></input>
+          </FieldLabel>
+        );
+      },
+    },
+  },
+};
+```
+
 ## Reference
 
 ### `<Puck>`
@@ -152,7 +183,7 @@ The `Config` object describes which components Puck should render, how they shou
 
 A `Field` represents a user input field shown in the Puck interface.
 
-- **type** (`text` | `textarea` | `number` | `select` | `radio` | `external` | `array`): The input type to render
+- **type** (`text` | `textarea` | `number` | `select` | `radio` | `external` | `array` | `custom`): The input type to render
 - **label** (`text` [optional]): A label for the input. Will use the key if not provided.
 - **arrayFields** (`object`): Object describing sub-fields for items in an `array` input
   - **[fieldName]** (`Field`): The Field objects describing the input data for each item
@@ -163,6 +194,12 @@ A `Field` represents a user input field shown in the Puck interface.
   - **value** (`string` | `number` | `boolean`)
 - **adaptor** (`Adaptor`): Content adaptor if using the `external` input type
 - **adaptorParams** (`object`): Paramaters passed to the adaptor
+- **render** (`Component`): Render a custom field. Receives the props:
+  - **field** (`Field`): Field configuration
+  - **name** (`string`): Name of the field
+  - **value** (`any`): Value for the field
+  - **onChange** (`(value: any) => void`): Callback to change the value
+  - **readOnly** (`boolean` | `undefined`): Whether or not the field should be in readOnly mode
 
 ### `Data`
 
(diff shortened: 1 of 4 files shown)


matthewlynch:
Looks good!

chrisvxd:
@localhosted Please have a look at this proposal and let me know if it works for your use case!

README.md
```diff
@@ -110,6 +110,37 @@ const myPlugin = {
...
 ```
 
+## Custom fields
+
+Puck supports custom fields using the `custom` field type and `render` method.
+
+In this example, we optionally add the `<FieldLabel>` component to add a label:
+
+```tsx
+import { FieldLabel } from "@measured/puck";
```
chrisvxd, line 120: Needs testing. I was getting issues when testing via demo application without reaching directly into the dir (i.e. I had to import via `import { FieldLabel } from "@measured/puck/components/InputOrGroup"`)

README.md
```diff
@@ -110,6 +110,37 @@ const myPlugin = {
...
+
+export const MyComponent: ComponentConfig = {
+  fields: {
+    myField: {
+      type: "custom",
+      render: ({ field, name }) => {
```
chrisvxd, line 126: Missing `onChange` and `value`

simon360:
@chrisvxd love this!

I'm trying to think through how I might implement a custom `markdown` field type with this, ideally without the need for a client-side Markdown renderer on the customer-facing pages (which tends to be a bit chunky in terms of bundle size, and is generally nicer to handle server-side).

For Markdown in particular, there are two "values" to consider: the pure markdown itself, which would presumably be shown in a `textarea` element, and the HTML produced by a Markdown renderer like `remark`, which would be fed into the preview area, and perhaps need to be saved alongside the original Markdown.

Could probably work around the need to save two values when hitting publish, by saving only the Markdown itself and implementing a Markdown renderer in the page rendering piece itself. But I'm not sure if there's a good way to say "the field should show this value, but the preview should render something else." Put another way: I can picture how to render Markdown to HTML when rendering the real page, but I don't see a way to convert Markdown to HTML on the edit screen for the purpose of rendering a preview.

In other words:
* the field shows `value`
* the preview shows `renderToMarkdown(value)`
* hitting Publish sends `value` to the API
* in, say, Next.js, you would use `renderToMarkdown` in the server-side rendering to convert the markdown to HTML, which can be passed through `dangerouslySetInnerHrtml`

Sorry if this is a bit rambly... just wondering if you've got any idea how this could work, and/or if this RFC could accommodate it?

chrisvxd:
@simon360 As discussed via DMs, this could be solved by using JSON for the `value` and storing two values, one for the editor and one for the markdown component.

We don't currently have a first party method to strip out the additional editor metadata, so the end user payload will be slightly larger than necessary. This is a similar problem for external adaptors.

Alternatively you could store the data as HTML and convert it back into markdown for the editor. I've not tested this before but don't see why it wouldn't work.

localhosted:
This would works well for my needs

This way would it be possible to also overwrite the field renderer for any type (string, number, etc) or only for the "custom" type?

chrisvxd:
@localhosted It's up to you to determine what type to support in your `render` method when using `custom`.

If you look at the example in the PR description, you'll notice I'm rendering a `text` input.

matthewlynch (approved):

monospaced (commented):
A couple of things I'm wondering about. Can `render` be used for types other than `custom`? And if so, what's the expected behaviour in that case? Also what's the desired behaviour of type `custom` if `render` is not provided?

chrisvxd:
@monospaced The `type` is actually ignored here, and consequently `render` may appear for the `text` type, which is probably a bit confusing.

Suggest we should either:

1. lock it down so it only ever appears for the `custom` type (my preference)
2. or open it up, so it can be used for any type (I don't think this makes sense because `type` is actually just the input to show, and not the data type)

Any thoughts?

monospaced:
> 1. lock it down so it only ever appears for the `custom` type (my preference)

@chrisvxd this is my preference also 👍🏻

monospaced (approved):
Great! 🚀

Issue #48, opened 2023-09-06
Ensure code generated by create-puck-app can be built


chrisvxd:
The code generated by the generator doesn't get built on publish, which makes it easy to publish code that will fail when the user tries to build it. We should build all recipes as generated by the generator as part of a GitHub workflow for PRs, failing if any build fails.

To test generators, we need to:
* either generate a new app and compile that
* or share code between the recipes and generator, and build the recipe (see #10)

Example: #46

chrisvxd:
@abhishek276533 following on from #59, I can reproduce on a clean-clone. It's possible something was getting cached.

I'll investigate and report back.

```
yarn run v1.22.19
$ next build
- info Creating an optimized production build
- info Compiled successfully
- info Linting and checking validity of types
- info Collecting page data
[    ] - info Generating static pages (0/4)TypeError: Cannot read properties of null (reading 'root')
    at Module.generateMetadata (/Users/chrisvilla/Projects/tmp/puck/recipes/next/.next/server/chunks/801.js:175:29)
    at /Users/chrisvilla/Projects/tmp/puck/recipes/next/node_modules/next/dist/lib/metadata/resolve-metadata.js:199:24
    at /Users/chrisvilla/Projects/tmp/puck/recipes/next/node_modules/next/dist/server/lib/trace/tracer.js:117:36
    at NoopContextManager.with (/Users/chrisvilla/Projects/tmp/puck/recipes/next/node_modules/next/dist/compiled/@opentelemetry/api/index.js:1:7057)
    at ContextAPI.with (/Users/chrisvilla/Projects/tmp/puck/recipes/next/node_modules/next/dist/compiled/@opentelemetry/api/index.js:1:516)
    at NoopTracer.startActiveSpan (/Users/chrisvilla/Projects/tmp/puck/recipes/next/node_modules/next/dist/compiled/@opentelemetry/api/index.js:1:18086)
    at ProxyTracer.startActiveSpan (/Users/chrisvilla/Projects/tmp/puck/recipes/next/node_modules/next/dist/compiled/@opentelemetry/api/index.js:1:18847)
    at /Users/chrisvilla/Projects/tmp/puck/recipes/next/node_modules/next/dist/server/lib/trace/tracer.js:106:107
    at NoopContextManager.with (/Users/chrisvilla/Projects/tmp/puck/recipes/next/node_modules/next/dist/compiled/@opentelemetry/api/index.js:1:7057)
[...]

Issue #10, opened 2023-06-11
Generate templates from recipes


chrisvxd:
The code in `/recipes` and `/packages/create-puck-app/templates` is currently duplicated. We should reconcile these to prevent users from committing changes to the wrong directory. See #120 for example.

## Solution

Suggest we should add a script that either:

1. (**Preferred**) Generate the template from the recipe, as recipes are the logical entry point for people to make changes.
2. Generate the recipe from the template

## Considerations

* If going with option 1, preserve the handlebars template files (package.json, anything else)
* Add a CI check to make sure there is no diff (run the script and check for working changes. None should be present)

PR #59, opened 2023-09-10
ci: Add workflow file for pull requests

abhishek276533:
Adds ci file which checks lint & formatting along with build for pull requests & direct push

@chrisvxd Let me know if there are any changes you want

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -22,6 +22,9 @@ jobs:
           node-version: "18.16.1"
           registry-url: "https://registry.npmjs.org"
 
+      - name: Check Yarn Version
+        run: yarn --version
+
       - name: Install dependencies
         run: yarn
 
@@ -31,18 +34,5 @@ jobs:
             echo "Linting and formatting checks passed."
           else
             echo "Linting or formatting checks failed. Please fix the issues."
-            exit 1
-          fi
-
-      - name: Install dependencies and Build generated code
-        run: |
-          cd recipes/next
-          yarn 
-          yarn build
-
-      - name: Check for build failures
-        run: |
-          if [ $? -ne 0 ]; then
-            echo "Build failed. Please fix the issues."
             exit 1
           fi
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -34,10 +34,9 @@ jobs:
             exit 1
           fi
 
-      - name: Create and build generated code
+      - name: Build generated code
         run: |
-          npx create-puck-app test-ci-app
-          cd test-ci-app
+          cd recipes/next
           yarn build
 
       - name: Check for build failures
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,45 @@
+# This workflow will run all checks required for a PR to be merged.
+
+name: Build and Test ci
+
+on:
+  push:
+    branches: [main]
+  pull_request:
+    branches: [main]
+
+jobs:
+  build-and-test:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout code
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -0,0 +1,45 @@
...
+            echo "Linting or formatting checks failed. Please fix the issues."
+            exit 1
+          fi
+
+      - name: Build generated code
+        run: yarn build
+
+      - name: Check for build failures
+        run: |
+          if ! ${{ success() }}; then
```
chrisvxd, line 42: Looks like `success` is not defined -

https://github.com/measuredco/puck/actions/runs/6136047430/workflow#L41

chrisvxd (changes requested):
Thanks for the contribution @abhishek276533! Very necessary PR.

Unfortunately the build failed - please see my comment.

chrisvxd:
I just realised this issue is marked as fixing #48, but that's not quite right. This PR adds linting / tests, but doesn't generate and run an app (per #48). That doesn't make this PR any less valid, but we shouldn't close #48.

.github/workflows/ci.yml
abhishek276533: I've added the fix for it in the latest commit.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,48 @@
...
+      - name: Check linting and formatting
+        run: |
+          if yarn lint && yarn format --check; then
+            echo "Linting and formatting checks passed."
+          else
+            echo "Linting or formatting checks failed. Please fix the issues."
+            exit 1
+          fi
+
+      - name: Create and build generated code
```
chrisvxd, line 37: This won't succeed because the generator requires user input and has no way to accept the defaults.

Suggest for now you we don't run the generator and instead just build `recipe/next` because this is (manually, see #10) kept in sync with the generator.

Alternatively, we could remove this step for now.

chrisvxd (changes requested):
Nice, nearly there now! 👍

Build is failing due to issue with generator step - please see comment.

.github/workflows/ci.yml
abhishek276533: Ah I see, Sure have added the changes

.github/workflows/ci.yml
```diff
@@ -0,0 +1,47 @@
...
+            echo "Linting and formatting checks passed."
+          else
+            echo "Linting or formatting checks failed. Please fix the issues."
+            exit 1
+          fi
+
+      - name: Build generated code
+        run: |
+          cd recipes/next
+          yarn build
```
chrisvxd, line 40: `yarn build` is failing. I think because you need to also run `yarn` inside this directory. Recipes aren't included in the workspace so the packages won't get installed during the "Install dependencies" step.

chrisvxd (changes requested):

.github/workflows/ci.yml
chrisvxd: https://github.com/measuredco/puck/actions/runs/6138784148/job/16674530004?pr=59

abhishek276533: Yeah, makes sense. Have added the changes in the latest commit

chrisvxd:
@abhishek276533 Hmm, I can't reproduce but I think let's just remove the final "Install dependencies and Build generated code" build step for now and address #48 separately, along with #10. This at least gets the CI in.

abhishek276533:
@chrisvxd Have you tried running `yarn build` inside the `recipes/next` directory?
I've added the requested changes.

chrisvxd (approved):
Working now! Thanks @abhishek276533.

Yes I tested yarn build in the recipe and it worked. It might be my local environment but let's discuss in #48.

Issue #60, opened 2023-09-10
Feature Request: Sub objects


Danm72:
It would be nice to be able to add sub objects pretty much exactly the way `arrayFields` currently behaves just without the 'add item' UI!

For example in the demo site we have full blocks like `Hero` which only have some theme/config options but we also have components like `H1` on the page at the same time.

The `Hero` likely contains a H1 and it would be nice to be able to edit the H1 as a dedicated object.

![Export-1694337157447](https://github.com/measuredco/puck/assets/1926968/1afa4326-10bd-4533-ab48-c65277ce9b1a)

chrisvxd:
Great suggestion, discussion via PR #62

chrisvxd:
Closing this in favour of #64 for now. May revisit in future.

Danm72:
Awesome fair play you keeping this moving!

PR #62, opened 2023-09-10
Adding 'object' type as a datatype option

Danm72:
Which can be used to embed objects within the sidebar to embed other editor objects within blocks or to create more details objects.

In this case you can see the 'component' Heading is used to supply object fields to the Block Hero, the values are then used to actually style the heading object within the hero.

https://github.com/measuredco/puck/assets/1926968/f4fd1251-5f5a-46d9-8003-f45661e71c35

#60

### DIFF
--- a/apps/demo/app/configs/custom/blocks/Hero/index.tsx
+++ b/apps/demo/app/configs/custom/blocks/Hero/index.tsx
@@ -78,14 +78,14 @@ export const Hero: ComponentConfig<HeroProps> = {
     padding: "64px",
   },
   render: ({
-             align,
-             title,
-             description,
-             buttons,
-             padding,
-             imageUrl,
-             imageMode,
-           }) => {
+    align,
+    title,
+    description,
+    buttons,
+    padding,
+    imageUrl,
+    imageMode,
+  }) => {
     return (
       <Section
         padding={padding}
--- a/apps/demo/app/configs/custom/blocks/Hero/index.tsx
+++ b/apps/demo/app/configs/custom/blocks/Hero/index.tsx
@@ -1,147 +1,187 @@
 /* eslint-disable @next/next/no-img-element */
 import React from "react";
-import { ComponentConfig } from "@measured/puck";
+import {ComponentConfig} from "@measured/puck";
 import styles from "./styles.module.css";
-import { getClassNameFactory } from "@measured/puck/lib";
-import { Button } from "@measured/puck/components/Button";
-import { Section } from "../../components/Section";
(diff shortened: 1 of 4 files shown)


chrisvxd (commented):
Thanks for the issue and WIP PR @Danm72! Great work!

I'm think I'm in favour of adding support for object params (because we have `array`), and can certainly think of some use-cases where a flat model doesn't cut it. Styling inner components is likely just one use case of many (and one that will probably become less important with #37, which will allow for nested components).

A few things come to mind when looking through your PR and thinking about this feature:

1. I'm not a huge fan of the `array` UI in general, and not totally sure if it makes sense for `object` fields. It's totally fine to reuse as part of this initial implementation, but we might want to revisit it at some point. Which brings me onto...
2. Inline fieldsets at the top-level (with headings) are something I've considered as a way to break the form up into separate areas of related fields. It strikes me that the **Heading** scenario here might also work with an inline **Heading** fieldset. Again, not for this PR but I think worth the discussion point.
3. And finally, the only thing I think we should perhaps reconsider is whether to include the Heading fields in the Hero for the demo component. Whilst it shows the functionality well, I worry it bloats the demo out and isn't something need to necessary know about from first impressions.

PS how are you producing those lovely screen recordings? 😍

Danm72:
Awesome feedback I'll respond when I'm back at my PC but all sounds logical to me!

I agree about the array UI and might noodle about with the fieldset concept!

Good point on the demo.

Videos are made with the awesome https://www.screen.studio

chrisvxd:
Great! And no rush.

I've made a separate ticket to discuss a fieldset API - https://github.com/measuredco/puck/issues/64

Danm72:
Hey @chrisvxd added those small changes lemme know if you see anything else. Feel free to go in another direction with this!

chrisvxd:
Hey @Danm72! Just checking why you closed this?

Was going to review as soon as I had time. Happy with general direction

Danm72:
Hey Chris,

Whoops sorry I thought it was going stagnant and didn't want to leave it sitting there! 

Feel free to merge it or run with it I was just thinking your other PR around field rendering might be a little more advanced with the 'custom' type.

Maybe both flows could work. Change my PR to use type 'custom' instead of object and the default custom renderer is the array renderer?

Unsure if id go with object or custom..

chrisvxd:
Hi @Danm72! Been introspecting on this a bit and think we're going to close it out per your suggestion until after we've looked fieldsets and tabs in #64. Thanks for all your work :)

PR #80, opened 2023-09-16
fix the example code in README.md

philipodev:
Hi, this is a great project! 
Had some minor issues with the example code in README but i figured them out.

The example is lacking the `content` and `root` properties in the `initialData`. (so i get a TypeError for trying to read title of undefined root).
Also added the neccessary css import.

A better fix would probably be to make them optional instead of just empty objects. Happy to close this one and work on that instead if you feel like that would be a more preferable fix.

Cheers!

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,6 +16,7 @@ Render the editor:
 ```jsx
 // Editor.jsx
 import { Puck } from "@measured/puck";
+import "@measured/puck/dist/index.css";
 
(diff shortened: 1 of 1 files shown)


chrisvxd (approved):
Thanks @philipodev! Agree that making them optional would be better, but this is still accurate. Feel free to open another PR to make the underlying types optional and undo the data defaulting.

PR #105, opened 2023-09-22
fix: scope styles in OutlineList

chrisvxd:
Some environments were overriding them. Generally we should avoid any element selectors.

### DIFF
--- a/packages/core/components/OutlineList/styles.module.css
+++ b/packages/core/components/OutlineList/styles.module.css
@@ -18,12 +18,12 @@
   content: "";
 }
 
-.OutlineList li {
+.OutlineListItem {
   position: relative;
   margin-bottom: 4px;
 }
 
-.OutlineList li::before {
+.OutlineListItem::before {
   background: var(--puck-color-grey-7);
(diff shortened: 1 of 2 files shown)


monospaced (approved):

packages/core/components/OutlineList/styles.module.css
```diff
@@ -33,11 +33,11 @@
   content: "";
 }
 
-.OutlineList .OutlineList-clickableItem:hover {
+.OutlineListItem--clickable:hover {
```
chrisvxd, line 36: Hm, this style isn't getting applied

chrisvxd:
Commit name needs fixing, or needs squashing

packages/core/components/OutlineList/styles.module.css
monospaced: Perhaps?

```suggestion
.OutlineList  .OutlineListItem--clickable:hover {
```

chrisvxd: Resolved

Issue #15, opened 2023-06-12
Firefox doesn't respect zoom


chrisvxd:

chrisvxd:
Closed in #353

PR #107, opened 2023-09-24
fix: style clashing with dark mode root element

### LINKED ISSUE #106: Radix and Puck styles clashing
Various issues occur when using Radix with Puck resulting in illegible text.

See below screenshot from FF on MacOS, running dark mode and Radix using system theme.

image

Reported by @gothfemme in Discord.

gothfemme:
- Adds in styles so Puck doesn't inherit styles from the document body. 
- Fixes an issue where the preview root would inherit/be blocked by the editor frame. 
  - I'm just faking the frame with a box shadow.
- Fixes an issue where horizontal scrollbars were always visible for `SidebarSection`'s even when not overflowing.
- Closes #106

### DIFF
--- a/packages/core/components/Puck/index.tsx
+++ b/packages/core/components/Puck/index.tsx
@@ -269,6 +269,8 @@ export function Puck({
                   <header
                     style={{
                       gridArea: "header",
+                      color: "var(--puck-color-black)",
+                      background: "var(--puck-color-white)",
                       borderBottom: "1px solid var(--puck-color-grey-8)",
                     }}
                   >
@@ -398,7 +400,6 @@ export function Puck({
 
                   <div
                     style={{
-                      background: "var(--puck-color-grey-10)",
                       padding: 32,
                       overflowY: "auto",
                       gridArea: "editor",
@@ -410,8 +411,8 @@ export function Puck({
                     <div
                       className="puck-root"
                       style={{
-                        background: "white",
                         border: "1px solid var(--puck-color-grey-8)",
+                        boxShadow: "0px 0px 0px 3rem var(--puck-color-grey-10)",
                         zoom: 0.75,
                       }}
                     >
@@ -428,6 +429,7 @@ export function Puck({
                       fontFamily: "var(--puck-font-stack)",
                       display: "flex",
                       flexDirection: "column",
+                      background: "var(--puck-color-white)",
                     }}
                   >
                     <FieldWrapper data={data}>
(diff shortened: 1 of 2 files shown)


gothfemme:
I kept the styles in `Puck/index.tsx` as inline rather than creating a new file for it as I imagine that's outside the scope of this ticket

packages/core/components/Puck/index.tsx
```diff
@@ -410,8 +411,8 @@ export function Puck({
                     <div
                       className="puck-root"
                       style={{
-                        background: "white",
                         border: "1px solid var(--puck-color-grey-8)",
+                        boxShadow: "0px 0px 0px 32px var(--puck-color-grey-10)",
```
chrisvxd, line 415: Not sure I understand the reason for this change? Can you share a screenshot of the original issue?

This fix causes and additional white border in Safari and Chrome due to the zoom factor (FF doesn't respect zoom per #15)

image

packages/core/components/SidebarSection/styles.module.css
```diff
@@ -12,7 +13,7 @@
   background: white;
   padding: 16px;
   border-bottom: 1px solid var(--puck-color-grey-8);
-  overflow-x: scroll;
+  overflow-x: auto;
```
chrisvxd, line 16: Can you split this fix into a separate commit for our changelogs?

```
fix: only show sidebar scroll bars if necessary
```

chrisvxd (changes requested):
Excellent stuff @gothfemme! Left a couple of comments.

Updating the inline styles is definitely the right place for now. These really need to get pulled out at some point.

packages/core/components/SidebarSection/styles.module.css
gothfemme: yeah, can do.

packages/core/components/Puck/index.tsx
gothfemme: Ah yeah, it definitely isn't the ideal solution. Reason was setting background on the frame would cause the preview root to either:
1. Inherit from the frame if set to inherit
2. If no BG is set on preview root, the frame BG is all that would show rather than the background for the document body

image
The example here being the 2nd case with the frame BG taking over.

This was my primary reason for asking about portals in #106. I could also imagine a fix involving pseudo-elements, though would need to move some styles out of inline styles.

chrisvxd (approved):
Perfect 🚀

PR #110, opened 2023-09-25
docs: Added logo

yanneves:
Generated for upcoming asyncjs/async-website#351 talk using OpenJourney v4


chrisvxd (commented):
Love this @yanneves! 

However, we're actually in the process of creating a brand identity with designer, so I think we'll probably hang tight until that's come together.

Issue #90, opened 2023-09-20
Add tests to DropZone component


chrisvxd:
The DropZone component, introduced in #37, is complex and introduces many nuanced behaviours that can be easily broken.

It currently has no test coverage, which should be addressed.

## Requirements

* Configure testing-library
* Get as close to 100% coverage as reasonable

## Considerations

* Probably best done after #89

Issue #89, opened 2023-09-20
Add tests to data reducer


chrisvxd:
The DropZones PR #37 introduces a reducer (found in `/packages/core/libs/reducer`) to handle the increasing complexity necessary for supporting drop zones. 

See https://github.com/measuredco/puck/pull/37#discussion_r1330241072 and https://github.com/measuredco/puck/pull/37#discussion_r1330241072.

## Requirements

* Setup testing framework
* 100% test coverage of all reducer mutations
* No other tests should be added in this PR

## Considerations

* Jest is probably best shout here.
* We may be better off separating out the `useReducer` hook from the reducer to allow for pure unit tests.

PR #111, opened 2023-09-26
undo redo history

Yuddomack:
hi! I tried working on it.
but I did this task without fully understand, so need to some your feedback

thank you!


packages/core/lib/use-puck-data.ts
```diff
@@ -0,0 +1,136 @@
...
+    canForward,
+    canRewind,
+    rewind,
+    forward,
+    record,
+    itemSelector,
+    setItemSelector,
+    selectedItem,
+  };
+}
```
chrisvxd, line 136: Hm I'm a bit unsure about this particular file. Perhaps we would be be better off going with deep object diff library, and store the delta, rather than re-calling the dispatchers here.

You could use [deep-diff](https://www.npmjs.com/package/deep-diff) and the [applyChange API](https://github.com/flitbit/diff/tree/e2b5b86add402ae4373ab025ee7e9a98e3424023#changes), or try out [rfc6902](https://github.com/chbrown/rfc6902). I don't think it really matters which.

Example using `deepDiff` inside `reducer.ts`, without modifying existing reducer code to avoid polluting the state machine

```tsx

const storeHistory = (reducer: StateReducer) => {
  const history: any = [];

  return (data, action) => {
    const newData = reducer(data, action);

    const diff = deepDiff.diff(data, newData);

    history.push(diff);

    return newData;
  };
};

export const createReducer = ({ config }: { config: Config }): StateReducer =>
  storeHistory((data, action) => {
    if (action.type === "insert") {
      const emptyComponentData = {
        type: action.componentType,
        props: {
          ...(config.components[action.componentType].defaultProps || {}),
          id: generateId(action.componentType),
        },
      };
   ...
```

chrisvxd (changes requested):
Thanks for the big effort again @Yuddomack! 

I've left a suggestion for how to simplify the history recording using a diffing library.

packages/core/lib/reducer.ts
```diff
@@ -42,6 +42,13 @@ type ReplaceAction = {
   data: any;
 };
 
+type RecoveryAction = {
+  type: "recovery";
+  destinationIndex: number;
+  destinationZone: string;
+  data: any;
+};
+
```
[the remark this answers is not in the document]
Yuddomack:
```javascript
// before data
data = [
  { type: "Hero", ... },
  { type: "VerticalSpace", ... },
  { type: "Heading", ... },
  ...
]

// data after removed 'Hero' // dispatch({ type: "remove", index, zone })
data = [
  { type: "VerticalSpace", ... },
  { type: "Heading", ... },
  ...
]
```

umm I thought What information do I need to restore what was removed?
-> zone, index, (before removed) data

the data could be extracted using `getItem` when excuted remove action.
but I needed the ability to put it back in its original location. (`insert action` is based on new data object(emptyComponentData), it is judged that it cannot be reused)

so I added an implementation for the recovery action.
(Perhaps this idea should be discussed on https://github.com/measuredco/puck/pull/111#discussion_r1337537243 first.)

packages/core/lib/use-puck-data.ts
Yuddomack: Oh this idea looks really cool!
I'll come back using this ⛹️‍♂️

Yuddomack: Hi! @chrisvxd
I tried applying deep-diff. I still have a little considered about each responsibility, but it ready to review 🔥

packages/core/lib/__tests__/use-action-history.spec.tsx
```diff
@@ -0,0 +1,192 @@
...
+    });
+
+    test("rewind function of action should be execute", () => {
+      actions.reverse().forEach(({ rewind }) => {
+        act(() => renderedHook.result.current.rewind());
+        expect(rewind).toBeCalled();
+      });
+    });
+
+    test("canRewind should be true until before last rewind", () => {
```
chrisvxd, line 115:
```suggestion
    test("canRewind should be true until the start of the history is reached", () => {
```

packages/core/lib/__tests__/use-action-history.spec.tsx
```diff
@@ -0,0 +1,192 @@
...
+      actions.reverse().forEach((_, i) => {
+        act(() => renderedHook.result.current.rewind());
+        expect(renderedHook.result.current.canRewind).toBe(
+          i !== actions.length - 1
+        );
+      });
+    });
+  });
+
+  describe("when run forward all with action histories (after all rewinded)", () => {
```
chrisvxd, line 125:
```suggestion
  describe("when multiple action histories are recorded, rewound and then fast-forwarded", () => {
```

packages/core/lib/__tests__/use-action-history.spec.tsx
```diff
@@ -0,0 +1,192 @@
...
+    });
+
+    test("forward function of action should be execute", () => {
+      actions.forEach(({ forward }) => {
+        act(() => renderedHook.result.current.forward());
+        expect(forward).toBeCalled();
+      });
+    });
+
+    test("canForward should be true until before last forward", () => {
```
chrisvxd, line 156:
```suggestion
    test("canForward should be true until the end of the history is reached", () => {
```

packages/core/lib/__tests__/use-action-history.spec.tsx
```diff
@@ -0,0 +1,192 @@
...
+          renderedHook.result.current.record({
+            rewind,
+            forward,
+          })
+        );
+      });
+      act(() => renderedHook.result.current.rewind());
+    });
+
+    test("recent histories are rewritten when recording a new action ", () => {
```
chrisvxd, line 185:
```suggestion
    test("the forward history is overridden when recording a new action ", () => {
```

packages/core/lib/use-puck-history.ts
```diff
@@ -0,0 +1,71 @@
...
+  dispatch: (action: PuckAction) => void;
+}) {
+  const { canForward, canRewind, rewind, forward, record } = useActionHistory();
+
+  useHotkeys("ctrl+z", rewind, [rewind]);
+  useHotkeys("ctrl+y", forward, [forward]);
+
+  const [snapshot] = useDebounce(data, DEBOUNCE_TIME);
+
+  const handleRecordDiff = useDebouncedCallback((newData: Data) => {
```
chrisvxd, line 30: This is excellent. What do you think about adding tests here?

No worries if not - we don't have them anywhere else yet.

packages/core/lib/__tests__/use-action-history.spec.tsx
```diff
@@ -0,0 +1,192 @@
...
+      const newAction = { rewind: jest.fn(), forward: jest.fn() };
+      act(() => renderedHook.result.current.record(newAction));
+
+      expect(renderedHook.result.current.canForward).toBe(false);
+    });
+  });
+});
```
chrisvxd, line 192: Could you add this to the CI file so the tests run as part of this build?

```yml
      - name: Run tests
        run: yarn test
```

`.github/workflows/ci.yml` around L30 would be great

chrisvxd (changes requested):
This is really excellent work! Thank you 🙏

Very happy with this approach. Most of my feedback is on phrasing in the unit tests.

packages/core/lib/use-puck-history.ts
```diff
@@ -0,0 +1,71 @@
...
+      rewind: () => {
+        const target = structuredClone(data);
+        _diff.reduce((target, change) => {
+          revertChange(target, true, change);
+          return target;
+        }, target);
+        dispatch({ type: "set", data: target });
+      },
+    });
+  }, DEBOUNCE_TIME);
```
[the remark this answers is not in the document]
Yuddomack: If the user quickly modifies data, each action is recorded.

like this

```
quickly typing) This page was built with Puck -> This page was built with Puck1234
rewind) This page was built with Puck123
rewind) This page was built with Puck12
rewind) This page was built with Puck1
rewind) This page was built with Puck
```

and I thought that recording of actions at once was good for user experience. (also memory of browser)

```
quickly typing) This page was built with Puck -> This page was built with Puck1234
rewind) This page was built with Puck
```

that's it!

packages/core/lib/use-puck-history.ts
Yuddomack: When #89 or #90 is completed, I will try with refer to that!

I think this area would be meaningful if it were basically at the level of mocking the reducer or integration testing of the puck component, but I need a better understanding of your testing philosophy.

If so, I can also help write tests for other components (or even e2e with demo app).

Yuddomack:
Hi! @chrisvxd Your request has arrived!
- apply shortcut with each platform ([The meta key is the same as cmd on macOS and os key on Windows in react-hotkeys-hook](https://react-hotkeys-hook.vercel.app/docs/documentation/useHotkeys/basic-usage))
- resolve conflicts
- test descriptions + ci.yml
- some comments (for test, debounce)

packages/core/lib/use-puck-history.ts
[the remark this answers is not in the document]
chrisvxd: Right, perfect. Thanks for clarifying!

packages/core/package.json
```diff
@@ -8,26 +8,38 @@
   "scripts": {
     "lint": "eslint \"**/*.ts*\"",
     "build": "rm -rf dist && tsup index.ts",
+    "test": "jest",
```
chrisvxd, line 11: Oops, we need to add this to the root `package.json` (`turbo run test`) for the CI script to work! Would you add mind adding that too, please?

packages/core/lib/use-puck-history.ts
chrisvxd: That makes sense!

#89 will be unit tests, #90 more integration but I'm unclear on the strategy yet.

I'm working on #89 next. We'll pick this up afterwards.

packages/core/package.json
```diff
@@ -8,26 +8,38 @@
...
     "tsconfig": "*",
     "tsup": "^6.7.0",
     "tsup-config": "*",
     "typescript": "^4.5.2"
   },
   "dependencies": {
+    "deep-diff": "^1.0.2",
+    "event-emitter": "^0.3.5",
     "react-beautiful-dnd": "^13.1.1",
+    "react-feather": "^2.0.10",
```
chrisvxd, line 41: This looks like a merge conflict - react-feather should have been removed in favour of the `devDependency` (see #121)

packages/core/lib/use-puck-history.ts
```diff
@@ -0,0 +1,71 @@
...
+  data,
+  dispatch,
+}: {
+  data: Data;
+  dispatch: (action: PuckAction) => void;
+}) {
+  const { canForward, canRewind, rewind, forward, record } = useActionHistory();
+
+  useHotkeys("meta+z", rewind, { preventDefault: true });
+  useHotkeys("meta+y", forward, { preventDefault: true });
```
chrisvxd, line 26: Sorry, on macOS redo is actually CMD + SHIFT + Z.

Perhaps we should split this into two hotkeys?

```suggestion
  useHotkeys("cmd+shift+z", forward, { preventDefault: true });
  useHotkeys("ctrl+y", forward, { preventDefault: true });
```

or just define meta for both - I don't think it hurts

```suggestion
  useHotkeys("meta+shift+z", forward, { preventDefault: true });
  useHotkeys("meta+y", forward, { preventDefault: true });
```

(untested examples)

chrisvxd (changes requested):
Wonderful! A couple more tiny comments before approval.

We're planning on doing v0.8.0@latest release soon, so will hold off merging this until _after_ that goes in to give it a chance to be tested in canary.

packages/core/package.json
chrisvxd: You'll also need to add this to `turbo.json`

```
"test": {},
```

packages/core/lib/use-puck-history.ts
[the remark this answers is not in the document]
chrisvxd: Could we actually increase the DEBOUNCE_TIME here?

Maybe 250ms would be okay. `100` feels like it's still recording a lot of what I'm typing.

packages/core/lib/use-puck-history.ts
Yuddomack: cool! I like second one! (I checked it work)

Yuddomack:
Hi! @chrisvxd 
Additional requests are also prepared! (maybe? 😆)

packages/core/lib/use-puck-history.ts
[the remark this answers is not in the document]
chrisvxd: Much better!

chrisvxd (approved):
🚀 Epic work @Yuddomack!

chrisvxd:
@Yuddomack available in `0.9.0-canary.222697e`!

chrisvxd:
@Yuddomack would love to have you in the Discord, if you're interested: https://discord.com/invite/D9e4E3MQVZ

Yuddomack:
@chrisvxd cool! I joined it

PR #114, opened 2023-09-26
fix: prevent removal of fieldset padding

chrisvxd:
Some environments reset padding, resulting in broken styles. This PR explicitly adds the padding to override this reset.

## Example

image

## CSS

image

### DIFF
--- a/packages/core/components/InputOrGroup/styles.module.css
+++ b/packages/core/components/InputOrGroup/styles.module.css
@@ -97,8 +97,7 @@
   border: none;
   border-top: 1px solid var(--puck-color-grey-8);
   margin: 0;
-  padding-bottom: 16px;
-  padding-top: 16px;
+  padding: 16px;
 }
 
 .Input-arrayItem


packages/core/components/InputOrGroup/styles.module.css
```diff
@@ -97,8 +97,7 @@
   border: none;
   border-top: 1px solid var(--puck-color-grey-8);
   margin: 0;
-  padding-bottom: 16px;
-  padding-top: 16px;
+  padding: 16px;
```
monospaced, line 100: I _think_ I understand what this is about. Previously we were relying on inheriting (browser-default) styles to provide left and right padding to fieldsets - but since these styles aren't consistent, or could have been overridden, we're now making them explict?

monospaced (approved):
I have a clarifying comment, but approving anyway as this is a small change. But do let me know if I have this right or not 🙏🏻 And if it's something completely different, we should probably discuss.

packages/core/components/InputOrGroup/styles.module.css
chrisvxd: Spot on!

monospaced: Noice

PR #115, opened 2023-09-27
Remove :has to give Firefox full DropZone support

chrisvxd:
Relational pseudo-class [`:has`](https://developer.mozilla.org/en-US/docs/Web/CSS/:has) was used for determining child hover states in DropZones - that is, when a child _area_ is hovered, we don't want to show the hover state for the parent.

However, Firefox [doesn't support `:has`](https://caniuse.com/css-has).

This PR replaces `:has` with JavaScript. Using JavaScript to determine the hover state of a child is inherently less performant than using CSS, but I believe this is a fair trade-off and hopefully negligible.

## Testing

* Ensure Firefox behaviour now matches other browsers (such as Chrome)
* Compare hover and drag states remain identical with the existing demo for other browsers (i.e. Chrome)


monospaced (commented):
@chrisvxd My reservation with this, is that Firefox is scheduled to support `:has` in 119, release date 24/10/2023. It's already in the 119 nightlies, would expect it to show up in the 119 beta shortly (118 was released yesterday).

Maybe just wait?

https://bugzilla.mozilla.org/show_bug.cgi?id=1853701#c8
https://wiki.mozilla.org/index.php?title=Release_Management/Calendar&redirect=no#Future_branch_dates

chrisvxd:
Agree with you. Opening #117 to track in the meantime and closing this.

PR #121, opened 2023-10-02
perf: bundle used react-feather icons only

### LINKED ISSUE #118: React Feather not treeshaking correctly
Hi Puck Team,

I've noticed that the 'react-feather' library isn't tree-shaking as it should the full library is loading in (see [Bundlephobia](https://bundlephobia.com/package/@measured/puck@0.8.0-canary.e4cab76)).  Digging into this, it seems that 'react-feather' _should_ [treeshake itself now](https://github.com/feathericons/react-feather/issues/6) if the tsconfig has `module: esnext` set.  

I wonder if it's a problem with the tsconfig for react-library? https://github.com/measuredco/puck/blob/e4cab76e385ba735778de9cb2514a15338ead10a/packages/tsconfig/react-library.json#L8C4-L8C24 should "ESNext" be "esnext"?

chrisvxd:
Moving to dev dependencies causes `tsup` to bundle the external dependencies where necessary.

This increases the direct bundle size by 72KB but should drop the [328KB react-feather dependency](https://bundlephobia.com/package/@measured/puck@0.8.0-canary.e4cab76).

**Before**
CJS dist/index.js  94.00 KB

**After**
CJS dist/index.js  166.42 KB

Closes #118

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -20,6 +20,7 @@
     "@types/react-dom": "^18.2.0",
     "eslint": "^7.32.0",
     "eslint-config-custom": "*",
+    "react-feather": "^2.0.10",
     "tsconfig": "*",
     "tsup": "^6.7.0",
     "tsup-config": "*",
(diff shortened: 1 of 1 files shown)


monospaced (approved):
🪶

PR #122, opened 2023-10-02
Feat: enable `classNames` for Dropzone component

Leweyse:
It's possible to pass inline styles to the Dropzone component, but it would be great to be able to pass class names as well.

### DIFF
--- a/packages/core/components/DropZone/index.tsx
+++ b/packages/core/components/DropZone/index.tsx
@@ -1,4 +1,5 @@
 import { CSSProperties, useContext, useEffect } from "react";
+import classnames from "classnames";
 import { DraggableComponent } from "../DraggableComponent";
 import DroppableStrictMode from "../DroppableStrictMode";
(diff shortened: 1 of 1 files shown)


chrisvxd (commented):
Hey @Leweyse! Sorry for slow review.

I think the original `style` attribute might be getting applied in the wrong place (`content`) instead of the root of the component. You've mimicked that for className (which is sensible), but I'm thinking we might want to address `style` with a breaking change before introducing a new prop that will also require breaking.

We're also thinking about how to handle theming more broadly without adding `className` props for every element within each component.

Mind if we sit on this one for a while whilst resolve the above?

Leweyse:
Hi @chrisvxd!

No problem at all, I understand the situation. Actually, customize the editor theme is quite important for my project, so I look forward to that update.

Is there a way to track the progress and perhaps contribute to it?

chrisvxd:
@Leweyse Just created #139 (thought we already had one, but I guess not!)

Issue #100, opened 2023-09-21
DropZone follow ons


chrisvxd:
List of things that came up in #37 review. Some of these may be further spun out into smaller tickets:

## Maybe

* Make Cards full height in demo 
![](https://user-images.githubusercontent.com/98794/269591807-135f91a1-e030-4fce-a236-549af3514acd.png)
* Consider renaming Span to Manual span in columns demo

## Done
* ~Allow custom min-height on DropZones~
* ~Add docs~

## Moved

* #126
* #123
* #117
* #89, #90
* #103
* #270
* #269

PR #125, opened 2023-10-03
doc: document the DropZone component

chrisvxd:
Follow on from #37. Part of #100.


chrisvxd:
Note: Squash on merge, commit type should be `docs` not `doc`

monospaced (approved):

PR #131, opened 2023-10-04
fix: render `icon` if provided to FieldLabel

Leweyse:
Currently the `FieldLabel` component expects an optional prop `icon`, but it doesn't render it. This PR should fix that.


chrisvxd:
Available in `0.9.0-canary.fe1629e`

PR #133, opened 2023-10-05
Fix issues introduced in #132

chrisvxd:
Fixes 2 issues introduced in #132. Using `chore:` commits to avoid polluting diff.

### Before
image

### After

image

### Before

image

### After

image


monospaced (approved):
🩹

PR #135, opened 2023-10-06
feat: make root and content optional in data, easing initalisation

chrisvxd:
Follow on from #80, this makes `root` and `content` optional in the `Data` shape, making it simpler to instantiate Puck.

This was the original intention. Could also mark this as a `fix`. Could probably go either way.

## Before
```tsx
<Puck data={{ root: {}, content: [] }} config={config} />
```
## Now

```tsx
<Puck data={{}} config={config} />
```

CC @philipodev

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,10 +50,7 @@ const config = {
 };
 
 // Describe the initial data
-const initialData = {
-  content: [],
-  root: {},
-};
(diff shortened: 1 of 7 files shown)


chrisvxd:
Closing as significantly out of date and needs reworking.

Issue #87, opened 2023-09-20
Group components in the sidebar


chrisvxd:
A developer should be able to group components in the left hand sidebar.

### Option 1

Add the `tags` array to the component config object.

Additionally, this can be used restrict which components can be dropped into a DropZone (see #37)

### Option 2

As option 1, but singular `tag`.

### Option 3

A render function may provide similar functionality by allowing the developer to the override the UI. Similar to the fieldset proposal #64.

xaviemirmon:
Hey @chrisvxd 

I've created a PR for this feature following option 2.  https://github.com/measuredco/puck/pull/138

This is on the basis that if you tag a component with multiple tags, you'd then have to have multiple droppable instances of the component which doesn't make sense to me.  

I've also built this functionality in such a way that it renders the same as before so if you don't have any labels the behviour is the same.

chrisvxd:
Thanks @xaviemirmon. I think your rationale behind avoiding option 1 is interesting.

I think tags extend well to #103, which can take advantage of the tagging to specify which components can be dropped:

```tsx
<DropZone zone="content" tags=["layout"] />
```

(You could also use component names like `<DropZone zone="content" components=["HeadingBlock"] />`)

Option 3 is perhaps closer to the proposal for #64 for fieldsets, and what we've implemented in #58 for custom fields. I imagine you would add a new prop called `renderComponentList` (or something more pithy) to `Config`

```tsx

renderComponentList: () => {
  return <ComponentList>
    <ComponentList.Group title="Layout">
      <ComponentList.Item component="Section" />
      <ComponentList.Item component="Columns" />
    <ComponentList.Group />

    <ComponentList.Group title="Typography">
      <ComponentList.Item component="Heading" />
      <ComponentList.Item component="Paragraph" />
    <ComponentList.Group />
  </ComponentList>
}
```

This might be too verbose, but also provides more flexibility for the user. The APIs also aren't mutually exclusive.

_Interestingly, this option could mean we still reserve the use of `tags` (plural) on the `ComponentConfig` for the purposes of DropZones._

xaviemirmon:
@chrisvxd I agree and think the tags option for DropZone makes sense, and I have some thoughts on that, which I'll leave on the other issue.  

Thanks for outlining option three a bit more. I've got my head around the proposal now.  

[...]

PR #138, opened 2023-10-06
Adding tag functionality to component config

xaviemirmon:


packages/core/components/ComponentList/index.tsx
```diff
@@ -5,10 +5,24 @@ import { Draggable } from "react-beautiful-dnd";
...
 import { Grid } from "react-feather";
+import { Heading } from "../Heading";
 
 const getClassName = getClassNameFactory("ComponentList", styles);
 
 export const ComponentList = ({ config }: { config: Config }) => {
+  const groupedComponents = {};
+  const defaultTagLabel = config?.tags?.defaultLabel
+    ? config.tags.defaultLabel
+    : "Uncategorised";
```
chrisvxd, line 16: Can we keep "Components" for the default label here?

packages/core/types/Config.tsx
```diff
@@ -86,6 +87,9 @@ export type Config<
       Props[ComponentName]
     >;
   };
+  tags?: {
+    defaultLabel?: string;
+  };
```
chrisvxd, line 92: I think we can make better use of this API-namespace to future proof it

```
  tags: {
    default: {
      label: "Others"
    },
    layout: {
      label: "Layout"
    }
  }
```

We can then create a strict type-link between the tags specified here and the tags available for the components

packages/core/components/ComponentList/index.tsx
```diff
@@ -17,51 +31,67 @@ export const ComponentList = ({ config }: { config: Config }) => {
...
+          {Object.entries(groupedComponents)
+            .sort()
+            .map(([tag, components]) => {
+              return (
+                <>
+                  {Object.keys(groupedComponents).length > 1 ? (
+                    <Heading rank={3} size={"xs"}>
+                      {tag}
+                    </Heading>
+                  ) : null}
```
chrisvxd, line 43: I'm not totally sure that using Headings like this is the way to go for the UI here.

It also exacerbates some existing visual issues when cancelling a drag


https://github.com/measuredco/puck/assets/985961/5590e311-029e-42e5-a6db-4c42d2c7e627



I wonder if we should render multiple

chrisvxd (commented):
Thanks @xaviemirmon! This is an interesting proposal

#87 wasn't quite ready for picking up yet (we haven't marked it as "Ready" yet, still in "Triage" - we haven't got contribution guidelines yet so that's probably not clear). It's also marked with "opinions wanted" because we were keen to have some dialog before setting on a direction.

This PR implements option 2 in the proposals, which is valid, but I'd like to also discuss the other options available to us before we finalise this approach.

Regardless, this is still really useful input that can help with deciding the best path forwards. Let's pick up conversation in #87 and discuss what option 1 and 3 might also look like.

If we go with option 2, then we can just pick this PR up. I've given my feedback where appropriate!

packages/core/types/Config.tsx
xaviemirmon: Yeah, I like it. I wonder if it's also worth adding a weight field like so

` tags: {
    default: {
      label: "Others"
      weight: 0
    },
    layout: {
      label: "Layout"
      weight: 2
    }
  }`
  
  That way you could reorder the order that they display?

chrisvxd:
Converted to draft as we discuss #87

chrisvxd:
Closing in favour of #183

Issue #123, opened 2023-10-03
Support dragging between DropZone areas


chrisvxd:
#37 introduced the DropZone component, but it is not possible to drag between DropZone areas (or parents), per the "Rules of DropZones" detailed in the PR:

> 1. You can drag from the component list on the LHS into any DropZone
> 2. **You can drag components between DropZones, so long as those DropZones share a parent (also known as area)**
> 3. **You can't drag between DropZones that don't share a parent (or area)**
> 4. Your mouse must be directly over a DropZone for a collision to be detected

This restriction was implemented to overcome some of the limitations of react-beautiful-dnd (and drag-and-drop in general). I've been exploring this over the last few days but it is non-trivial. I'm breaking this issue out from #100 to keep things moving forwards, and may pause work on this for the time-being.

## Challenges

Issues that I've encountered:

1. After switching between areas, react-beautiful-dnd incorrectly offsets for the change of height during the dropping animation - **Solved by correcting the transform produced by react-beautiful-dnd during the drop animation**
2. Items need dynamically resizing during drag (which technically also affects dragging items between DropZones, not just areas) - **Solved by correcting the transform produced by react-beautiful-dnd whilst dragging based on the size of the destination DropZone**
3. After switching between areas and moving above the source area to the parent area, react-beautiful-dnd offsets the dragged item by the relative position of the source Droppable to the parent Droppable - **Not solved (see below)**

### Issues with patching 3

Unfortunately issue 3 is particularly challenging. I explored different approaches to correct this issue:

1. Modify the transform (again) - however, it is not clear how the offset is even being applied, making it difficult to calculate a valid correction
2. [Reparent](https://github.com/atlassian/react-beautiful-dnd/blob/master/docs/guides/reparenting.md) the element using the `renderClone` API - this solves the offset issue completely but breaks dragging of components that contain DropZones (nested Droppables are technically unsupported by react-beautiful-dnd). This API unmounts the Draggable on render, which causes the nested Droppables to unmount and react-beautiful-dnd to freak out.
3. Reparent the element using portals (instead of `renderClone`) to avoid unmounting - this introduces further positioning issues, but they may be solvable. Further exploration is required.
4. Migrating to dnd-kit - this is not a simple swap and introduces regressions that have already been resolved.

## Current state

https://github.com/measuredco/puck/assets/985961/bfa13a00-9ce3-4744-b094-a566fee83a53

JakeSidSmith:
[...]

PR #140, opened 2023-10-07
Fix: `position: fixed;` elements escape editor

JakeSidSmith:
Added transform to root element so `position: fixed;` elements sit relative to it.

Pictured below is the `apps/demo` with the following added to the `config/root.tsx`.

```tsx
    <header
        style={{
          position: "fixed",
          top: 0,
          left: 0,
          width: "100%",
          backgroundColor: "white",
          zIndex: 1000,
        }}
      >
```

It doesn't behave like a `position: fixed;` element normally would due to the scrolling being on an element outside of the one with the transform, but it does allow you to access the editor navigation while you have a fixed navbar.

Screenshot 2023-10-07 at 22 53 13

### DIFF
--- a/packages/core/components/Puck/index.tsx
+++ b/packages/core/components/Puck/index.tsx
@@ -456,6 +456,7 @@ export function Puck({
                       <div
                         style={{
                           border: "1px solid var(--puck-color-grey-8)",
+                          transform: "scale(1, 1)",
                         }}
                       >
                         <Page data={data} {...data.root}>


JakeSidSmith:
> Can't believe I got this far in my career without knowing this trick.

I only found out about it because of the bugs it caused. 😂

chrisvxd:
@JakeSidSmith Unfortunately I've had to roll this back in our demo deployment as it's causing the drag positioning to break, which relies on position fixed. I missed this during review. You should be able to recreate locally if you're interested.

I haven't got a patch yet, but I'll likely have to revert this PR if we can't find a fix.

The only fix I can think of is to use JS to patch the transform position from react-beautiful-dnd, but that's hella brittle (see #123).

JakeSidSmith:
> I only found out about it because of the bugs it caused. 😂

Ironically at least one of the bugs was related to drag and drop.

I shall have a look and see if I can find an easy fix.

chrisvxd:
Sweet irony. Thanks @JakeSidSmith!

JakeSidSmith:
Doesn't appear that an easy fix is available.

I think this is something that could/will resolve itself if the drag n drop library is changed (as in #123), and or the preview is rendered within an iframe.

Should probably roll this back until then.

chrisvxd:
Thanks @JakeSidSmith. I'll revert this in an upcoming PR.

PR #141, opened 2023-10-07
Feature: restrict dragged/dropped components

### LINKED ISSUE #103: Support restricting DropZones by tag or component
A user should be able to restrict the components that can be dropped into a DropZone.

For example, the user may wish to restrict the root DropZone to layout components, and nested DropZones to leaf components.

Follow on from #87

### Option 1 - single `restrict`  prop

```tsx
// only allow ButtonGroup
<DropZone restrict={["ButtonGroup"]} />
```

* `restrict` might be unclear, so we might be better off naming this prop `allow`

### Option 2 - both `allow` & `disallow`

```tsx
// only allow ButtonGroup
<DropZone allow={["ButtonGroup"]} />

// allow everything except for Columns
<DropZone disallow={["Columns"]} />
```

* Restrict to either `allow` _or_ `disallow` but never both, because that creates strange boolean logic

## Considerations

* To be able to restrict the root DropZone, the user will need to render DropZone instead of `children` in their root `render` method. We will need to establish how they will define the `zone` in this scenario. Perhaps we should export it?

JakeSidSmith:
Resolves #103

Adds `allow` and `disallow` props to `DropZone`. Either of these can be an array of component names.

I have made `DropZone` generic so that you _can_ (but are not required to) restrict which component names can be provided in the `allow` and `disallow` props. This means that if/when you rename a component you are forced to update DropZones where it was referenced.

E.g.

```tsx
type Props = {
  ButtonGroup: ButtonGroupProps;
  Card: CardProps;
  Columns: ColumnsProps;
  Hero: HeroProps;
  Heading: HeadingProps;
  Flex: FlexProps;
  Logos: LogosProps;
  Stats: StatsProps;
  Text: TextProps;
  VerticalSpace: VerticalSpaceProps;
};

<DropZone<Props> zone={`column-${idx}`} allow={["ThisIsNotOkay"]} />
```

### DIFF
--- a/packages/core/components/DropZone/index.tsx
+++ b/packages/core/components/DropZone/index.tsx
@@ -139,7 +139,7 @@ function DropZoneEdit<Props extends Record<string, any> = Record<string, any>>({
   if (
     draggedItem?.draggableId &&
     allow &&
-    !allow.includes(draggedItem?.draggableId)
+    !allow.includes(draggedItem.draggableId)
   ) {
     isEnabled = false;
   }
@@ -147,7 +147,7 @@ function DropZoneEdit<Props extends Record<string, any> = Record<string, any>>({
   if (
     draggedItem?.draggableId &&
     disallow &&
-    disallow.includes(draggedItem?.draggableId)
+    disallow.includes(draggedItem.draggableId)
   ) {
     isEnabled = false;
   }
--- a/packages/core/components/DropZone/index.tsx
+++ b/packages/core/components/DropZone/index.tsx
@@ -13,14 +13,19 @@ const getClassName = getClassNameFactory("DropZone", styles);
 
 export { DropZoneProvider, dropZoneContext } from "./context";
 
-type DropZoneProps = {
+type DropZoneProps<Props extends Record<string, any> = Record<string, any>> = {
   zone: string;
-  allow?: readonly string[];
-  disallow?: readonly string[];
+  allow?: readonly (keyof Props)[];
+  disallow?: readonly (keyof Props)[];
   style?: CSSProperties;
 };
 
-function DropZoneEdit({ zone, allow, disallow, style }: DropZoneProps) {
+function DropZoneEdit<Props extends Record<string, any> = Record<string, any>>({
+  zone,
+  allow,
+  disallow,
+  style,
+}: DropZoneProps<Props>) {
   const ctx = useContext(dropZoneContext);
 
   const {
@@ -339,7 +344,9 @@ function DropZoneEdit({ zone, allow, disallow, style }: DropZoneProps) {
   );
 }
 
-function DropZoneRender({ zone }: DropZoneProps) {
+function DropZoneRender<
+  Props extends Record<string, any> = Record<string, any>
+>({ zone }: DropZoneProps<Props>) {
   const ctx = useContext(dropZoneContext);
 
   const { data, areaId = "root", config } = ctx || {};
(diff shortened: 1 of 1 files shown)


packages/core/components/DropZone/index.tsx
```diff
@@ -129,6 +136,22 @@ function DropZoneEdit({ zone, style }: DropZoneProps) {
     }
   }
 
+  if (
+    draggedItem?.draggableId &&
+    allow &&
+    !allow.includes(draggedItem.draggableId)
```
chrisvxd, line 142: `draggableId` is the component ID (i.e. `Heading-1245tgf31f1`), not the component type (`Heading`)

We'll need to use the `getItem()` lib to lookup the item data for the dragged item, and then get the type off there. Suggest we might want to do this outside of the DropZone render lifecycle (via hook or further up tree).

packages/core/components/DropZone/index.tsx
```diff
@@ -13,12 +13,19 @@ const getClassName = getClassNameFactory("DropZone", styles);
 
 export { DropZoneProvider, dropZoneContext } from "./context";
 
-type DropZoneProps = {
+type DropZoneProps<Props extends Record<string, any> = Record<string, any>> = {
```
chrisvxd, line 16: Can we use a different shape and name for the generic type here?

```tsx
type DropZoneProps<ValidComponents extends string[] = string[]> = {
```

Then folk can instantiate with `keyof Config.components`

chrisvxd (changes requested):
Awesome. Left some feedback, but I'm not totally sure on `allow` / `disallow` being the valid props here.

#103 discusses restricting by both component and tag, but the implementation here is only supporting tags.

Noting [your comment](https://github.com/measuredco/puck/issues/87#issuecomment-1751850944) on #87, we could potentially combine the APIs.

```tsx
<DropZone zone="my-zone" allow={Object.keys(config.components.Layout)} />
```

However, until we have a set direction on #87, we might want to hold off finalising this one.

packages/core/components/DropZone/index.tsx
JakeSidSmith: I was aiming to match the type of the props-mapping-type (for lack of a better term). So you can reuse the same type here as you do when defining your config.

We could make it instead:

```tsx
type DropZoneProps<T extends string> = {...etc};
```

And then you can provide the following to reuse the existing types:

```tsx
<DropZone<keyof Props> />
```

JakeSidSmith:
> #103 discusses restricting by both component and tag, but the implementation here is only supporting tags.

Is this, for example, to allow only components that return an `li` to be dropped inside a list?

JakeSidSmith:
> I'm not totally sure on `allow` / `disallow` being the valid props here.

Was hoping for some feedback on these. Depending on what exactly we restrict the children to the prop name will vary. We'd probably be better off with `allowedChildren` as an example - don't want it to be too specific e.g. `allowedComponents` as that'd rule out any future allow/disallowing.

packages/core/components/DropZone/index.tsx
chrisvxd: Yup I think that makes sense

packages/core/components/DropZone/index.tsx
chrisvxd: Yes, exactly. Further up the tree because the `getItem()` lib is a bit expensive and we might want it elsewhere. We could add do the lookup in <Puck> and add it to our (ever-growing) DropZone context.

chrisvxd:
> > https://github.com/measuredco/puck/issues/103 discusses restricting by both component and tag, but the implementation here is only supporting tags.
>
> Is this, for example, to allow only components that return an li to be dropped inside a list?

I completely misspoke in my first comment. Should say "implementation here is only supporting **components**".

By "tags" I mean the tagging implementation from #87.

chrisvxd:
Closing in favour of #257

PR #143, opened 2023-10-10
Reorder array items

### LINKED ISSUE #69: Ability to reorder array items
Would be great if subitems (like the features inside of a featureList) could be reordered via drag and drop. (Even if it's only sorting the list in the righthand sidebar.) As I understand it, the only way now to reorder sub items is to delete them and re-enter each one.

chrisvxd:
Please merge #154 first.

Closes #69.

## Description

This PR primarily:
* Enables drag-and-drop of array fields
* Refreshes the UI for array fields

It also:
* Reverts #140, since this is messing with drag position (preserved `fix:` type so we can spot for de-dupe on release)
* Refactors InputOrGroup into smaller files
* Adds missing default item props to the ButtonGroup demo
* Prevents style pollution for select fields
* Introduces a new internal Draggable component
* Introduces a new internal DragIcon component, which should be ported to the ComponentList (along with other UI refresh)

## Example

https://github.com/measuredco/puck/assets/985961/59cf4289-96e0-4031-bb91-a959c951907e

## Testing

1. Go to demo deployment
2. Create a new page
3. Add a ButtonGroup
4. Add a new button by clicking the "+"
5. Try reordering the buttons, changing the values and deleting them
6. Test again with undo/redo


packages/core/components/InputOrGroup/fields/ArrayField/index.tsx
```diff
@@ -0,0 +1,190 @@
...
+export const ArrayField = ({
+  field,
+  onChange,
+  value,
+  name,
+  label,
+}: InputProps) => {
+  const [valueWithIds, setValueWithIds] = useState<ItemWithId[]>(value);
+
+  // Create a mirror of value with IDs added for drag and drop
```
chrisvxd, line 33: This is clashing with undo/redo history 🤔

chrisvxd: Resolved in new upstream PR #154

packages/core/components/Puck/index.tsx
```diff
@@ -477,7 +477,6 @@ export function Puck({
                       <div
                         style={{
                           border: "1px solid var(--puck-color-grey-8)",
-                          transform: "scale(1, 1)",
```
monospaced, line 480: Assume this will get wiped out as already merged

monospaced (approved):

packages/core/components/Puck/index.tsx
chrisvxd: Yup - will rebase!

PR #146, opened 2023-10-11
fix: limit ExternalInput to only try to render properties that are strings

FreddieRidell:
it is currently possible for externals to be any kind of JSON data. Only some types can be rendered as react children. This MR checks the type of each property and only renders the ones that are strings

### DIFF
--- a/packages/core/components/ExternalInput/index.tsx
+++ b/packages/core/components/ExternalInput/index.tsx
@@ -1,4 +1,4 @@
-import { useEffect, useState } from "react";
+import { useMemo, useEffect, useState } from "react";
 import styles from "./styles.module.css";
 import getClassNameFactory from "../../lib/get-class-name-factory";
 import { Field } from "../../types/Config";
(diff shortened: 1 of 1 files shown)


packages/core/components/ExternalInput/index.tsx
```diff
@@ -18,6 +18,10 @@ export const ExternalInput = ({
   const [data, setData] = useState<Record<string, any>[]>([]);
   const [isOpen, setOpen] = useState(false);
   const [selectedData, setSelectedData] = useState(value);
+  const keys = useMemo(
+    () => Object.keys(data).filter((key) => typeof data[key] === "string"),
+    [data]
+  );
```
chrisvxd, line 24: Without this fix, I presume that Puck just crashes?

If we're keeping this fix, I think we should probably also support other renderable types, like `number`.

You could always filter the data returned in your adaptor, rather than here, although I suppose that means Puck can still break if you forget to do that.

chrisvxd (changes requested):
Thanks for the contribution @FreddieRidell! Left a comment

packages/core/components/ExternalInput/index.tsx
FreddieRidell: Yeah, it just crashes.

Wouldn't filtering the data out in the adaptor just remove it completly? I encountered this crash when using external `_data` with an `items` property that I wanted to be spread into my component. I want to keep that data, I just want to not render it.

Perhaps the best solution is to stop it crashing now, then investigate more powerful solutions as part of the issue I've created?

chrisvxd: That's a good point - yes it would remove it completely!

chrisvxd (approved):
🚀 Approved but holding off merge until we revert #140 (currently in #143, but will be split into separate PR)

Issue #18, opened 2023-06-14
Add more behaviour to plugins


chrisvxd:
* `setData` prop to allow plugins to override the data
* A mechanism to block page publishing (for example, if the heading outline is invalid in `plugin-heading-analyzer`)

johannbuscail:
Hey @chrisvxd, any update on this ? This is a feature that's really needed, especially the `setData` method. 
I'm using Puck with AI, and it would be a game changer to be able to modify the editor data even after init.

Thanks !

johannbuscail:
I think in general, the best way to access a `setData` function would be with a `useRef` hook, giving the dispatch function. I also think, all the functions such as `insert()`, `remove()`, etc.. should be exported and made available on the library.

Access to all those "behind the scenes" will make the library much more customizable.

PR #151, opened 2023-10-12
Enhancement(core/Puck): renderHeader

ahmedrowaihi:
# Problem
When customizing the header via ```renderHeader```, there is no way to control the Left Sidebar visibility
```jsx
// no way to control the header sidebar visibility
<Puck config={config} data={initialData} onPublish={save} renderHeader={({
 children, 
 data,
 dispatch 
 }) => ...} />
```

# Fix
- pass sidebar state toggle to the ```renderHeader```

usage example:
```diff
<Puck config={config} data={initialData} onPublish={save} renderHeader={({
 children, 
 data,
 dispatch,
+ toggleLeftSidebar
 }) => ...} />
```

### DIFF
--- a/packages/core/components/Puck/index.tsx
+++ b/packages/core/components/Puck/index.tsx
@@ -77,6 +77,7 @@ export function Puck({
   renderHeader?: (props: {
     children: ReactNode;
     data: Data;
+    toggleLeftSidebar: (visible?: boolean) => void;
     dispatch: (action: PuckAction) => void;
   }) => ReactElement;
   renderHeaderActions?: (props: {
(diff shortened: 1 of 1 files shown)


chrisvxd (changes requested):
Thanks for the ping @ahmedrowaihi! I did clock this but wanted to get through a bunch of other PRs first.

I actually have a slightly different proposal for this, now that #153 has been merged.

Can we expose the `dispatch` method to the `renderHeader` method instead of `toggleLeftSidebar`?

This would allow the user to control all data and UI state inside the `renderHeader` method, not just the side bar toggling.

We can also expose this to plugins too in #18.

chrisvxd:
Oh, so it is 😆 We might also need to expose the application state. FWIW, I'm about to rename a bunch of stuff in #158.

chrisvxd:
@ahmedrowaihi #158 is merged, so you can expose the appState via renderHeader (and remove `data`) if you need it.

PR #153, opened 2023-10-12
Record app state in reducer and undo/redo history

chrisvxd:
This PR refactors the data reducer to support application state (as well as data). This enables also application state to be added to the undo/redo history, which is necessary to resolve [this comment](https://github.com/measuredco/puck/pull/143/files#r1355826936), in which undo/redo history is corrupting the data.

Specifically, this PR adds:

* undo/redo history for the left hand side bar
* undo/redo history for the currently selected item


monospaced (approved):
This is great. One small comment feel free to ignore or spin out to a follow up refactor issue.

The naming `appData.data` and `appData.state` is a litte clunky to me. In particular `state` clashes with the reducer/React concept. And `Data.data` is repetitive. Wondering if `state` could be clarified to `uiState` or even `ui`? And perhaps `appData` could be `app`?

```js
app: {
 data: {},
 ui: {}
}
```

chrisvxd:
Agree with comment! Will do follow on after unblocking PR chain.

PR #156, opened 2023-10-17
fix: ensure layer tree consistently shows selected item

chrisvxd:
## Problem

The expansion state of the LayerTree component is based on the pathData data, which is set by components mounting and calling `registerPath`. However, the pathData object contained selectors with hardcoded indexes for where each item was in their zone. If the item moved, the pathData wouldn't be updated. This was noticed as the LayerTree would sometimes avoid expanding the parent area (i.e. Columns in the demo application) of a selected item because it couldn't find it.

This would not occur for components in the default-zone.

https://github.com/measuredco/puck/assets/985961/8bc051cf-da42-4d04-9ca6-320537ef6f1a

## Fix

This fix changes the underlying implementation of pathData, going from storing the selectors to the zoneCompound IDs (i.e. `[MyComponent:zone1,MyComponent:zone2]`), and calulating the selectors dynamically when necessary through a new `use-breadcrumbs` hook.

This hook isn't required for the LayerTree component, which relies on `is-child-of-zone`. `is-child-of-zone` can now easily determine whether or not the item is a child by checking for the presence of an area in the pathData.

This change also adds test coverage.

https://github.com/measuredco/puck/assets/985961/575a166b-08d8-4cda-8c5d-3f581361f592

### Test the layer tree

To test the original problem:

1. Create new page in demo app
2. Drag in Columns
3. Add Card to first column
4. Duplicate your Columns
5. Reorder your Columns, so the order is swapped
6. Select the Card in each column
7. Check that the LayerTree expands the Columns and shows the selected Card

### Test the breadcrumbs

Because this changes the underlying pathData model, we must also re-check the breadcrumbs.

1. Create new page in demo app
2. Drag in Columns
3. Add Flex inside the first column
4. Add Card to first Flex zone
5. Select the Card
6. Confirm that each breadcrumb works in the right-hand side bar
7. Confirm that the breadcrumbs appear the same as the [deployed demo](https://puck-demo-measured.vercel.app/)

### DIFF
--- a/packages/core/lib/__tests__/use-breadcrumbs.spec.tsx
+++ b/packages/core/lib/__tests__/use-breadcrumbs.spec.tsx
@@ -1,0 +1,67 @@
+import { DropZoneContext } from "../../components/DropZone/context";
+import { Config, Data } from "../../types/Config";
+import { Breadcrumb, convertPathDataToBreadcrumbs } from "../use-breadcrumbs";
+
+const item1 = { type: "MyComponent", props: { id: "MyComponent-1" } };
+const item2 = { type: "MyComponent", props: { id: "MyComponent-2" } };
+const item3 = { type: "MyComponent", props: { id: "MyComponent-3" } };
+
+const data: Data = {
+  root: { title: "" },
+  content: [item1],
+  zones: {
+    "MyComponent-1:zone": [item2],
+    "MyComponent-2:zone": [item3],
+  },
+};
+
+const config: Config = {
+  components: {
+    Comp: {
+      defaultProps: { prop: "example" },
+      render: () => <div />,
+    },
+  },
+};
+
+const dropzoneContext: DropZoneContext = {
+  data,
+  config,
+  pathData: {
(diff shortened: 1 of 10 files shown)


packages/core/lib/__tests__/use-breadcrumbs.spec.tsx
```diff
@@ -0,0 +1,78 @@
...
+      path: ["MyComponent-1:zone", "MyComponent-2:zone"],
+      label: "MyComponent",
+    },
+  },
+};
+
+describe("use-breadcrumbs", () => {
+  describe("convert-path-data-to-breadcrumbs", () => {
+    it("should convert path data to breadcrumbs", () => {
+      const result: Breadcrumb[] = [
```
chrisvxd, line 43: This needs removing

monospaced (approved):
Found the code in this one quite hard to parse. Not a fault of the code, I think we're just deep in the guts. But couldn't see anything obvious, and I did do some through testing in the deployed preview (comparinfg it to the deployed demo).

PR #157, opened 2023-10-17
fix: logic in ExternalInput key filtering

FreddieRidell:
My previous MR (https://github.com/measuredco/puck/pull/146) was broken.

Here's a working version


chrisvxd (approved):
I should have caught that. Thanks for the fix.

Issue #168, opened 2023-10-18
Long page title/path can overrun undo/redo controls


monospaced:
<img width="997" alt="Screenshot 2023-10-18 at 12 18 28" src="https://github.com/measuredco/puck/assets/98794/c9dceec8-3612-4c59-94a2-46ffcc2adc8c">

PR #174, opened 2023-10-19
fix: puck header style for #168

Yuddomack:
I slightly modified the style of the header area for #168

### DIFF
--- a/packages/core/components/Puck/index.tsx
+++ b/packages/core/components/Puck/index.tsx
@@ -318,7 +318,7 @@ export function Puck({
                             display: "grid",
                             padding: 16,
                             gridTemplateAreas: '"left middle right"',
-                            gridTemplateColumns: "288px auto 343px",
+                            gridTemplateColumns: "344px auto 344px",
                             gridTemplateRows: "auto",
                           }}
                         >
--- a/packages/core/components/Puck/index.tsx
+++ b/packages/core/components/Puck/index.tsx
@@ -318,7 +318,7 @@ export function Puck({
                             display: "grid",
                             padding: 16,
                             gridTemplateAreas: '"left middle right"',
-                            gridTemplateColumns: "288px auto 288px",
+                            gridTemplateColumns: "288px auto 343px",
                             gridTemplateRows: "auto",
                           }}
                         >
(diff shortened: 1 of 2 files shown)


packages/core/components/Puck/index.tsx
```diff
@@ -318,7 +318,7 @@ export function Puck({
                             display: "grid",
                             padding: 16,
                             gridTemplateAreas: '"left middle right"',
-                            gridTemplateColumns: "288px auto 288px",
```
chrisvxd, line 321: We should have both left and right areas the same width, otherwise the title text won't be entered.

Also suggest we should use the value `344px` instead of `343px` (I've tended to keep things as multiples of 4).

chrisvxd (changes requested):
Sorry for slow review but looks good @Yuddomack! One minor comment.

packages/core/components/Puck/index.tsx
```diff
@@ -318,7 +318,7 @@ export function Puck({
...
+                            gridTemplateColumns: "288px auto 344px",
```
chrisvxd, line 321: @Yuddomack thanks for updating, although you missed my other comment:

```suggestion
                            gridTemplateColumns: "344px auto 344px",
```

chrisvxd (changes requested):

packages/core/components/Puck/index.tsx
Yuddomack: oh I see, but can you explain what this (`otherwise the title text won't be entered.`) means? (To recognize context)

- 288px 344px 



- 344px 344px 



If the intention was to make the middle of both sides appear, I think I would need to calculate an additional width equal to the button?!

chrisvxd: Sorry that was a typo - "entered" should read "centred".

The second scenario looks correct for me. We may wish to look deeper into these styles in the future.

Yuddomack: ah centered! fix immediately 🔥

chrisvxd (approved):

chrisvxd:
Merged, thanks @Yuddomack!

Issue #139, opened 2023-10-06
Theming


chrisvxd:
Puck is being deployed in a wide variety of situations that require different themes. We should support theming to support this.

Some ideas:

* classNames for everything that can be customised
* style tags for everything that can be customised
* a discrete API exposing a limited subset of style overrides
* custom render functions to enable users to take ownership of specific parts of the application

Let's have a discussion and form this into a more fully fledged series of proposals.

Related #122

Leweyse:
Hi @chrisvxd !

I definitely support using custom properties. A great example of a similar implementation is `shadcn` (reference: [CSS properties](https://ui.shadcn.com/docs/theming)).

I use Tailwind CSS and would love it if the theme customization could be easily extended to Custom Fields or Plugins in my projects.

JakeSidSmith:
I think if you want to appease everyone you're probably going to need both inline `style` and `className` support.

And just to throw an extra complication into the works, there'll be people using [styled-components](https://styled-components.com/) that would like to be able to wrap your existing components instead of applying global styles. This might be able to be achieved in a similar way to how [react-markdown](https://www.npmjs.com/package/react-markdown#appendix-b-components) allows customizing elements....

If you had an interface that allowed overriding/extending the styles in one place (in addition to `style` and `className` on each component) it'd be incredibly powerful. E.g.

```tsx
noIdeaWhatToCallThis: {
  Button: (ExistingComponent) => ({className, props}) => (
    <ExistingComponent {...props} className={classNames(className, 'extra')} />
  ),
  // Styled component example
  Button: (ExistingComponent) => styled(ExistingComponent)`color: red;`
}
```

chrisvxd:
@Leweyse also a big fan of custom properties, so there could be something here. This might be a good way of doing a relatively light and controlled custom theming layer.

@JakeSidSmith I think what you're suggesting is the "custom render functions" option combined with "classNames", which is full UI override.

This is definitely the most powerful option and would allow people to do whatever they want, but I have concerns around the trade-off in making the internals part of a public API. If you remove a component internally, you now have to issue a breaking change. I imagine the risk of this happening decreases as Puck becomes more stable.

BleedingDev:
I agree with the simple solution with custom properties. They are well-known and supported and they can cover a lot of use-cases. Custom render functions would be for complete white-labelling, which is way to advanced (and heavily paid in solutions like Builder.io).

chrisvxd:
[...]

PR #175, opened 2023-10-19
Feat: theme customization

Leweyse:
# Theme Customization #139 

This is a proposal for a relatively light and controlled custom theming layer.

## Points to cover

* Provide the user with a set of custom properties that can be used to change Puck theme.
* Do not overwhelm the user with many options to customize.
* Control the options provided.

## Challenge

Color customization is not easy if we want to keep the custom theming layer light. For example:

To use different color shades or opacity values (eg: Dropzone is selected or hovered) there are two options: create multiple custom properties for each interaction (hover, selected, dragging over, etc.) or create one custom property that Puck internally will use with different [alpha values](https://developer.mozilla.org/en-US/docs/Web/CSS/alpha-value).

In this proposal I take the second option using [`hsl()`](https://developer.mozilla.org/en-US/docs/Web/CSS/color_value/hsl). It's been used for most components to handle different interactions: `ComponentListItem` and `Layer` use it for hover interactions, `Dropzone` and `DraggableComponent` use it for the different states to highlight the dropzone, the target, etc.

 ## Solution
 
 Default theme:

https://github.com/measuredco/puck/assets/69996340/87cb5ba7-d883-4293-8e00-dba617b0fb6a

Custom theme:

https://github.com/measuredco/puck/assets/69996340/322895f3-dec4-4682-8625-56037405b46d

Config used:

```css
:root {
  --puck-background: 229 22% 29%;
  --puck-foreground: 216 100% 98%;
  --puck-dropzone: 26 80% 73%;

  --puck-accent: 10 61% 78%;
  --puck-accent-foreground: 0 0% 0%;

  --puck-muted: 10 61% 78% / 0.05;
  --puck-muted-foreground: var(--puck-accent);

  --puck-popover: 229 22% 15%;
  --puck-popover-foreground: 216 100% 98%;

  --puck-border: 229, 22%, 50%;

  --puck-input: 0 0% 50%;
  --puck-input-alpha: 0.5;

  --puck-shadow: 229 22% 29%;

  --puck-radius: 0.5rem;
}
```

### DIFF
--- a/packages/core/components/Heading/styles.module.css
+++ b/packages/core/components/Heading/styles.module.css
@@ -4,6 +4,7 @@
   font-family: var(--puck-font-stack);
   font-weight: 700;
   margin: 0;
+  word-break: break-word;
 }
 
 .Heading b {
--- a/packages/core/components/Puck/index.tsx
+++ b/packages/core/components/Puck/index.tsx
@@ -401,7 +401,7 @@ export function Puck({
                             display: "grid",
                             padding: 16,
                             gridTemplateAreas: '"left middle right"',
-                            gridTemplateColumns: "288px auto 288px",
+                            gridTemplateColumns: "344px auto 344px",
                             gridTemplateRows: "auto",
                           }}
                         >


Leweyse:
Hi @chrisvxd!

This is not a complete proposal, but I'd like to know your opinion before moving forward.

This approach is quite challenging, specially if the goal is to provide more customization options.

chrisvxd:
Hey @Leweyse! This is awesome.

I'd say that I wasn't quite ready for us to jump into #139 (we typically marge issues as "Ready" when we're ready, but we haven't documented that process anywhere yet), but also appreciate it's easier to discuss things with code sometimes.

General points on this implementation:

1. Generally, I like CSS properties as an easy way to apply a theming layer, and it is compatible with some of the other options proposed in #139 (like full component overrides)
2. If you can try and keep _other_ CSS changes out of this PR, that would make it easier to review. I noticed you've changed a few pixel values, or added a shadow here and there.
3. We should avoid using `inherit` for colors because we don't know what other CSS is running on the page. We don't want the user's global CSS polluting the Puck CSS. This happens frequently. See #98, #107, #105.
4. We probably shouldn't be, but we use the Puck components inside the demo app. I'm noticing this in your preview frame. I don't think it really matters, because users won't be running the demo app.

> To use different color shades or opacity values (eg: Dropzone is selected or hovered) there are two options: create multiple custom properties for each interaction (hover, selected, dragging over, etc.) or create one custom property that Puck internally will use with different [alpha values](https://developer.mozilla.org/en-US/docs/Web/CSS/alpha-value).
> 
> In this proposal I take the second option using [hsl()](https://developer.mozilla.org/en-US/docs/Web/CSS/color_value/hsl). It's been used for most components to handle different interactions: ComponentListItem and Layer use it for hover interactions, Dropzone and DraggableComponent use it for the different states to highlight the dropzone, the target, etc.

I _think_ I would prefer the finer grained control, but I'm not totally sure. I imagine full control will probably be necessary for consumers to achieve on-brand color palettes, and accessibility. We could possibly use HSL (or [`color-mix`](https://developer.mozilla.org/en-US/docs/Web/CSS/color_value/color-mix)) as a fallback, if the user doesn't specify all the values.

Your CSS property suggestions are interesting. I need to think about this properly before giving a proper response because this is the most important piece of the proposal -- they essentially act as a public facing API and any renames become breaking changes.

Initial impressions:

* I think colors properties should all carry the `color` prefix, i.e. `puck-color-foreground`
* As suggested above, we might want to introduce more properties for different states (i.e. `puck-color-interactive`, `puck-color-interactive-hover`, `puck-color-interactive-focus`, `puck-color-button-text` for links and buttons)
* How would someone configure spacing and typography?

I'll probably have more thoughts on this, but wanted to share my first take. Would also be interested to hear what @monospaced thinks.

BleedingDev:
May I suggest LCH instead of HSL, because they keep the lightness as perceived by human? :)

https://caniuse.com/?search=lch

chrisvxd:
Great thanks @Leweyse! No rush on this one, just when you get the time. If you have to close it, we'll likely pick this up in due course.

PR #177, opened 2023-10-19
Add incremental static regeneration

chrisvxd:
~Please review #176 first~

## Problem

The original Next.js recipe produces dynamic routes. This often results in slow response times (500ms) as the server must re-render every Puck route at request time.

Next.js 13 supports incremental static regeneration. The existing `next` recipe implemented a single catch-all route for Puck, including the magic `/edit` behaviour - that is, you can add `/edit` to the end of any URL to spin up a Puck editor. 

We want to statically render all routes for the user, but not for the `/edit` route. Unfortunately Next.js 13 doesn't have a mechanism to statically render (or cache) parts of a catch-all route.

## Solution

This PR separates out the `/edit` functionality from the user-facing catch-all route to a separate `/puck` catch-all route. This means we can now use incremental static regeneration for the user-facing route (`[...puckPath]`) by enabling [`export const dynamic = "force-static";`](https://nextjs.org/docs/app/api-reference/file-conventions/route-segment-config#dynamic), and keep the editor route `/puck/[...puckPath]` dynamic.

Additionally, this PR introduces [middleware](https://nextjs.org/docs/app/building-your-application/routing/middleware) to retain the "magic" functionality of putting "/edit" on the end of any URL, rewriting it to the `/puck` catch-all route. It also disables direct access of `/puck`.

After the first render, pages now respond in ~50ms or less.

## Considerations

* This adds more files, but each file is doing a bit less
* We could maybe bundle this up
* Some users may not want to have static generation of all user-facing pages, as this strips headers and cookies. They can easily opt-out of this by removing the `export const dynamic = "force-static";` line from the catch-all router, but it seems like a sensible default.
* There's no upgrade path for previously generated apps - it has to be done manually


recipes/next/middleware.ts
```diff
@@ -0,0 +1,24 @@
...
+import type { NextRequest } from "next/server";
+
+export async function middleware(req: NextRequest) {
+  const res = NextResponse.next();
+
+  if (req.method === "GET") {
+    // Rewrite routes that match "/[...puckPath]/edit" to "/puck/[...puckPath]"
+    if (req.nextUrl.pathname.endsWith("/edit")) {
+      const pathWithoutEdit = req.nextUrl.pathname.split("/edit")[0];
+      const pathWithEditPrefix = `/puck${pathWithoutEdit}`;
```
monospaced, line 12: XSS vulnerability?

```suggestion
      const pathWithEditPrefix = `/puck${encodeURIComponent(pathWithoutEdit)}`;
```

recipes/next/README.md
```diff
@@ -30,10 +31,12 @@ You can do this for any route on the application, **even if the page doesn't exi
 
 To adopt this recipe you will need to:
 
-- **IMPORTANT** Add authentication to `/edit` routes. This can be done by modifying the example API routes in `/app/api/puck/route.ts` and server component in `/app/[...puckPath]/page.tsx`. **If you don't do this, Puck will be completely public.**
+- **IMPORTANT** Add authentication to `/edit` routes. This can be done by modifying the example API routes in `/app/api/puck/route.ts` and server component in `/app/edit/[...puckPath]/page.tsx`. **If you don't do this, Puck will be completely public.**
```
monospaced, line 34:
```suggestion
- **IMPORTANT** Add authentication to `/edit` routes. This can be done by modifying the example API routes in `/app/puck/api/route.ts` and server component in `/app/puck/[...puckPath]/page.tsx`. **If you don't do this, Puck will be completely public.**
```

recipes/next/app/puck/[...puckPath]/page.tsx
```diff
@@ -0,0 +1,39 @@
+/**
+ * This file implements a *magic* catch-all route that renders the Puck editor.
+ *
+ * This route exposes /edit/[...puckPath], but middleware.ts rewrites all URL requests
```
Comment (monospaced) [inline recipes/next/app/puck/[...puckPath]/page.tsx:4]:
```suggestion
 * This route exposes /puck/[...puckPath], but middleware.ts rewrites all URL requests
```

recipes/next/middleware.ts
```diff
@@ -0,0 +1,24 @@
...
+
+      return NextResponse.rewrite(new URL(pathWithEditPrefix, req.url));
+    }
+
+    // Disable "/edit/[...puckPath]"
```
monospaced, line 17:
```suggestion
    // Disable "/puck/[...puckPath]"
```

monospaced (changes requested):

recipes/next/README.md
chrisvxd: Fixed

recipes/next/middleware.ts
chrisvxd: Fixed

recipes/next/app/puck/[...puckPath]/page.tsx
Comment (chrisvxd) [reply recipes/next/app/puck/[...puckPath]/page.tsx]: Fixed

recipes/next/middleware.ts
monospaced: Would require `eval(pathWithoutEdit)` so not a concern in reality.

monospaced (approved):

PR #178, opened 2023-10-19
Add next-puck library

chrisvxd:
This is a proposal takes the latest `next` recipe from #177 and provides users with an upgrade path.

It does this by creating an opinionated library, `next-puck`. This library provides a CLI which the user uses _instead_ of the `next` CLI.

The user can call any `next` CLI command via `next-puck`, and `next-puck` will inject the necessary routes into their `/app` directory before calling the underlying `next` command. These files will be regenerated each time the `next-puck` CLI is used to invoke a `next` CLI command.

Assuming the user is either using the `next-minimal` recipe (or has followed the instructions and run `next-puck setup`), the injected routes will be ignored in the user's `.gitignore` file.

Additionally, the CLI generates a `middleware.ts` that imports directly from the `next-puck` library.

This provides an upgrade path - the user can upgrade the `next-puck` library to update the code in their routes and middleware.

## Configuration

To configure the library, the user can modify the `next-puck.config.tsx` file at the root of their Next project, implementing the `readPageServer` and `writePageServer` methods.

For more information, see the `next-puck` README provided in this PR.

## Considerations

* This is just a proposal and might not be the best way to approach this problem. In fact, the upgrade path might not be a real problem at all.
* We might want to make this the default, and remove the `next-minimal` recipe
* We should dog-food this on the Measured website before merging
* This approach could be reused for a batteries-included `vercel` recipe (including auth and database)


BleedingDev:
Will the part of CLI be asking which route should be editable? :) I would love to have something like `/` handled on my own and `/editor` be handled by puck (with all other paths like `/editor` for viewing the page and `/editor/edit` for editing the page. So one level deeper in this example (but I'd love to have it anywhere in the app).

chrisvxd:
@BleedingDev The CLI doesn't currently allow for that, but it could do.

Having said that, I'm unsure whether or not we're going to actually merge this in. In my testing, it didn't prove to be that useful - it's easier just to manage the Next routes manually. I found I almost immediately needed to make significant changes to the setup.

Issue #179, opened 2023-10-19
Array type props throwing error if undefined


sagarchoudhary96:
@chrisvxd  In the recent V0.10.0, i'm seeing an error which is being thrown if there is a props defined of type Array but if  that value is not present in props i.e it is undefined then it's throwing error when selecting that component. i belive this is what causing it -> https://github.com/measuredco/puck/blob/922a83913ae2cb402d0e8fa075cbcc0c906b61cc/packages/core/components/InputOrGroup/fields/ArrayField/index.tsx#L33C12-L33C29

<img width="269" alt="Screenshot 2023-10-20 at 4 43 23 AM" src="https://github.com/measuredco/puck/assets/16102594/fb919a68-54f5-43f7-b219-66ceb0c08c4d">

If we can add a check here for value to be defined or use empty array it should work. 

```
Array.from(value || [])
```

chrisvxd:
@sagarchoudhary96 please confirm if this is resolved in `0.10.1-canary.47ab3c9`

sagarchoudhary96:
@chrisvxd  tested it out and it fixes the issue.

chrisvxd:
Awesome, closing now. Thanks @sagarchoudhary96!

sagarchoudhary96:
@chrisvxd i missed one check in previous PR, ihave raised another PR for same.

PR #180, opened 2023-10-19
fix: use empty array when arrayfield value is undefined when creating new array

sagarchoudhary96:
#179


sagarchoudhary96:
@chrisvxd  i wasn't able to find any contributing guideline in the repo, so hope this follow the conventions, lemme know if any changes are needed, also would be great to have the Contributing guidelines present to know how to setup this project locally and make contribution to the project. If it's already there please do share the link. Thanks :)

chrisvxd (approved):
Nice! I'm going to merge and trigger a canary release. Please test that out in your environment and confirm that it works for your use case.

And @sagarchoudhary96, we don't have contributing guidelines yet but they are next on the list :)

PR #184, opened 2023-10-21
array field undefined value length check

sagarchoudhary96:


sagarchoudhary96:
#179

sagarchoudhary96:
@chrisvxd , sorry but there was one check which got missed in the previous fix. This should resolve the issue completely.

chrisvxd (approved):
Good catch, thanks again @sagarchoudhary96

sagarchoudhary96:
@chrisvxd  what is the canary version for this change.

chrisvxd:
`0.10.1-canary.68afd09`

They're automatically published on each merge with the commit, in this case `68afd09`. You should always be able to see it pop up a few minutes after merge on NPM: https://www.npmjs.com/package/@measured/puck?activeTab=versions

PR #187, opened 2023-10-25
Introduce dynamic props via resolveProps API

chrisvxd:
This PR that introduces new APIs for expanding the capability of adaptors

* The `resolveProp` config API allows users to modify props dynamically without writing them to the Puck data model on a per-component basis.
* The `resolveData` util function allows users to apply their dynamic props on the server or client before rendering their page with the `<Render>` component
* The `mapProp` adaptor API to enable the user to return change the shape of the data returned by the adaptor before applying it as a prop

Additionally:
* It removes the `_data` API for adaptors in favour of `resolveProp`. This is a breaking change.
* It makes the types stricter for Fields (considered a Fix).

## resolveProps Usage

The Hero component in the demo app contains an example

```tsx
const config:Config = {
  HeadingBlock: {
    fields: {
      text: {
        type: "text"
      }
    },
    resolveProps: async (props) => {
      // modify props. May include async operations.
      return { props: { ...props, text: "Hello, world" }, locked: { text: true } }
    },
    render: ({ text }) => {
      return <h1>{text}</h1>
    }
  }
}
```

Pairing it with an adaptor

```tsx
const quotesAdaptor = {
  name: "Quotes API",
  fetchList: async () => {
    const quotesResponse = await fetch("/api/quotes");
    
    return quotesResponse.json();
  }
};

const config:Config = {
  HeadingBlock: {
    fields: {
      quote: {
        type: "external",
        adaptor: quotesAdaptor
      },
      text: {
        type: "text"
      }
    },
    resolveProps: async (props) => {
      if (props.quote?.id) {
        const quote = await fetch(`/api/quotes/${props.quote.id}`);
        
        return { props: { text: quote.json().content }, locked: { text: true } }
      }
      
      return { props, locked: { text: false } };
    },
    render: ({ text }) => {
      return <h1>{text}</h1>
    }
  }
}
```

## resolveData usage

This should be used by the user to resolve data before rendering their <Render> component. It should ideally be used server-side, but can also be used client side in a `useEffect` hook.

```tsx
export const MyRSCPage =  async () => {
  const data = await getPage(); // user's database query
  const resolvedData = await resolveData(data);

  return <Render data={data} />
}
```

## mapProp usage

```tsx
const quotesAdaptor = {
  name: "Quotes API",
  fetchList: async () => {
    const quotesResponse = await fetch("/api/quotes");
    const data =  quotesResponse.json();

    return { id: data.id, title: data };
  },
  // Called _after_ the user selects an item from the adaptor
  mapProp: (prop => ({ id: prop.id }))
};

const config:Config = {
  HeadingBlock: {
    fields: {
      quote: {
        type: "external",
        adaptor: quotesAdaptor
      },
      text: {
        type: "text"
      }
    },
    render: ({ text }) => {
      return <h1>{text}</h1>
    }
  }
}
```

## Tasks

* [x] add tests

### DIFF
--- a/README.md
+++ b/README.md
@@ -500,6 +500,7 @@ External fields can be used to load content from an external content repository,
 - **adaptor** (`Adaptor`): Content adaptor responsible for fetching data to show in the table
   - **name** (`string`): The human-readable name of the adaptor
   - **fetchList** (`(adaptorParams: object) => object`): Fetch a list of content and return an array
+  - **mapProp** (`(selectedItem: object) => object`): Map the selected item into another shape
 - **adaptorParams** (`object`): Paramaters passed to the adaptor
 
 ### Custom Fields
--- a/README.md
+++ b/README.md
@@ -454,17 +454,57 @@ The `Config` object describes which components Puck should render, how they shou
 
 A `Field` represents a user input field shown in the Puck interface.
 
-- **type** (`text` | `textarea` | `number` | `select` | `radio` | `external` | `array` | `custom`): The input type to render
+### All Fields
+
 - **label** (`text` [optional]): A label for the input. Will use the key if not provided.
-- **arrayFields** (`object`): Object describing sub-fields for items in an `array` input
-  - **[fieldName]** (`Field`): The Field objects describing the input data for each item
-- **getItemSummary** (`(object, number) => string` [optional]): Function to get the name of each item when using the `array` or `external` field types
-- **defaultItemProps** (`object` [optional]): Default props to pass to each new item added, when using a `array` field type
-- **options** (`object[]`): array of items to render for select or radio inputs
+
+### Text Fields
+
+- **type** (`"text"`)
+
+### Textarea Fields
+
+- **type** (`"textarea"`)
+
+### Number Fields
+
+- **type** (`"number"`)
+
+### Select Fields
+
+- **type** (`"select"`)
+- **options** (`object[]`): array of items to render
   - **label** (`string`)
   - **value** (`string` | `number` | `boolean`)
-- **adaptor** (`Adaptor`): Content adaptor if using the `external` input type
+
+### Radio Fields
+
+- **type** (`"radio"`)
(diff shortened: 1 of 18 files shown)


README.md
```diff
@@ -275,22 +434,70 @@ The `Config` object describes which components Puck should render, how they shou
...
+- **defaultItemProps** (`object` [optional]): Default props to pass to each new item added, when using a `array` field type
+
+### External Fields
+
+External fields can be used to load content from an external content repository, like Strapi.js, using an `Adaptor`.
+
+- **type** (`"external"`)
+- **adaptor** (`Adaptor`): Content adaptor responsible for fetching data to show in the table
+  - **name** (`string`): The human-readable name of the adaptor
+  - **fetchList** (`(adaptorParams: object) => object`): Fetch a list of content and return an array
```
monospaced, line 494: Could Adaptors fetch an object instead of a list?

packages/core/components/DropZone/context.tsx
```diff
@@ -5,7 +5,7 @@ import {
   useCallback,
   useState,
 } from "react";
-import { Config, Data } from "../../types/Config";
+import { Config, Data, MappedItem } from "../../types/Config";
```
monospaced, line 8: imported but not used?

monospaced (approved):
Couple of quetions, approving for expediency

README.md
chrisvxd: They could fetch an object but this method needs to return an array to populate the table

monospaced: Wonder if "a list of content" is correct in that case? Maybe `Fetch ~a list of~ content and return an array`? I could be missing the point here 😊

chrisvxd: I've updated to `Fetch content from a third-party API and return an array` for clarity

packages/core/components/DropZone/context.tsx
chrisvxd: Addressed

PR #189, opened 2023-10-26
💄 Transitions: ComponentList, titleIcon, LeftSideBar

ahmedrowaihi:
small satisfactions


chrisvxd:
@ahmedrowaihi this has been on my list to review, sorry for the slow response! Feel free to leave it open, and I'll get to it soon.

I'm still not convinced now is the right time for animations, but happy to give feedback anyway

Issue #166, opened 2023-10-18
React Server Components


chrisvxd:
Users have reported issues when trying to use Puck with React Server Components (RSC).

Specifically, users wish to use RSC for their individual components.

We don't have a proposal for tackling this yet, but please use the comments to report your findings.

4leite:
I am interested in contributing to this.

Here is a working example app without modification to core using RSC for individual components: https://github.com/4leite/puck/tree/server-render-demo

Some thoughts:
This implementation passes a 'DropZone' prop to the component. When editing, we pass the DropZone component from core. When previewing or server rendering, we pass the [render.tsx](https://github.com/4leite/puck/blob/server-render-demo/apps/demo/app/%5B%5B...puckPath%5D%5D/render.tsx) server component instead.

Nextjs static analysis doesn't like the "render" field on the config object (or any functions for that matter), so we need to move this out of the block component files. It makes sense to collocate the fields and default props with the component so rather than moving the whole object, we export the partial ComponentConfig and compose it with a helper function.

RSC doesn't have access to context, instead we pass the data as props to render.tsx. We are piggybacking on the defaultProps of the item, and using 'withDropZone' to conditionally render nested DropZones. 'withDropZone' should probably live in 'item' rather than 'item.props'.

A side effect of this implementation is that it may partially solve #71 as the dropzones are no longer rendered when not editing. The flipside is there may be differences in the render for edit and preview.

Passing the DropZone also provides an avenue for implementing #103 by passing a custom DropZone wrapper.

NB: The implementation above also instantiates a [libsql](https://github.com/tursodatabase/libsql) database using a postinstall script. This is to facilitate testing the server side render and publish defined in [actions.tsx](https://github.com/4leite/puck/blob/server-render-demo/apps/demo/app/%5B%5B...puckPath%5D%5D/actions.ts) Deploying this to production would require a persistent file system, or (better) a proper database connection.

chrisvxd:
@4leite Passing DropZones in as props is a really interesting proposal (and quite obvious, now you mention it). Some thoughts:

* This could easily be baked into `<Render>` as part of core
* We might want to consider naming it using render props convention (`renderDropZone`) rather than as a component (`DropZone`)
* I don't think we need to have a `withDropZone` boolean. We can pass `renderDropZone` (or `DropZone`) to all components.
* We might be able to avoid a breaking change, since whilst importing DropZone from the package won't work with RSC, it would still work for non RSC builds.
[...]

PR #192, opened 2023-10-27
 refactor(apps/demo): improve rsc with example server actions and preview

4leite:
added "use client" directives
added server actions with static db
added preview mode for client side data
moved client only code out of config file

#166

### DIFF
--- a/apps/demo/app/[...puckPath]/client.tsx
+++ b/apps/demo/app/[...puckPath]/client.tsx
@@ -9,6 +9,7 @@ import headingAnalyzer from "@measured/puck-plugin-heading-analyzer/src/HeadingA
 import config from "../../config";
 import { useLocalData } from "./hooks";
 import { publishPageData } from "./actions";
+import { join } from "path";
 
 export function Client({
   path,
@@ -52,7 +53,7 @@ export function Client({
           headerPath={path}
           renderHeaderActions={() => (
             <>
-              <Button href={`${path}/preview`} newTab variant="secondary">
+              <Button href={join(path, "/preview")} newTab variant="secondary">
                 Preview page
               </Button>
             </>
--- a/apps/demo/app/[...puckPath]/client.tsx
+++ b/apps/demo/app/[...puckPath]/client.tsx
@@ -6,29 +6,23 @@ import { Render } from "@measured/puck/components/Render";
 import { useEffect, useState } from "react";
 import { Button } from "@measured/puck/components/Button";
 import headingAnalyzer from "@measured/puck-plugin-heading-analyzer/src/HeadingAnalyzer";
-import config, { initialData } from "../../config";
+import config from "../../config";
+import { useLocalData } from "./hooks";
+import { publishPageData } from "./actions";
 
-const isBrowser = typeof window !== "undefined";
+export function Client({
+  path,
+  data: serverData,
+  isEdit,
+}: {
+  data: Data;
+  path: string;
+  isEdit: boolean;
+}) {
+  const [localData, setLocalData] = useLocalData(path);
 
-export function Client({ path, isEdit }: { path: string; isEdit: boolean }) {
-  // unique b64 key that updates each time we add / remove components
-  const componentKey = Buffer.from(
-    Object.keys(config.components).join("-")
-  ).toString("base64");
-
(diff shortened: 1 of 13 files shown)


apps/demo/config/root.tsx
```diff
@@ -2,34 +2,16 @@ import { ReactNode } from "react";
...
-        fontWeight: isActive ? "600" : "400",
-      }}
-    >
-      {label}
-    </a>
-  );
-};
-
-function Root({ children, editMode }: RootProps) {
+function Root({ children, suffix = "" }: RootProps) {
```
chrisvxd, line 14: How is suffix set?

apps/demo/app/[...puckPath]/client.tsx
```diff
@@ -52,15 +47,14 @@ export function Client({ path, isEdit }: { path: string; isEdit: boolean }) {
         <Puck
           config={config}
           data={data}
-          onPublish={async (data: Data) => {
-            localStorage.setItem(key, JSON.stringify(data));
-          }}
+          onChange={(data) => setLocalData(data)}
+          onPublish={(data) => publishPageData(path, data)}
```
Comment (chrisvxd) [inline apps/demo/app/[...puckPath]/client.tsx:51]: I don't know if having a separate localStorage-powered "preview" route and separate published route makes sense here, because the publish route can never be updated by the user.

The point of the demo app is to act like a sandbox for a how real application might work, and the fake "publishing" workflow is important to that.

_Having said that, we have been thinking about an official Puck website. The thought crossed my mind that we could make the homepage the demo application, in which case a preview mode might make sense (but this isn't something I want to solve now)_

chrisvxd (commented):
This is really interesting, thanks for sharing! I haven't done a full review (and noted this is in draft), but left a few comments.

It might also be better to update the `next` recipe (under `/recipes`) as this actually feeds the generator. We might consider updating the demo app to use the `next` recipe if that makes sense, too.

apps/demo/app/[...puckPath]/client.tsx
Comment (4leite) [reply apps/demo/app/[...puckPath]/client.tsx:51]: see my comments below, but happy to revert to the local storage solution

4leite:
> This is really interesting, thanks for sharing! I haven't done a full review (and noted this is in draft), but left a few comments.

Thanks for your time, yes I wanted to get your thoughts before committing to this.

> It might also be better to update the `next` recipe (under `/recipes`) as this actually feeds the generator. We might consider updating the demo app to use the `next` recipe if that makes sense, too.

Happy to do either or both. My motivation is to have the demo app be able to test that any changes to support rsc are working, with the minimal changes. I realise this may conflict with the purpose of the demo app.

It sounds like.
1) The demo should showcase the publish flow, primarily deployed and hosted by you, with visitors able to try publishing changes that are visible only to them. In which case the trivial solution is localstorage and implementing any kind of server side in this context is non trivial.
2) The generator should create a mvp puck editor. It seems like the tricky thing here is that any viable implementation of puck requires server side storage - probably a database connection - and my suspicion is we don't want to be opinionated about that. This means it won't be immediately deployable.

Does that sound accurate? 

Assuming the above, then making these changes to the app generator makes sense to me, with the intention that in the future the demo would be generated (maybe with a localstorage provider option) .

Are you interested in some or all of the following:
1) merging this PR with changes including reverting the view / publish flow to localstorage.
2) me creating a separate pr with similar changes to the generator, but supporting a persistent storage flow
3) A follow up generator app PR with my currently working local database solution [libsql](https://turso.tech/libsql).

The problem is there is currently no way to verify in the project that puck/core will work with server rendering, which makes verifying any changes I make impossible, and it's not really solvable with automated tests either without introspecting the nextjs code which sounds like hard work.

apps/demo/config/root.tsx
4leite: It's not at the moment

BleedingDev:
@4leite 
I think you can check if RSC and Server Actions works. It is based on the version of React, therefore you can do `import { version } from "react"`, if it contains `canary` string you have support for Server Actions (which are not yet in Stable React, but are in Stable Next.js). It is probably the best way to check this, because even `Modern.js` supports (or soon will) RSC and Server Actions are in negotiation as well.

4leite:
Closing in favor of #199

PR #197, opened 2023-11-01
feat: support headless render for react server components

4leite:
add `puckCtx: { DropZone }` to props for blocks.
add PuckComponent type
include puckCtx in existing DropZoneEdit and DropZoneRender
add rsc.js export to package

see #166 

The core package is a cjs, which means that any "use client" needs to be at the top of the bundle and marks the entire module as client code.
Regardless, as there is client side code in the module, next will complain when trying to import this module into a server component.
Options:
1). Provide a separate cjs module within in this package without any client code 
2). Provide a separate package with the above
3). Provide an esm export with "preserve modules" and correct banner, eg https://esbuild.github.io/api/#banner
4). Do not provide an rsc renderer,  client to implement their own - would require exposing "setupZone" function

I have chosen 1) in this PR.


4leite:
Closing in favor of #199

4leite:
Closing in favor of #199

chrisvxd:
Thanks @4leite! This is in my queue to review soon. Sorry for slow response.

Feel free to join us [on discord](https://discord.gg/D9e4E3MQVZ) if you want to chat more in-depth about this, too.

packages/core/components/ServerRender/index.tsx
```diff
@@ -0,0 +1,82 @@
...
+        editMode={false}
+        id={"puck-root"}
+      >
+        <DropZoneRender config={config} data={data} zone={rootDroppableId} />
+      </config.root.render>
+    );
+  }
+
+  return <DropZoneRender config={config} data={data} zone={rootDroppableId} />;
+}
```
chrisvxd, line 82: This needs prettier formatting before linting will succeed

packages/core/rsc.tsx
```diff
@@ -0,0 +1 @@
+export { Render } from "./components/ServerRender";
```
chrisvxd, line 1: This needs prettier formatting before linting will succeed

4leite:
> ... but I wonder if we should consider a breaking change to the `DropZone` API, enforcing the use of `renderDropZone` instead of `<DropZone>` and updating the existing `Render` component to avoid context.

As implemented it's opt in, but yes I feel like deprecating the existing Dropzone api and removing it later is a good option. I'm interested if anyone has any use case for the context rendered Dropzone that this doesn't solve.
 
> I'm not ready to make a call on this yet, but would value any input on how that might work (especially regarding the modules).

There's no way around having separate modules for the rsc renderer that I can see using cjs. It's maybe doable but complicated with esm and preserve modules. The issue is it's one file, and nextjs sees the hooks and other client code in the file so it marks it as needing "use client" above it in the tree. Exporting the renderer as part of the main module will still work fine for those not trying to optimise for static server rendered content.

packages/core/components/DropZone/index.tsx
```diff
@@ -175,6 +175,7 @@ function DropZoneEdit({ zone, style }: DropZoneProps) {
                 const defaultedProps = {
                   ...config.components[item.type]?.defaultProps,
                   ...item.props,
+                  puckCtx: { renderDropZone: DropZone },
```
chrisvxd, line 178: I find `puckCtx` a bit awkward, and think `ctx` (aka "context") is a bit of a misnomer.

What about simply `puck`?

```suggestion
                  puck: { renderDropZone: DropZone },
```

chrisvxd (changes requested):
Thanks for the clarity @4leite. I've been thinking about this some more and think your proposal is the best option.

Going with ESM doesn't really buy us anything, as we still need to split the output into separate files. That's the only way RSC is going to behave, so file-denominated bundling is the way to go.

We can do a follow on to deprecate the legacy DropZone API, and capture the breaking change in #202.

I've left a few comments before we merge.

packages/core/types/Config.tsx
```diff
@@ -98,12 +99,36 @@ export type Content<
...
+  fields?: Fields<ComponentProps>;
+  resolveData?: (
+    data: DataShape,
+    params: { changed: Partial<Record<keyof ComponentProps, boolean>> }
+  ) =>
+    | Promise<Partial<ComponentDataWithOptionalProps<ComponentProps>>>
+    | Partial<ComponentDataWithOptionalProps<ComponentProps>>;
+};
+
+type RootComponentConfig<
```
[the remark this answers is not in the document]
4leite: I actually created this to resolve an issue I had - was unaware RootComponentConfig was ever in main.
But it highlighted a bug a missed -  the implementation did not support using renderDropZone in a root component.

Have removed the type difference and fixed the bug.

packages/core/components/Render/index.tsx
```diff
@@ -1,33 +1,15 @@
...
-          title={title}
-          editMode={false}
-          id={"puck-root"}
-        >
-          <DropZone zone={rootDroppableId} />
-        </config.root.render>
-      </DropZoneProvider>
-    );
-  }
-
```
4leite, line 27: @chrisvxd  Render now uses the ServerRender component, just wrapping in context for backwards compatibility

4leite:
hmm, looks like the server render is not working in the demo app. Will have a look, but might need to walk back the change to use ServerRender in the Render component

chrisvxd (approved):
I'm thinking we may actually want to keep DropZones as-is (rather than deprecating), but we don't need to make that decision here. Going to merge. Awesome stuff, @4leite!

chrisvxd:
@4leite seems like it needs a rebase before I can merge

PR #199, opened 2023-11-02
feat: add example rsc with nested dropzone to next generator

4leite:
Depends on #197
See #166


chrisvxd:
Thanks @4leite! This is in my queue to review soon. Sorry for slow response.

Feel free to join us [on discord](https://discord.gg/D9e4E3MQVZ) if you want to chat more in-depth about this, too.

4leite:
@chrisvxd  Thanks for merging #197 
This is update to reflect the change on that PR and is ready for review.

chrisvxd (commented):
Thanks @4leite, although I'm not sure this PR is quite right and perhaps not necessary. The next recipe is designed to show a minimal example of how to implement Puck in a Next environment with best practices. DropZones aren't core usage, so I don't think we should include them (or modify the config at all), but rely on the docs.

The only change I think we could consider making is to modify the catch-all page to use the RSC Render component instead of the client one [here](https://github.com/measuredco/puck/blob/main/recipes/next/app/%5B...puckPath%5D/page.tsx#L42), but I'm also happy with this RSC being left as opt-in via docs for the time being.

chrisvxd:
Closing for the time being.

PR #218, opened 2023-11-12
👌 IMPROVE(create-puck-app): enhancements

ahmedrowaihi:
- clean unnecessary awaits
- initialize git optionally
- prevent Windows's git crash due to commits with single quotes

### DIFF
--- a/packages/create-puck-app/index.js
+++ b/packages/create-puck-app/index.js
@@ -103,7 +103,7 @@ program
       return;
     }
 
-    await fs.mkdirSync(appName);
+    fs.mkdirSync(appName);
 
     const packageManager = !!options.useNpm
       ? "npm"
@@ -130,7 +130,7 @@ program
       let data;
 
       if (path.extname(filePath) === ".hbs") {
-        const templateString = await fs.readFileSync(filePath, "utf-8");
+        const templateString = fs.readFileSync(filePath, "utf-8");
 
         const template = Handlebars.compile(templateString);
         data = template({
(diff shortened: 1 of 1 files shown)


packages/create-puck-app/index.js
```diff
@@ -77,10 +77,17 @@ program
         required: true,
         default: "next",
       },
+      {
+        type: "confirm",
+        name: "initializeGit",
+        message: "Do you want to initialize and commit using Git?",
+        default: true,
+      },
```
chrisvxd, line 85: I'd still prefer to auto initialize, as I believe this is pretty standard in most create-* projects (although happy to be proven wrong).

packages/create-puck-app/index.js
```diff
@@ -155,26 +162,30 @@ program
...
+          execSync("git init", { cwd: appPath, stdio: "inherit" });
+
+          execSync("git add .", { cwd: appPath, stdio: "inherit" });
+          execSync('git commit -m "build(puck): generate app"', {
+            cwd: appPath,
+            stdio: "inherit",
+          });
+        } catch (error) {
+          console.log("Failed to commit git changes");
+        }
```
chrisvxd, line 187: Good addition!

chrisvxd (changes requested):
Awesome, thanks @ahmedrowaihi! I think I'd prefer to keep automatic git initialisation, rather than add a prompt. Please see my comment!

chrisvxd (approved):
🚀

PR #230, opened 2023-11-15
Improve responsive behaviour

### LINKED ISSUE #225: Improve responsive behaviour
Puck currently has several responsive issues making it difficult to use on mobile.

Ideally, Puck will be fully responsive. The underlying drag-and-drop implementation is already mobile friendly, so there are mainly CSS issues outstanding.

monospaced:
Closes #225 

First pass of responsive improvements

* Editor supports viewport widths down to `638px` without breaking
* Pragmatic widths for the editor page preview (helps alleviate some of the pain when editing responsive pages in a preview area narrower than the reported viewport width)
* Tightens up spacing in editor sidebars to maximise available space for longer names
* Ensures long strings can always be handled (defensive CSS)
* Prevents form zoom in iOS
* Adds some missing responsive behaviour to the demo app

For follow up work, suggest we make more granular issues, I can this handle as I go, for example

* Add a toggle for the right-hand sidebar
* Support viewports below `638px` wide

### Before

Screenshot 2023-11-16 at 12 33 17

### After

Screenshot 2023-11-16 at 12 33 40

### DIFF
--- a/packages/core/components/InputOrGroup/styles.module.css
+++ b/packages/core/components/InputOrGroup/styles.module.css
@@ -48,7 +48,7 @@
   border-color: var(--puck-color-grey-8);
   border-radius: 4px;
   font-family: inherit;
-  padding: 12px 16px;
+  padding: 12px 15px;
   width: 100%;
 }
 
--- a/packages/core/components/Puck/styles.module.css
+++ b/packages/core/components/Puck/styles.module.css
@@ -1,0 +1,151 @@
+/*
+ * Puck's responsive layout uses minimum viewport widths slightly _below_ common
+ * framework/device breakpoints, and ensures that the width of the resulting
+ * Puck page preview (zoomed at 75%) is slightly _above_ common framework/device
+ * breakpoints. This can help alleviate some of the pain when editing responsive
+ * pages in a preview area that is narrower than the reported viewport width.
+ *
+ * Viewport | Puck page @ zoom 0.75
+ * --------------------------------
+ * 638px    | 321px
+ * 766px    | 321px
+ * 990px    | 604px
+ * 1022px   | 646px
+ * 1198px   | 801px
+ * 1398px   | 1025px
+ * 1598px   | 1212px
+ */
+
+ .Puck {
+  --puck-frame-width: minmax(266px, auto);
+  --puck-side-bar-width: minmax(186px, 250px);
+  --puck-space-px: 16px;
+  bottom: 0;
+  display: grid;
+  grid-template-areas: "header header header" "left editor right";
+  grid-template-columns: 0 var(--puck-frame-width) var(--puck-side-bar-width);
+  grid-template-rows: min-content auto;
+  height: 100vh;
+  left: 0;
+  overflow-x: auto;
+  position: fixed;
+  right: 0;
+  top: 0;
+}
+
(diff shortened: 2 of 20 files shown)


packages/core/components/InputOrGroup/styles.module.css
```diff
@@ -48,7 +48,8 @@
   border-color: var(--puck-color-grey-8);
   border-radius: 4px;
   font-family: inherit;
-  padding: 12px 16px;
+  font-size: 16px;
+  padding: 10.25px 15px;
```
chrisvxd, line 52: Where's this fractional value coming from?

packages/core/components/InputOrGroup/styles.module.css
```diff
@@ -48,7 +48,8 @@
   border-color: var(--puck-color-grey-8);
   border-radius: 4px;
   font-family: inherit;
-  padding: 12px 16px;
+  font-size: 16px;
```
chrisvxd, line 51: I find this size a bit uncomfortable on desktop. If this is required for mobile, can we restrict by viewport?

packages/core/components/Puck/styles.module.css
```diff
@@ -0,0 +1,151 @@
...
+ * Viewport | Puck page @ zoom 0.75
+ * --------------------------------
+ * 638px    | 321px
+ * 766px    | 321px
+ * 990px    | 604px
+ * 1022px   | 646px
+ * 1198px   | 801px
+ * 1398px   | 1025px
+ * 1598px   | 1212px
+ */
```
chrisvxd, line 17: This is very smart 👏

chrisvxd (changes requested):
This is fantastic! Barely any comments.

Also happy for you to pick up the suggested follow-ons.

packages/core/components/InputOrGroup/styles.module.css
monospaced: To match the existing height with the new `font-size`

packages/core/components/InputOrGroup/styles.module.css
monospaced: Not really, as it's specifically an iOS touch thing, so should not be conflated with viewport.

chrisvxd: This is the only change I'm unsure on. The font-size feels way too chunky on desktop viewports to me now, in particular because the input font size is larger than the surrounding body font size.

image

I expect desktop to be the primary medium and would rather not compromise the consistency at this time.

How would you feel about either:

1. Leaving the font-size as-is and letting iOS zoom,
2. Only changing the font-size on smaller viewports, but accept that it will zoom on larger iOS devices (perhaps conflation, but I think that's an okay compromise) or
3. Prevent iOS from zooming with [`maximum-scale=1`](https://stackoverflow.com/a/46254706/1703260)

monospaced: I'll remove for now, cause not directly related to the PR, but suspect we'll need to return to this!

monospaced: Done

chrisvxd: Ty!

chrisvxd (approved):

PR #242, opened 2023-11-21
Fixes from responsive work

chrisvxd:
Introduces some fixes for bugs introduced in #225.

Measured site header in latest canary:

image


packages/core/components/Button/Button.module.css
```diff
@@ -8,9 +8,9 @@
   align-items: center;
   gap: 8px;
   letter-spacing: 0.05ch;
-  font-family: var(--puck-font-family);
+  font-family: var(--puck-font-family-proportional);
```
monospaced, line 11: soz!

packages/core/components/Puck/index.tsx
```diff
@@ -436,13 +436,13 @@ export function Puck({
...
                             </div>
-                            <div>
+                            <>
                               {renderHeaderActions &&
                                 renderHeaderActions({
                                   state: appState,
                                   dispatch,
                                 })}
-                            </div>
+                            </>
```
monospaced, line 445: In this case we should wrap the <button> in the demo `renderHeaderActions` in a `<div>` instead of fragment `<>`, as otherwise the `<button>` is a direct flex child which cuases some layout issue in some responsive scenarioas (e.g. very long title text and/or slug).

monospaced (changes requested):

packages/core/components/Puck/index.tsx
monospaced: Here https://github.com/measuredco/puck/blob/6a56af016f0813e3a7eaea9c2d2448359709e93a/apps/demo/app/%5B...puckPath%5D/client.tsx#L63

chrisvxd: I've squashed that in

apps/demo/puck.config.tsx
monospaced, line 1: Should this have been deleted?

chrisvxd: Yup - added an additional commit. It's unused.

monospaced (approved):

PR #243, opened 2023-11-21
chore: stop dnd scoll container warning

chrisvxd:
Having `overflow-x: auto` on a parent of the Droppable causes the following dnd warning whenever dragging an item:

> dnd.esm.js:31 @hello-pangea/dndDroppable: unsupported nested scroll container detected.A Droppable can only have one scroll parent (which can be itself)Nested scroll containers are currently not supported.We hope to support nested scroll containers soon: https://github.com/atlassian/react-beautiful-dnd/issues/131

See https://github.com/atlassian/react-beautiful-dnd/issues/131\#issuecomment-1634398431

Introduced in #225, which has not been released so using `chore`.

### DIFF
--- a/packages/core/components/Puck/styles.module.css
+++ b/packages/core/components/Puck/styles.module.css
@@ -16,7 +16,7 @@
  * 1598px   | 1212px
  */
 
- .Puck {
+.Puck {
   --puck-frame-width: minmax(266px, auto);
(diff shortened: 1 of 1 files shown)


chrisvxd:
@monospaced I'm sure this one has implications for what you're working on. There might be another way to fix it, but currently unclear.

monospaced (approved):

PR #246, opened 2023-11-22
Improve narrow viewport support 2

### LINKED ISSUE #232: Improve support for viewports below 638px wide
Currently you can side scroll, so not completely broken, but not great either.

Thoughts

* Allow toggling of right-hand sidebar (this has value beyond narrow screen support - if we can hide both sidebars, this allows full-width previewing of pages in the editor)
* On narrow screens
  * Right hand sidebar could be full-width, don't have to see preview at same time as editing fields
  * Both sidebars could be hidden by default 
  * Elements of navigation could collapse into hamburger (probably the undo-redo, user actions and publish button)
  * Biggest challenge is the left sidebar drag-drop into preview (could potentaily retain side scroll for this, needs testing)

monospaced:
Close #232

### DIFF
--- a/packages/core/components/IconButton/IconButton.module.css
+++ b/packages/core/components/IconButton/IconButton.module.css
@@ -1,11 +1,11 @@
 .IconButton {
+  align-items: center;
   background: transparent;
   border: none;
+  border-radius: 4px;
   color: var(--puck-color-grey-3);
   display: flex;
-  align-items: center;
   justify-content: center;
-  border-radius: 4px;
   padding: 4px;
 }
 
@@ -13,4 +13,14 @@
   background: var(--puck-color-grey-9);
   color: var(--puck-color-blue);
   cursor: pointer;
+}
+
+.IconButton-title {
+  clip: rect(0 0 0 0);
+  clip-path: inset(100%);
+  height: 1px;
+  overflow: hidden;
+  position: absolute;
+  white-space: nowrap;
+  width: 1px;
 }
--- a/packages/core/components/IconButton/IconButton.module.css
+++ b/packages/core/components/IconButton/IconButton.module.css
@@ -3,7 +3,7 @@
   background: transparent;
   border: none;
   border-radius: 4px;
-  color: var(--puck-color-grey-3);
+  color: currentColor;
   display: flex;
   justify-content: center;
   padding: 4px;
--- a/packages/core/components/MenuBar/styles.module.css
+++ b/packages/core/components/MenuBar/styles.module.css
@@ -1,0 +1,47 @@
+.MenuBar {
+  background-color: var(--puck-color-white);
+  border-bottom: 1px solid var(--puck-color-grey-8);
+  display: none;
+  left: 0;
+  margin-top: 1px;
+  padding: 8px 16px;
+  position: absolute;
+  right: 0;
+  top: 100%;
+  z-index: 2;
+}
+
+.MenuBar--menuOpen {
+  display: block;
+}
+
+@media (min-width: 638px) {
+  .MenuBar {
+    border: none;
+    display: block;
+    margin-top: 0;
+    overflow-y: visible;
+    padding: 0;
+    position: static;
+  }
+}
+
+.MenuBar-inner {
+  align-items: center;
+  display: flex;
+  flex-wrap: wrap;
+  gap: 8px 16px;
+  justify-content: flex-end;
+}
+
+@media (min-width: 638px) {
+  .MenuBar-inner {
+    display: flex;
+    flex-direction: row;
+    flex-wrap: nowrap;
+  }
+}
+
+.MenuBar-history {
+  display: flex;
+}
--- a/packages/core/components/Puck/index.tsx
+++ b/packages/core/components/Puck/index.tsx
@@ -16,7 +16,7 @@ import { Plugin } from "../../types/Plugin";
 import { usePlaceholderStyle } from "../../lib/use-placeholder-style";
 
 import { SidebarSection } from "../SidebarSection";
-import { Globe, Sidebar, ChevronLeft, ChevronRight } from "react-feather";
+import { ChevronDown, ChevronUp, Globe, Sidebar } from "react-feather";
 import { Heading } from "../Heading";
 import { IconButton } from "../IconButton/IconButton";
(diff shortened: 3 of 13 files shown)


packages/core/components/IconButton/IconButton.module.css
```diff
@@ -14,3 +14,13 @@
   color: var(--puck-color-blue);
   cursor: pointer;
 }
+
+.IconButton-title {
```
chrisvxd, line 18: Change is good but the commit mentions `InputButton`. Should be `IconButton`.

packages/core/components/IconButton/IconButton.module.css
```diff
@@ -3,7 +3,7 @@
   background: transparent;
   border: none;
   border-radius: 4px;
-  color: var(--puck-color-grey-3);
+  color: currentColor;
```
chrisvxd, line 6: This unexpectedly causes the icon colour to change in some scenarios (such as the array item action overlay). Curious to know the reason?

Fix might be to change the color of the array item action overlay wrapper.

packages/core/components/Puck/index.tsx
```diff
@@ -273,6 +273,52 @@ export function Puck({
...
+        });
+      }
+    };
+
+    window.addEventListener("resize", handleResize);
+
+    return () => {
+      window.removeEventListener("resize", handleResize);
+    };
+  }, []);
```
chrisvxd, line 317: *chef's kiss*

packages/core/components/Puck/index.tsx
```diff
@@ -273,6 +273,52 @@ export function Puck({
...
+      sidebar === "left" ? "rightSideBarVisible" : "leftSideBarVisible";
+
+    dispatch({
+      type: "setUi",
+      ui: {
+        [`${sidebar}SideBarVisible`]: !sideBarVisible,
+        ...(!widerViewport ? { [oppositeSideBar]: false } : {}),
+      },
+    });
+  };
```
chrisvxd, line 287: Perf: could use `useCallback` here to avoid recreating the function on each render

packages/core/components/Puck/styles.module.css
```diff
@@ -119,13 +155,14 @@
...
   grid-area: editor;
-  overflow-y: auto;
+  overflow: auto;
   position: relative;
 }
 
 .Puck-root {
   box-shadow: 0 0 0 calc(var(--puck-space-px) * 2) var(--puck-color-grey-10);
   margin: var(--puck-space-px);
+  min-width: 321px;
```
chrisvxd, line 182: I think the minimum should be bigger, to serve iPhone ~12 dimensions by default (488px or whatever it is). Makes a big difference to me when testing.

packages/core/components/Puck/styles.module.css
```diff
@@ -35,9 +35,33 @@
...
+
+@media (min-width: 458px) {
+  .Puck {
+    --puck-frame-width: minmax(266px, auto);
+  }
+}
+
+@media (min-width: 638px) {
+  .Puck {
+    --puck-side-bar-width: minmax(186px, 250px);
```
chrisvxd, line 61: I wonder if we should actually introduce this a viewport down, at 458px, so the inputs are a bit more usable. I think the inputs are more important than the preview when editing.

chrisvxd (changes requested):
This is a brilliant improvement! Left some comments.

General feedback:

* I half expected the right sidebar to automatically open after I tapped the component
* Perhaps we should actually grey the preview out with an overlay when the side bar is open (on narrow viewports)
* Should tapping the greyed out preview hide the right sidebar?

These could probably all be fixed in post since this is already a huge upgrade.

packages/core/components/MenuBar/styles.module.css
```diff
@@ -0,0 +1,46 @@
...
+    position: static;
+  }
+}
+
+.MenuBar-inner {
+  display: flex;
+  flex-wrap: wrap;
+  gap: 8px 16px;
+  justify-content: flex-end;
+}
```
chrisvxd, line 34: We should also add `align-items: center`, so the use can add items of any size using `renderHeaderActions`

(fwiw, our secondary Button is also not rendering at the same height as the primary one, which is how I noticed this)

packages/core/components/MenuBar/styles.module.css
```diff
@@ -0,0 +1,46 @@
...
+  border-bottom: 1px solid var(--puck-color-grey-8);
+  display: none;
+  left: 0;
+  margin-top: 1px;
+  padding: 8px 16px;
+  position: absolute;
+  right: 0;
+  top: 100%;
+  z-index: 1;
+}
```
chrisvxd, line 12: It's possible to render the overlay actions on top of this. Can probably be fixed in post.

image

packages/core/components/Puck/styles.module.css
monospaced: Mostly agree, but will require some unpicking, so suggest we do in post.

packages/core/components/IconButton/IconButton.module.css
monospaced: I wanted the sidebar toggle icons to have a different a colour depending if the sidebar is open or not, this was not possible with a hardcoded colour for the `IconButton`, so switched to inherit from parent. In general inherit is a more composable pattern.

Will update array field action 👍🏻

packages/core/components/MenuBar/styles.module.css
monospaced: Fixed

packages/core/components/MenuBar/styles.module.css
monospaced: Done

packages/core/components/IconButton/IconButton.module.css
monospaced: Done

packages/core/components/Puck/styles.module.css
monospaced: Spun out to #247

packages/core/components/Puck/styles.module.css
monospaced: Spun out to #247

chrisvxd (approved):

Issue #226, opened 2023-11-14
Add type object  in fields


prashanthvdckap:
Add type object  in fields so that we can have key pair fields

chrisvxd:
Thanks @prashanthvdckap. This is a duplicate of #79.

chrisvxd:
Hi @prashanthvdckap. This was delivered in #256 and should be available in `0.13.0-canary.b95050d`. Please let me know if that works for you.

Issue #142, opened 2023-10-10
Support dot notation for fields


chrisvxd:
#60 made the request for a new `object data type for more complex, nested data shapes to be managed by Puck fields.

This was proposed in PR #62, but it was decided that the `array` interface might not be appropriate for this, and that we are conflating 1) UI concerns and 2) data model concerns.

Consequently, we closed #60/#62 and opened #64 to handle 1) the UI rendering by introducing a React-ish `renderFields` API.

### Option 1 - Nested paths

#64 handles the UI-layer for rendering complex objects. Additionally, we could allow for reaching into nested objects via the field API

```tsx
export const Block: ComponentConfig<BlockProps> = {
  fields: {
    "data.label": {  type: "text", },
  },
 // Optionally render the fields  
  renderFields: () => {
    return <>
      <Fieldset title="Data">
        <Field name="data.label" />
      </Fieldset>
    </>
  },
}
```

### Option 2 - object field type

The original proposal from #62, we could implement an `object` field type that reuses the array API and UI.

```tsx
export const Block: ComponentConfig<BlockProps> = {
  fields: {
    data: {
      type: "object",
      // Collapse `arrayFields` into `fields`?
      fields: {
        label: { type: "text" }
      },
    },
  },
}
```

### Option 3 - combo

We could also offer both APIs and allow the user to determine what is best for their use-case.

## Considerations

* How to type nested paths

chrisvxd:
Object fields where implemented in https://github.com/measuredco/puck/pull/256

Issue #64, opened 2023-09-10
RFC: Add `fieldset` support to fields API


chrisvxd:
Currently, we have no way to create groups of fields. #62 adds `object` support, and spun out [this discussion](https://github.com/measuredco/puck/pull/62#pullrequestreview-1618841059).

> Inline fieldsets at the top-level (with headings) are something I've considered as a way to break the form up into separate areas of related fields. It strikes me that the Heading scenario here might also work with an inline Heading fieldset. 
 

### Option 1

Add `renderFields` method

Example adding a **Style** fieldset to Heading

```tsx
export const Heading: ComponentConfig<HeadingProps> = {
  renderFields: () => {
    return <>
      <Field name="text" />
      <Field name="level" />
      <Fieldset title="Style">
        <Field name="size" />
        <Field name="align" />
        <Field name="padding" />
      </Fieldset>
    </>
  },
  fields: {
    text: { type: "text" },
    size: {
      type: "select",
      options: sizeOptions,
    },
    level: {
      type: "select",
      options: levelOptions,
    },
    align: {
      type: "radio",
      options: [
        { label: "Left", value: "left" },
        { label: "Center", value: "center" },
        { label: "Right", value: "right" },
      ],
    },
    padding: { type: "text" },
  },
}
```

### Option 2

Change `fields` API or add new `fieldsets` API

```tsx
export const Heading: ComponentConfig<HeadingProps> = {
  fieldsets: [
    {
      title: "",
      fields: {
        text: { type: "text" },
        level: {
          type: "select",
          options: levelOptions,
        },
      },
    },
    {
      title: "Style",
      fields: {
        size: {
          type: "select",
          options: sizeOptions,
        },

        align: {
          type: "radio",
          options: [
            { label: "Left", value: "left" },
            { label: "Center", value: "center" },
            { label: "Right", value: "right" },
          ],
        },
        padding: { type: "text" },
      },
    },
  ],
```

### Option 3

Something else; could be some combination of the above or another proposal.

---

Related #62

Danm72:
This is the way we currently model our page data, maybe its worth also considering multiple tabs within the right panel e.g content, styles, etc

On the above Option 2 definitely reads better to me. Option 1 sort of feels like a ceremony that can be skipped.
https://github.com/measuredco/puck/assets/1926968/ad9c3e97-512d-43e7-95ef-e729db571b2f

chrisvxd:
Going with `renderFields`. Marking as ready for dev.

[...]

PR #256, opened 2023-11-24
Add object fields

### LINKED ISSUE #79: Defining a props as an Object.
i have a Component with props as 
```
{
 name: string;
price: {"total": number; "string": number}
}
```
@chrisvxd 
how can i define the config for the price prop for puck editor. i see we added support for "custom" type but not sure if this can be done with it .

chrisvxd:
This has come up numerous times (#60, #79, #226).

Whilst I still consider Fieldsets/Tabs (#64) and dot-notation (#142) good ways to solve for this, adding the `object` field type is much a lower effort way of supporting most of these use cases. These two APIs can also exist in parallel.

Closes #79.

### DIFF
--- a/packages/core/components/InputOrGroup/fields/ObjectField/styles.module.css
+++ b/packages/core/components/InputOrGroup/fields/ObjectField/styles.module.css
@@ -1,0 +1,17 @@
+/**
+ * ObjectField
+ */
+
+.ObjectField {
+  display: flex;
+  flex-direction: column;
+  background-color: white;
+  border: 1px solid var(--puck-color-grey-8);
+  border-radius: 4px;
+}
+
+.ObjectField-fieldset {
+  border: none;
+  margin: 0;
+  padding: 16px 15px;
+}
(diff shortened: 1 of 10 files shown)


packages/core/components/InputOrGroup/fields/ObjectField/styles.module.css
```diff
@@ -0,0 +1,33 @@
...
+.ObjectFieldItem {
+  background: white;
+  border-top-left-radius: 4px;
+  border-top-right-radius: 4px;
+  display: block;
+  margin-bottom: 1px; /* Simulate border with margin so we can hide it whilst dragging */
+
+  border-bottom: 0;
+  outline-offset: 0px !important; /* Important helps to override Nextra docs */
+  outline: 1px solid var(--puck-color-azure-6) !important; /* Important helps to override Nextra docs */
```
monospaced, line 26: These styles are not being applied, because the component never uses `getClassNameItem()`. In the docs site the CSS module class is `undefined`. So perhaps this is the issue, and these `!important` aren't needed?

monospaced (changes requested):

packages/core/components/InputOrGroup/fields/ObjectField/styles.module.css
monospaced: Screenshot 2023-11-27 at 12 05 59

chrisvxd: Good catch! Lazy copy pasta from array fields. I removed the whole concept of the `ObjectFieldItem` here, which included these unused styles.

PR #260, opened 2023-11-24
Add search to external fields

### LINKED ISSUE #203: Add search UI to external fields
External fields currently show data in a tabular form. It does not provide a mechanism for searching or filtering data.

We should add UI to:

1. Search the user's external data, passing a `query` parameter to the `fetchList` function
2. Filter the user's external data, allowing the user to customise the filters available and passed to the `fetchList` function under `filters` or similar
3. Paginate the user's external data. This may require a new API or breaking change to the `fetchList` function.
4. Add a button to "refresh" data, calling the user's `fetchList` function.

## Considerations

* Not all data is tabular. Should we also allow people to override the UI using a render method, or push people towards `custom` field types?

chrisvxd:
This PR adds a search input to external fields with the new `showSearch` and `initialQuery` APIs added to the external field type.

https://github.com/measuredco/puck/assets/985961/53b6a864-b09d-4faa-a4cc-745b1f936d69

It is a partial implementation of #203. I chose to leave out pagination (due to complexity) and filtering (unclear how common a use-case this would be) for the time-being. We can track those in new tickets.

Closes #203

### DIFF
--- a/packages/core/components/ExternalInput/index.tsx
+++ b/packages/core/components/ExternalInput/index.tsx
@@ -1,11 +1,12 @@
-import { useMemo, useEffect, useState } from "react";
+import { useMemo, useEffect, useState, useCallback } from "react";
 import styles from "./styles.module.css";
 import getClassNameFactory from "../../lib/get-class-name-factory";
 import { ExternalField } from "../../types/Config";
-import { Link, Unlock } from "react-feather";
+import { Link, Search, Unlock } from "react-feather";
 import { Modal } from "../Modal";
 import { Heading } from "../Heading";
 import { ClipLoader } from "react-spinners";
+import { Button } from "../Button";
 
 const getClassName = getClassNameFactory("ExternalInput", styles);
 const getClassNameModal = getClassNameFactory("ExternalInputModal", styles);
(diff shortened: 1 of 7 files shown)


packages/core/components/ExternalInput/index.tsx
```diff
@@ -100,13 +113,44 @@ export const ExternalInput = ({
...
+              <form
+                className={getClassNameModal("searchForm")}
+                onSubmit={(e) => {
+                  e.preventDefault();
+
+                  search(searchQuery);
+                }}
+              >
+                <label className={getClassNameModal("search")}>
+                  <span style={{ visibility: "hidden", width: 0 }}>Search</span>
```
monospaced, line 134: This will be hidden from assistive technology, and neither thew feather icon of the placeholder will provide an accessible alternative.

Best approach would be to use this CSS snippet here instead, which would make the label text inclusively available.

```
  clip: rect(0 0 0 0);
  clip-path: inset(100%);
  height: 1px;
  overflow: hidden;
  position: absolute;
  white-space: nowrap;
  width: 1px;
```

packages/core/components/ExternalInput/index.tsx
```diff
@@ -100,13 +113,44 @@ export const ExternalInput = ({
...
+                  <div className={getClassNameModal("searchIcon")}>
+                    <Search size="18" />
+                  </div>
+                  <input
```
monospaced, line 138: Could maybe use `type="search"` (and follow the convention of `name=q`), see https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/search

packages/core/components/ExternalInput/index.tsx
```diff
@@ -147,12 +191,14 @@ export const ExternalInput = ({
                 })}
               </tbody>
             </table>
-          </div>
 
-          <div className={getClassNameModal("noContentBanner")}>No content</div>
+            <div className={getClassNameModal("loadingBanner")}>
+              <ClipLoader size={24} />
```
monospaced, line 196: This needs an `aria-label="Loading"`

monospaced (changes requested):

packages/core/components/ExternalInput/index.tsx
chrisvxd: Addressed all of these comments. Also added a new commit to fix some style leakage in Nextra.

monospaced (approved):
ty!

PR #263, opened 2023-11-24
Add libs for prop migration

### LINKED ISSUE #41: Add mechanism for prop migration
## Problem

_Source: https://news.ycombinator.com/item?id=37392861_

Sometimes a user might need to rename or remove a property provided to a component. For example, remaining `description` to `subtitle`.

The user currently has two options:

1. Manually migrate their saved data before updating their Puck configuration
2. (Preferred) Retain the old prop and map it to the new prop inside the `render` function.

These solutions either require repetitive work, or require the user to republish each page manually.

## Proposal

Add a first-party mechanism for transforming data on a per-component basis.

This would probably live under a separate package (`/packages/transform` or `/packages/migrate`) and provide a user with a mechanism to  1) patch their data on the fly in a standardised manor and 2) batch transform their puck data consistently to adhere to the new props.

### Create transformer

```tsx
import { createTransform } from "@measured/puck-transform";

const transform = createTransform({
  HeadingBlock: ({ subtitle, ...props }) => {
    return {
      ...props,
      description: subtitle,
    };
  },
});
```

### Patch data inline

```tsx
// Pull from backend
const data = {
  type: "HeadingBlock",
  props: {
    subtitle: "Hello, world",
  },
};

export default () => <Puck data={transform(data)} config={config} />;
```

### Patch all existing data

```tsx
// Example data
const allData = [
  {
    path: "/",
    data: {
      root: {},
      content: [
        {
          type: "HeadingBlock",
          props: {
            subtitle: "Hello, world",
          },
        },
      ],
    },
  },
];

// Run the transformer
const patchedData = allData.map((page) => ({
  ...page,
  data: transform(page.data),
}));

// Write to your db
const save = () => {};
save(patchedData);
```

## Considerations

* TypeScript - How to tightly manage the types?

chrisvxd:
* Provide a user-facing `transform` API to enable the user to map new props for their components
* Provide internal `transformData` API to enable Puck to update the data model without impacting most users. The exception to this is that the user may change to modify the data outside of Puck, after Puck has handed it over. 
* Tighten up types to prevent data mapping issues from occurring, and reduce the number of conditions Puck has to consider internally

## Tasks

* [ ] Add docs

Closes #41


apps/docs/pages/docs/integrating-puck/_meta.json
```diff
@@ -5,5 +5,7 @@
   "categories": {},
   "dynamic-props": {},
   "external-data-sources": {},
-  "server-components": {}
+  "server-components": {},
+  "prop-migration": {},
+  "data-migration": {}
```
chrisvxd, line 10: Just remembered - these need adding to the TOC on the Getting Started page

monospaced (approved):

apps/docs/pages/docs/integrating-puck/_meta.json
chrisvxd: No, already added to TOC!

PR #264, opened 2023-11-29
fix: fix initial UI state on mobile

chrisvxd:
A race condition with the data resolvers meant that the UI state was being reset after the sidebars were collapsed. This is fixed by avoiding UI state updates on the initial resolveData call.


monospaced (approved):
🙏🏻

PR #265, opened 2023-11-29
Composable puck

### LINKED ISSUE #224: Make Puck more composable
We should consider supporting composable layouts for Puck, enabling more sophisticated implementations.

For example, as part of the docs site, we are reimplementing Puck with a custom layout for interactive examples. This should be supported OOTB.

### Default - no children

If no children are passed to `<Puck>`, the default layout will be used.

### Render children

When the user defines children, the user can take over the Puck layout.

```tsx
export const Editor = () => {
  return <Puck>{({ appState, dispatch }) => (
    <>
      <Puck.Preview />
      <Puck.Form />
    </>
  )}</Puck>
}
```

Should also support direct children, instead of render children

```tsx
export const Editor = () => {
  return <Puck>
    <Puck.Preview />
    <Puck.Form />
  </Puck>
}
```

### usePuck hook
Which can be combined with a `usePuck` hook to access the context

```tsx
export const MyComponent = () => {
  const { dispatch } = usePuck();

  return <button onClick={() => dispatch()}>Click me</button>
}
```

chrisvxd:
Merge blocked by #41, but ready for review regardless.

## Description

This PR

* adds compositional Puck via `children`
* adds a `usePuck` hook for building custom components within compositional Puck
* adds an `overrides` prop to override specific parts of the interface, replacing `renderHeader` and `renderHeaderActions`
* adds a `ui` prop to set the initial UI state
* adds an `isDragging` boolean in the UI state
* makes `onPublish` optional
* removes the `renderComponentList` prop (breaking change)
* updates the plugin API to align with the `overrides` functionality (breaking change, but not listed as `feat` so won't appear in changelog - might want to revisit)

👉 [See an example page](https://puck-demo-git-composable-puck-measured.vercel.app/custom-ui/edit) with custom overrides 👈 

Rather than reimplement the docs here, I suggest reading the [new Extending Puck section](https://puck-docs-git-composable-puck-measured.vercel.app/docs/extending-puck/custom-interfaces) in the docs.

Additional reading:
* [Updated components API reference](https://puck-docs-git-composable-puck-measured.vercel.app/docs/api-reference/components)
* [New Overrides API reference](https://puck-docs-git-composable-puck-measured.vercel.app/docs/api-reference/overrides)
* [New `usePuck` API reference](https://puck-docs-git-composable-puck-measured.vercel.app/docs/api-reference/functions/use-puck)
* [Updated Plugin API reference](https://puck-docs-git-composable-puck-measured.vercel.app/docs/api-reference/plugin)

## Tasks

- [x] Rename `customUi` to `skin`, `outfit` or `theme`?
- [x] Rename `Puck.fields` to `Puck.form`?
- [x] Document
- [x] Add children to overrides.headerActions for plugin currying
- [x] Ensure Plugin API supports `customUi.fields`
- [x] Consider adding an explicit `feat` for breaking the plugin API
- [x] Finalise naming for `appState` (vs `state`) in `usePuck` hook
- [x] Remove padding and border from Puck fields by default?
- [x] Add experimental warning to overrides?

Closes #224


packages/core/components/Puck/components/Components/index.tsx
```diff
@@ -0,0 +1,18 @@
...
+import { useAppContext } from "../../context";
+import { ComponentList } from "../../../ComponentList";
+import { useMemo } from "react";
+
+export const Components = () => {
+  const { config, state, customUi } = useAppContext();
+
+  const componentList = useComponentList(config, state.ui);
+
+  const Wrapper = useMemo(() => customUi.componentList || "div", [customUi]);
```
chrisvxd, line 11: This component is called `<Puck.Components />`, but the `customUi` API refers to `componentList`, as do the other exported components `ComponentListDroppable` and `ComponentListDraggable`.

This might be fine, because the former is intended for composability and componentList is the internal name, but we may want to align somehow.

apps/docs/pages/docs/api-reference/actions.mdx
```diff
@@ -1,8 +1,6 @@
 # Actions
 
-> **⚠️ The plugin API is highly experimental and is likely to experience breaking changes.**
-
-Actions enable you to make changes to the internal [AppState](/docs/api-reference/app-state) via the [`dispatch` method](functions#dispatch).
+Actions enable you to make changes to the internal [AppState](/docs/api-reference/app-state) via the a dispatcher.
```
monospaced, line 3: typo "the a"

apps/docs/pages/docs/api-reference/actions.mdx
```diff
@@ -35,5 +36,28 @@ Change a value on Puck's [UI state](/docs/api-reference/app-state/#ui).
...
+| ------- | ----------------------------- | ----------------------------------------- | -------- |
+| `type`  | `type: "setData"`             | "setData"                                 | Required |
+| `state` | `state: { data: {}, ui: {} }` | [AppState](/docs/api-reference/app-state) | Required |
+
+#### Example
+
+```tsx
+dispatch({
+  type: "set",
+  data: { data: {}, ui: {} },
```
monospaced, line 61: Should be `state:` ?

packages/core/components/Puck/components/Fields/index.tsx
```diff
@@ -0,0 +1,199 @@
...
+  children: ReactNode;
+  isLoading: boolean;
+  itemSelector?: ItemSelector | null;
+}) => {
+  return (
+    <div className={getClassName()}>
+      {children}
+      {isLoading && (
+        <div className={getClassName("loadingOverlay")}>
+          <ClipLoader />
```
monospaced, line 37: `aria-label="Loading"`

monospaced (changes requested):

apps/docs/pages/docs/api-reference/actions.mdx
chrisvxd: Yes, good catch

packages/core/components/Puck/components/Fields/index.tsx
chrisvxd: Added to all ClipLoaders. Should probably spin this out to a component at some point.

apps/demo/app/custom-ui/[...puckPath]/client.tsx
```diff
@@ -0,0 +1,356 @@
...
+          paddingLeft: 16,
+          paddingRight: 16,
+          borderBottom: "1px solid #ddd",
+          overflowX: "auto",
+        }}
+      >
+        {tabs.map((tab, idx) => {
+          const isCurrentTab = currentTab === idx;
+          return (
+            <div
```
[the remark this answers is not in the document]
Comment (chrisvxd) [reply apps/demo/app/custom-ui/[...puckPath]/client.tsx]: Switched to button and added `cursor: pointer` to the inline CSS. Still no hover styles because I didn't want to create a stylesheet for the time-being, but this is an improvement.

monospaced (approved):

PR #227, opened 2023-11-14
Feat/remix recipe

EarthlingDavey:
Add remix recipe - I've only touched the `remix/recipe` folder. So I don't know if there's extra stuff I need to do to prepare the package for release.

It's working fine - but with some TS errors like : https://github.com/measuredco/puck/issues/185

Do you mind reviewing, and advising the best way to fix the errors.

I can also remove or amend the devcontainer if needed.

Thanks

PR #266, opened 2023-11-30
🐛 FIX(Recipes/Remix): explicit exports to fix paths links/metas

ahmedrowaihi:
This PR assures that the routing catch is working
I noticed a bug in previous implementations ( using ```export * from ``` ) where remix was not able to detect the functions in all $puckPaths  so I just exported them explicitly and tested all sites and they are working by this PR

### DIFF
--- a/recipes/remix/app/getPuckPath.ts
+++ b/recipes/remix/app/getPuckPath.ts
@@ -1,0 +1,17 @@
+// puck is behaving weirdly between loaders/actions, so this is a workaround assuers that the path is always correct
+export function getPuckPath({
+  params,
+  request,
+}: {
+  params: any;
+  request: Request;
+}) {
+  let referer = request.headers.get("referer") || "http://localhost:3000";
+  let puckPath =
+    params.puckPath || new URL(referer).pathname.replace("/edit", "");
+
+  // Get path, and default to slash for root path.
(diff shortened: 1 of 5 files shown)


EarthlingDavey:
Hey @ahmedrowaihi thanks for sharing the PR. I worked on the Remix recipe and the bug must have slipped through.

Do you mind sharing the minimum steps to reproduce?

chrisvxd:
Thanks @ahmedrowaihi. As @EarthlingDavey said, please can you provide steps to reproduce? It worked during my testing of #227, but it's possible something was missed.

EarthlingDavey:
Is it possibe that the bug was happening when you were on a nested route like `/slash1/slash2/edit`. The Remix recipe doesn't support nested routes (yet).

EarthlingDavey:
Hi @ahmedrowaihi ok, I understand. 

For some more context... the Remix recipe was released only a few days ago, so there may be other undiscovered bugs too. The good news is that I doubt it will have made it to production anywhere - so we can take our time to discover and apply fixes.

If you hit any other bugs, I'm sure other Issues or PRs will be welcome. Also, you can tag me in Discord or GitHub, and I'll try and make a timely reply.

Thanks for your support on the project, I look forward to any further contributions :) And, have a great weekend!

PR #275, opened 2023-12-06
feat: add Puck class to outer div

monospaced:
Helpful hook for override styling.

### DIFF
--- a/packages/core/components/Puck/index.tsx
+++ b/packages/core/components/Puck/index.tsx
@@ -328,7 +328,7 @@ export function Puck({
   }, []);
 
   return (
-    <div>
+    <div className="Puck">
       <AppProvider
         value={{ state: appState, dispatch, config, componentState }}
       >


packages/core/components/Puck/index.tsx
```diff
@@ -328,7 +328,7 @@ export function Puck({
   }, []);
 
   return (
-    <div>
+    <div className="Puck">
```
chrisvxd, line 331: All other classes are using CSS modules, but this will be `.Puck`, exposing it to style pollution in a way that the others aren't.

Did we have this class before? If not, suggest we wait for #139

monospaced: It was `puck` before FWIW.

Without a way to target the rendered site inside the prview frame, it is not possible to use puck on the Measured site, since that uses `100vh` to ensure the footer is at the bottom of the viewport. Currently I am using this, which does work, but you could argue it would be better to have a `Puck` class to attach this CSS class to. This would not be resolved by theming.

```
/* prevent 100vh when rendered in Puck editor */
body > div .msrd {
  min-height: 100%;
}
```

chrisvxd (approved):

Issue #273, opened 2023-12-05
puck.config.tsx not watched by Remix recipe


chrisvxd:
## Description

Currently, if the user changes their `puck.config.tsx` in the remix recipe, they must reboot the server.

This is because remix only watches the contents of `/app` by default.

### Option 1

Add `puck.config.tsx` to [`watchPaths`](https://github.com/remix-run/remix/issues/1601#issuecomment-1785020576)

### Option 2

Move `puck.config.tsx` within the `/app` directory.

chrisvxd:
FYI @EarthlingDavey, if you ever do another pass on the remix recipe :)

chrisvxd:
I think I'd go with option 1!

EarthlingDavey:
Like last time, should I make a PRs into the main branch?

chrisvxd:
Closed by #279. Thanks @EarthlingDavey!

PR #279, opened 2023-12-07
Fix: puck.config.tsx not watched by Remix recipe

EarthlingDavey:
Fixes the issue https://github.com/measuredco/puck/issues/273

For the remix recipe, `puck.config.tsx` was moved to the `app` directory.

Two options were considered as discussed in the issue: https://github.com/measuredco/puck/issues/273

With Option 1, having the config at root and adding it to watchPaths.

I noticed an intermittent bug where changes in render function would be reflected in the browser, but if a change was undone and saved the browser would not update. To reproduce: 

- Start the dev server with `cd recipe/remix` `yarn dev`.
- Have a `h1` tag in a render function.
- Change to a `h2` and hit save.
- See the changes in the browser.
- Change back to a `h1` tag.
- Intermitently, the browser will not reflect the change.

I suspect the bug is related to some kind of caching of the config file.

With Option 2, moving the config file to the app directory. There were no issues with hmr and all changes were reflected in the browser on saves. So this is the option that I went with.

## Also

Update root eslint because eslint-config-next was not needed in remix recipe - specifically it caused a warning on remix.config.js


chrisvxd (changes requested):
Awesome, thanks again @EarthlingDavey. Happy to go with moving the config.

* Can you update the README for the recipe to mention `/app/puck.config.tsx` instead of `puck.config.tsx`?
* The linting is failing. Would be great if you could run `yarn lint` and fix any issues that arise.

chrisvxd (approved):

Issue #270, opened 2023-12-04
Add min/max to number fields


chrisvxd:
The `number` field type should support `min` and `max` props, passed directly to the input.

## Example

```tsx
const example = {
  "type": "number",
  "min": 0,
  "max": 10
}
```

chrisvxd:
Available in `0.14.0-canary.4932a6e`. Thanks @jperasmus!

jperasmus:
Pleasure

PR #306, opened 2023-12-20
Add min and max properties to number field

jperasmus:
# tl;dr;
Tiny PR to address #270

## details
I wasn't sure if it started to make sense to pull the `NumberField` into its own component or add the 2 new props to the existing `DefaultField`. Went for the pragmatic approach. Happy to split it out as well if needed.

### DIFF
--- a/apps/docs/pages/docs/integrating-puck/categories.mdx
+++ b/apps/docs/pages/docs/integrating-puck/categories.mdx
@@ -33,14 +33,14 @@ const config = {
 };
 ```
 
-You can also change the label, collapse and hide categories:
+You can also change the title, collapse and hide categories:
 
 ```tsx {5,6,10} copy showLineNumbers
 const config = {
   categories: {
     typography: {
       components: ["HeadingBlock", "ParagraphBlock"],
-      label: "Text",
+      title: "Text",
       defaultExpanded: false, // Collapse this category by default
     },
     foundational: {
@@ -63,7 +63,7 @@ const config = {
       components: ["HeadingBlock", "ParagraphBlock"],
     },
     other: {
-      label: "Other components",
+      title: "Other components",
     },
   },
   // ...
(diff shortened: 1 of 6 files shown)


chrisvxd (commented):
This is great, thanks @jperasmus!

If you have time, would you mind adding it to the demo app for the Columns and Flex components?

* [Here](https://github.com/measuredco/puck/blob/main/apps/demo/config/blocks/Columns/index.tsx#L41) for Columns, min 0, max 12
* [Here](https://github.com/measuredco/puck/blob/main/apps/demo/config/blocks/Flex/index.tsx#L22) and [here](https://github.com/measuredco/puck/blob/main/apps/demo/config/blocks/Flex/index.tsx#L29C20-L29C20) for Flex, min 0, no max

If not, don't worry. Once this is merged, I'll add it to the docs.

apps/docs/pages/docs/integrating-puck/categories.mdx
```diff
@@ -33,14 +33,14 @@ const config = {
...
-You can also change the label, collapse and hide categories:
+You can also change the title, collapse and hide categories:
 
 ```tsx {5,6,10} copy showLineNumbers
 const config = {
   categories: {
     typography: {
       components: ["HeadingBlock", "ParagraphBlock"],
-      label: "Text",
+      title: "Text",
```
jperasmus, line 43: Just added this extra small fix to the docs that I noticed was wrong.

jperasmus:
Made the changes as requested and also updated the docs with a basic example showing the 2 new optional fields.

apps/docs/pages/docs/integrating-puck/categories.mdx
chrisvxd: Ah nice one ty

chrisvxd (approved):
Works beautifully! Thanks for updating the docs and fixing my typos, too!

PR #310, opened 2023-12-22
Now Publish will appears if render action button is not provided.

rajatdhoot123:
```
      renderHeaderActions
      ```

Publish button can be handled with renderHeaderAction.

My use case want to rename publish button to export currently there is no option to do that. We can have on prop to change the text of publish button but feel as we currently have a prop to add more button we can handle publish button also by this section.

### DIFF
--- a/packages/core/components/MenuBar/index.tsx
+++ b/packages/core/components/MenuBar/index.tsx
@@ -29,6 +29,7 @@ export const MenuBar = ({
   renderHeaderActions?: (props: {
     state: AppState;
     dispatch: (action: PuckAction) => void;
+    onPublish?: (data: Data) => void;
   }) => ReactElement;
   setMenuOpen: Dispatch<SetStateAction<boolean>>;
 }) => {
@@ -77,22 +78,25 @@ export const MenuBar = ({
           </IconButton>
         </div>
         <>
-          {renderHeaderActions &&
+          {renderHeaderActions ? (
             renderHeaderActions({
               state: appState,
               dispatch,
-            })}
+              onPublish: () => onPublish && onPublish(data),
+            })
+          ) : (
+            <div>
+              <Button
+                onClick={() => {
+                  onPublish && onPublish(data);
+                }}
+                icon={<Globe size="14px" />}
+              >
+                Publish
+              </Button>
+            </div>
+          )}
         </>
-        <div>
-          <Button
-            onClick={() => {
-              onPublish && onPublish(data);
-            }}
-            icon={<Globe size="14px" />}
-          >
-            Publish
-          </Button>
-        </div>
       </div>
     </div>
   );

### CODE REVIEW

**Comment by rajatdhoot123** [comment]
I am not able to test because there is some issue with `npm run dev`

```
 ERROR  run failed: error preparing engine: Invalid persistent task configuration:
You have 10 persistent tasks but `turbo` is configured for concurrency of 10. Set --concurrency to at least 11
```

File: packages/core/components/MenuBar/index.tsx [hunk 1]
```diff
@@ -77,22 +78,25 @@ export const MenuBar = ({
...
+            <div>
+              <Button
+                onClick={() => {
+                  onPublish && onPublish(data);
+                }}
+                icon={<Globe size="14px" />}
+              >
+                Publish
+              </Button>
+            </div>
```
chrisvxd, line 97: This should be passed as `children` to `renderHeaderActions` and `overrides.headerActions`, so the user can still choose to render the default button

chrisvxd (commented):
Thanks @rajatdhoot123. I understand the use-case here, but it's a breaking change on a commonly used feature, so need to be careful. I'm going to hold off merging this just yet.

If you need this urgently, suggest that you can override the full header to achieve a similar result.

FWIW, to test the demo app you can `cd apps/demo && npm run dev`, instead of doing it at the project root.

packages/core/components/MenuBar/index.tsx
rajatdhoot123: From this I understand you want me to rewrite the component in this way.

```
{renderHeaderActions && (
            renderHeaderActions({
              state: appState,
              dispatch,
              onPublish: () => onPublish && onPublish(data),
            })
          ) 
```

And Completely remove this button code as we can directly pass publish button as children to `renderHeaderActions`

rajatdhoot123:
And also, I want you to thanks for this library as I am tried using craftjs which is too complicated. Puck is really too easy to use.

packages/core/components/MenuBar/index.tsx
chrisvxd: Sorry for the delay! Xmas etc :)

Not quite, I want something like this 

```tsx
const PublishButton = ( { onPublish } ) => (
  <div>
    <Button
      onClick={() => {
        onPublish && onPublish(data);
      }}
      icon={<Globe size="14px" />}
    >
      Publish
    </Button>
  </div>
);
```

```tsx
          {renderHeaderActions ? (
             renderHeaderActions({
               state: appState,
               dispatch,
               children: <PublishButton onPublish={onPublish} />
             </div>
             })
           ) : (
             <PublishButton onPublish={onPublish} />
           )
```

and updating the [`<CustomHeaderActions>` call](https://github.com/measuredco/puck/blob/f2f0c9d00f08b28da765abef8887407d47deb96b/packages/core/components/Puck/index.tsx#L405) to take children

```tsx
                        <CustomHeaderActions>
                          <PublishButton onPublish={onPublish} />
                        </CustomHeaderActions>
```

I don't think there's any point in passing in the `onPublish` function either.

chrisvxd:
Closing as no activity.

PR #315, opened 2023-12-24
Fix file options types to allow readonly arrays

jperasmus:
# tl;dr;
Allow both inline and read-only arrays for fields that take an `options` array (`select` and `radio` fields)

## details
Currently, if you provide a readonly array (by specifying `as const` on the array), Puck will throw a TypeScript error.

```tsx
const selectOptions = [
  {
    label: "A",
    value: "a",
  },
  {
    label: "B",
    value: "b",
  },
] as const

export const SomeComponent: ComponentConfig<SomeComponentProps> = {
  fields: {
    selectField: {
      type: "select",
      options: selectOptions, // <-- This has a TS error
    },
  },
}
```

### DIFF
--- a/packages/core/types/Config.tsx
+++ b/packages/core/types/Config.tsx
@@ -7,6 +7,13 @@ type WithPuckProps<Props> = Props & {
   id: string;
 };
 
+type FieldOption = {
+  label: string;
+  value: string | number | boolean;
+};
+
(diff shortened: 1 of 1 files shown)


chrisvxd (approved):
Nice easy win, thanks @jperasmus!

PR #324, opened 2024-01-12
feat: move CSS import in Next.js recipe

### LINKED ISSUE #319: Remove CSS import from core root export?
Having `import "./styles/global.css"` in [/packages/core/index.ts](https://github.com/measuredco/puck/blob/main/packages/core/index.ts#L1) causes a side effect of importing all the puck CSS — including `https://rsms.me/inter/inter.css` and the Inter font(s) — if _any_ of the exports in the file are used.

So if a user imports anything at all from `@/core` (including types, hooks and helpers) they will get _all_ of the puck CSS (and the Inter files) bundled in their app. And unless they are conditionally importing in `/edit` routes onlly, these styles and fonts will also end up in the user's published pages.

I guess the CSS import is there because the index file also exports some UI components (`Button`, `Drawer`, etc.), which need the styles to work. But in the demo app and the next recipe, we explicitly import the puck styles with `import "@/core/styles.css";`. So i think this is a better pattern – you must explicilty import the styles.

Some users may be replying on this import, so removing it would be a breaking change.

It would also be good to update the demo and next reciepe so that Inter and the puck CSS aren't bundled into the published pages by default. Currently we're crossing the streams, because the demo actually _uses_ Inter, and even some of the puck components (e.g. `Button`) for convenience. But in real-world use cases, you’d mostly likely _not_ want the puck CSS and Inter bundled into production.

monospaced:
Closes #319 

I guess this issue could also exist in the Remix recipe, but don't know enough about that to change anything there.

### DIFF
--- a/recipes/next/app/layout.tsx
+++ b/recipes/next/app/layout.tsx
@@ -1,4 +1,3 @@
-import "@measured/puck/puck.css";
 import "./styles.css";
 
 export default function RootLayout({
--- a/recipes/next/app/puck/[...puckPath]/page.tsx
+++ b/recipes/next/app/puck/[...puckPath]/page.tsx
@@ -11,6 +11,7 @@
  * NB this route is public, and you will need to add authentication
  */
 
+import "@measured/puck/puck.css";
 import { Client } from "./client";
 import { Metadata } from "next";
 import { getPage } from "../../../lib/get-page";


chrisvxd (approved):
Good fix!

It's not an issue in the remix recipe: https://github.com/measuredco/puck/blob/main/recipes/remix/app/routes/edit.tsx#L2

Should be a `fix`, but I'll correct that on merge

PR #325, opened 2024-01-12
Add Contentful field

chrisvxd:
WIP


chrisvxd:
@monospaced would appreciate your draft review and thoughts of this approach vs a docs-only approach. I will not mark as ready until we have discussed further.

monospaced (commented):
Looks good to me! API is clear and seems a good balance between sane defaults and flexibility to customise. 

I think this package approach is a lot more useful and substantial than documentation-only. It also _serves as_ a form of documentation, with examples, to use as a starting point for a bespoke contentful (or other) integration.

Issue #327, opened 2024-01-15
Improve a11y of clickable table rows in `ExternalInput`


monospaced:
Text content in the `ExternalInput` modal is currently selected by an `onClick` handler on a `tr`. There is no keyboard support, and the pattern is a [particular annoyance for screen reader users](https://adrianroselli.com/2020/02/block-links-cards-clickable-regions-etc.html#Update02).

PR #331, opened 2024-01-15
External field improvements

### LINKED ISSUE #261: Add filters to external fields
Provide an API to filter the user's external data, allowing the user to customise the filters available and passed to the `fetchList` function under `filters` (or similar).

(Spun out from #203)

## Considerations

* What is the appropriate UI here?
* Is this better left to custom fields?
* Can we reuse the `fields` API?

chrisvxd:
This PR introduces new features to the External field. These features were spun out of work on the upcoming Contentful package (#325).

## Features

* A new `mapRow` API for External fields, to select which fields appear in the table
* New `filterFields` (and `initialFilters`) APIs to enable the user to create filter forms using the Fields API
* General UI and docs refinements

## Demo

Please read the docs for examples

---

Closes #261

### DIFF
--- a/packages/core/components/ExternalInput/styles.module.css
+++ b/packages/core/components/ExternalInput/styles.module.css
@@ -76,15 +76,17 @@
   display: flex;
   flex-direction: column;
   position: relative;
+  min-height: 50vh;
   max-height: 90vh;
 }
 
 .ExternalInputModal-masthead {
-  background-color: var(--puck-color-white);
+  background-color: var(--puck-color-grey-12);
+  border-bottom: 1px solid var(--puck-color-grey-09);
   display: flex;
   flex-wrap: wrap;
   gap: 24px;
-  padding: 32px 24px;
+  padding: 24px;
 }
 
 .ExternalInputModal-tableWrapper {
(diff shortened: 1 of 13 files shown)


packages/core/components/ExternalInput/styles.module.css
```diff
@@ -104,39 +142,35 @@
...
+  width: 1%; /* Prevent growing */
+  white-space: nowrap; /* Prevent growing */
 }
 
 @media (hover: hover) and (pointer: fine) {
   .ExternalInputModal-tbody .ExternalInputModal-tr:hover {
-    background: var(--puck-color-grey-11);
-    color: var(--puck-color-azure-04);
+    background: var(--puck-color-azure-11);
+    color: var(--puck-color-azure-05);
```
monospaced, line 173: Foreground color needs to be `azure-04` for sufficent contrast. The `04` level is always the foreground text color in light more. Also `azure-12` is the default hover background color in light mode.

packages/core/components/ExternalInput/styles.module.css
```diff
@@ -145,11 +179,17 @@
   .ExternalInputModal-tbody
     .ExternalInputModal-tr:hover
     .ExternalInputModal-td:first-of-type {
-    border-left: 4px solid var(--puck-color-azure-04);
+    border-left: 4px solid var(--puck-color-azure-05);
```
monospaced, line 182: See above

monospaced (changes requested):
Looking good! Couple of minor things. Also noticed that the whole table is now able to recieve keyboard focus, but not sure how or if intentional?

Screenshot 2024-01-15 at 13 28 49

chrisvxd:
@monospaced backatchya. And no, the focus change was not intentional but suggest we pick it up in #327

monospaced (approved):

PR #340, opened 2024-01-24
fix: component resolver issue if component not present in config

### LINKED ISSUE #339: Error while loading puck editor
Screenshot 2024-01-24 at 12 04 04 AM
Screenshot 2024-01-24 at 12 04 14 AM

i'm seeing this error while loading puck editor at `const flatcontent = ` line. it's happening when the data has any component which is not present in the component config passed to puck editor  @chrisvxd. in previous versions this error was not there.

sagarchoudhary96:
Fixes #339

### DIFF
--- a/packages/core/lib/use-resolved-data.ts
+++ b/packages/core/lib/use-resolved-data.ts
@@ -47,7 +47,7 @@ export const useResolvedData = (
 
     const flatContent = Object.keys(newData.zones || {})
       .reduce((acc, zone) => [...acc, ...newData.zones![zone]], newData.content)
-      .filter((item) => !!config.components[item.type].resolveData);
+      .filter((item) => !!config.components[item.type]?.resolveData);
 
     const applyIfChange = (
       dynamicDataMap: Record<string, ComponentData>,


sagarchoudhary96:
@chrisvxd  this fixes it.

chrisvxd (approved):
Makes sense! Sounds like you have an out-of-date data payload, but there's no harm in us checking within Puck, too.

Thanks for the fix.

sagarchoudhary96:
so actually, the use case we use puck editor is to build pages and those can be built through automations etc too, and what might happen is we might have added some new components to those pages  through one flow but didn't have added support yet to the puck editor and hence this could happen when opening those pages in puck editor.

chrisvxd:
Interesting use case @sagarchoudhary96! Can you share more about your project?

sagarchoudhary96:
we basically use some automations in place to generate these custom pages for each of our user based on their answers , and then these pages are available to  other team members to be able to easily edit/update accordingly as per any additional changes needed using the editor interface made using puck editor.

Issue #326, opened 2024-01-15
Add `max` to array field API


chrisvxd:
## Description

Users may wish to limit the number of items added to an array field

## Proposal

A new `max` API for the array field type

```js
const config = {
  components: {
    Example: {
      fields: {
        items: {
          type: "array",
          arrayFields: {
            title: { type: "text" },
          },
          max: 5 // new
        },
      },
  },
};
```

xaviemirmon:
@chrisvxd love this! +1 from me.  Could we also add a min to this issue?

```
const config = {
  components: {
    Example: {
      fields: {
        items: {
          type: "array",
          arrayFields: {
            title: { type: "text" },
          },
          max: 5 // new
          min: 2 // new
        },
      },
  },
};
```

chrisvxd:
Implemented by #341

xaviemirmon:
I am just playing around with the demo on this @chrisvxd.  I feel like the plus button should be hidden when you reach max and the delete button when you reach min.  Should this be re-opened to include this or a new issue raised?

PR #341, opened 2024-02-02
added possibility to set a max and min amount of items in the array f…

jabba-the-bug:
Implemented what was described in this issue
https://github.com/measuredco/puck/issues/326

the code will disable delete/add buttons once the specified limit is reached
it will also swap the default icon for the alert-circle icon

### DIFF
--- a/packages/core/components/InputOrGroup/fields/ArrayField/index.tsx
+++ b/packages/core/components/InputOrGroup/fields/ArrayField/index.tsx
@@ -1,6 +1,6 @@
 import getClassNameFactory from "../../../../lib/get-class-name-factory";
 import styles from "./styles.module.css";
-import { List, Plus, Trash } from "lucide-react";
+import { List, Plus, Trash, AlertCircle } from "lucide-react";
 import { FieldLabelInternal, InputOrGroup, type InputProps } from "../..";
 import { IconButton } from "../../../IconButton";
 import { reorder, replace } from "../../../../lib";
@@ -205,6 +205,7 @@ export const ArrayField = ({
                                 <div className={getClassNameItem("actions")}>
                                   <div className={getClassNameItem("action")}>
                                     <IconButton
+                                      disabled={field.min !== undefined && field.min >= localState.arrayState.items.length}
                                       onClick={(e) => {
                                         e.stopPropagation();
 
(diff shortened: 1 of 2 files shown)


packages/core/components/InputOrGroup/fields/ArrayField/index.tsx
```diff
@@ -1,6 +1,6 @@
 import getClassNameFactory from "../../../../lib/get-class-name-factory";
 import styles from "./styles.module.css";
-import { List, Plus, Trash } from "lucide-react";
+import { List, Plus, Trash, AlertCircle } from "lucide-react";
```
chrisvxd, line 3: I don't think we need the AlertCircle icons here. I think we should just rely on the `disabled` states.

chrisvxd (changes requested):
Thanks for the contribution @jabba-the-bug, great stuff! Left one comment.

Can you also fix the formatting issues from the [build](https://github.com/measuredco/puck/actions/runs/7759899158/job/21487787155?pr=341)? You can do that with `yarn format`.

I'd also appreciate it if you added an example to the demo app - perhaps limit the buttons on the hero to `min: 1` and `max: 4`. 

Finally, I'll need to add documentation to this. I'm happy to do it in a follow up PR.

jabba-the-bug:
- Fixed the min, max example for the hero buttons
- removed use of AlertCircle
- ran yarn format
- added some documentation for use of min, max

Let me know if there is anything else needed.
I've never used prettier before got a bunch of changed files only commited the files I've touched, hope that was correct

packages/core/components/InputOrGroup/fields/ArrayField/index.tsx
```diff
@@ -288,6 +293,10 @@ export const ArrayField = ({
 
                 <button
                   className={getClassName("addButton")}
+                  disabled={
+                    field.max !== undefined &&
+                    localState.arrayState.items.length >= field.max
+                  }
```
chrisvxd, line 299: Looks like we're missing disabled styles for this and the IconButton. I'll open a separate ticket for that.

chrisvxd (changes requested):
Looks good! I'll make a few tweaks to docs after merge, but can you run `yarn format` again? Your build is still failing.

jabba-the-bug:
Ran yarn format :)

chrisvxd (approved):
🎉

Issue #317, opened 2023-12-29
Expose history to `usePuck` hook


chrisvxd:
## Description

The `history` object should be exposed to the `usePuck` hook, along with `forward()` `back()`, `hasFuture` and `hasPast`.

prashanthvdckap:
Hi @chrisvxd Thanks for keeping track. When will this feature go live?

chrisvxd:
Hi @prashanthvdckap - suspect not until next month at the earliest as we have other items in our queue. Feel free to contribute the feature if you have the time.

prashanthvdckap:
Hi @chrisvxd  i have exposed history in usePuck hook. please review.
Thanks

chrisvxd:
Closed in #342

PR #342, opened 2024-02-06
exposed history state in usePuck hook

prashanthvdckap:
exposed history state in usePuck hook 

#317 is fixed


chrisvxd (approved):

chrisvxd:
Awesome stuff, thanks @prashanthvdckap!

Issue #269, opened 2023-12-04
Accessibility, markup & CSS improvements


chrisvxd:
Following #247, we should do a pass to address accessibility, markup and CSS issues, and related problems.

This is blocked by #247 and #265. It blocks #139, which should be done with CSS properties.

## Tasks

* [ ] Sort CSS properties
* [ ] Systematise hex colors with alpha channel, or comment on them
* [ ] Consistent sorting of imports

chrisvxd:
@monospaced let's use this to track and implement broad improvements. Please drop a comment with anything you've already identified for improvement, or anything that comes up during #247.

adampage:
For the [accessibility audit](https://github.com/measuredco/puck/issues/269#:~:text=breakpoints%C2%A0%23372-,Accessibility%20audit,-Contrast), I‘d recommend having a look at the W3C’s [Authoring Tool Accessibility Guidelines (ATAG)](https://www.w3.org/WAI/standards-guidelines/atag/) standard. ATAG is a nice companion to [WCAG](https://www.w3.org/WAI/WCAG22/quickref/) since it identifies some distinct success criteria applicable to CMS UIs.

Beyond its coverage for the accessibility of the interface itself, I also appreciate that ATAG has a whole section on **supporting authors** (encouraging them, even) to [**produce accessible content**](https://www.w3.org/TR/ATAG20/#principle_b2). 😅 Puck’s fantastic [heading analyzer plugin](https://github.com/measuredco/puck/tree/main/packages/plugin-heading-analyzer) is a heartening example of where you’ve already done work in this space.

Warning: ATAG was [last published in 2015](https://www.w3.org/TR/ATAG20/#principle_b2:~:text=W3C%20Recommendation-,24%20September%202015,-This%20version%3A), so please forgive it for potentially being out-of-touch with modern content management concepts and techniques.

PR #353, opened 2024-02-26
feat: add iframes and viewport previewing

chrisvxd:
👉 **[Demo](http://puck-demo-git-iframe-measured.vercel.app/edit)** 👈

## Description

This PR:

* adds a fully drag-and-drop compatible iframe to Puck.Preview, enabling style isolation and media queries
* adds viewport switching for default small, medium and large
* adds zooming to the viewport switching
* introduces a new API, `viewports`, for customising the default viewports
* exposes viewport control via the UI state and dispatcher (with exception of zoom)

To support this, we needed to implement two additional libraries:

1. [measured/dnd](https://github.com/measuredco/dnd/tree/iframes): A fork of [hello-pangea/dnd](https://github.com/hello-pangea/dnd) (an actively maintained fork of react-beautiful-dnd) that supports both iframes and CSS transforms.
2. [measured/auto-frame-component](https://github.com/measuredco/auto-frame-component): A new library to sync styles from the host to the iframe

https://github.com/measuredco/puck/assets/985961/93a76dd5-2a48-4200-872d-e4eb57ad269a

## Testing

Play around with the viewports, ensuring drag-and-drop still works as expected.

* [Demo](http://puck-demo-git-iframe-measured.vercel.app/edit)
* [Docs](https://puck-docs-git-iframe-measured.vercel.app/docs/api-reference/components/puck#viewports)

## Notes

* Whilst viewports and drag-and-drop work on mobile, managing a large viewport (such as desktop) can cause iOS Safari to crash due to high GPU usage.
* We could consider removing animations, or introducing an API to disable animations.
* We _should_ disable transition when prefers-reduced-motion is set.

## Future tasks

* Add API for iframe style isolation
* Add API and UI for adding a select menu of viewports (rather than just icon buttons)


packages/core/components/Puck/index.tsx
```diff
@@ -80,38 +84,82 @@ export function Puck<
...
+          key,
+          diff: Math.abs(viewportWidth - value.width),
+        }))
+        .sort((a, b) => (a.diff > b.diff ? 1 : -1));
+
+      const closestViewport = viewportDifferences[0].key;
+
+      clientUiState = {
+        // Hide side bars on mobile
+        ...(window.matchMedia("(min-width: 638px)").matches
```
monospaced, line 114: Might be worth annotating this value. Assume it's dervied from some calculation of widths of various bit of Puck UI chrome? Possibility of modifying CSS further down the line and not realising this value also needs to change perhaps.

packages/core/components/Puck/styles.module.css
```diff
@@ -194,3 +207,23 @@
...
+.Puck-canvas {
+  background: var(--puck-color-grey-11);
+  display: flex;
+  grid-area: editor;
+  flex-direction: column;
+  padding: var(--puck-space-px);
+  overflow: auto;
+}
+
+@media (min-width: 1198px) {
```
monospaced, line 220: These `min-width` values can probably change now, as they were calculated as best compromise for the the preview width, given it didn't have it's own viewport context. But not blocking, potential optimisation for later.

monospaced (changes requested):
Some minor bits, not all need action. Code review only, still need to test.

monospaced:
## Testing

* Unlike the current implementation, the preview does not scroll when dragging a component up or down from the current preview scroll position. So can only drop on an area of the preview that is already scrolled into view. Tested in FF & Chrome. If the place you want to drop is not visible, you need to put the dragged component down, then the scroll the window, then drag the component again. Assume this is a limitation of iFrame approach? Noting as does impact usability. Potentially I just need to get used to it - you can get around it with a repeated drag>drop>scroll>drag>drop>scroll manoeuvre. Could also zoom in to show more of the preview I guess. But other than that, I can't discern a difference in dnd bevaiour from the current version 🎉
* Stats component default state breaks layout in narrow viewport. Could potentially add some media query CSS to fix this now.
Screenshot 2024-03-21 at 12 17 06

packages/core/components/Puck/styles.module.css
chrisvxd: I think it might be best to do this as a follow on (possibly as part of #269 to avoid bloating this PR)

packages/core/components/ViewportControls/default-viewports.ts
```diff
@@ -0,0 +1,7 @@
+import { Viewports } from "../../types/Viewports";
+
+export const defaultViewports: Required<Viewports> = [
+  { width: 360, height: "auto", icon: "Smartphone", label: "Small" },
+  { width: 768, height: "auto", icon: "Tablet", label: "Medium" },
+  { width: 1280, height: "auto", icon: "Monitor", label: "Large" },
+];
```
chrisvxd, line 7: @monospaced do these seem like sensible defaults to you?

packages/core/components/Puck/index.tsx
monospaced: Right yes, all of the media query stuff should be updated now that the preview has it's own viewport. One for later.

packages/core/components/Puck/styles.module.css
monospaced: Agree

packages/core/components/ViewportControls/default-viewports.ts
monospaced: Yes, I did consider that in review, and I think they do ✅

packages/core/components/Puck/components/Canvas/index.tsx
```diff
@@ -0,0 +1,165 @@
...
+
+    return () => {
+      observer.disconnect();
+    };
+  }, []);
+
+  return (
+    <div
+      className={getClassName("canvas")}
+      onClick={() =>
```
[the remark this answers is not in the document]
chrisvxd: Let's pick up in #269?

packages/core/components/DraggableComponent/index.tsx
```diff
@@ -92,8 +94,21 @@ export const DraggableComponent = ({
...
+            style={{
+              top: 56 / zoomConfig.zoom,
+            }}
+          >
+            <div
+              className={getClassName("actions")}
+              style={{
+                transform: `scale(${1 / zoomConfig.zoom}`,
+                top: -48 / zoomConfig.zoom,
+                right: 6.5 / zoomConfig.zoom,
```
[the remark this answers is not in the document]
chrisvxd: These were already hardcoded in CSS, but I've extracted them as `const` values, and tweaked them a bit since things have moved.

monospaced (approved):
Let's goooooo

PR #355, opened 2024-02-27
Refactor: dropzone context

4leite:
Use discriminated union type for DropZoneContext to provide a better typed context.

Facilitates the removal of a bunch of guard clauses and better documents the expected use.

Export DropZoneContext inside useDropZoneContext, useDropZoneEditContext and useDropZoneRenderContext.
Throw error on incorrect mode - we could add an ErrorBoundary to catch this, shouldn't be necessary due to:.
The current behaviour of gracefully showing a message on missing context is retained by moving it to the parent <DropZone> component.

### DIFF
--- a/packages/core/components/DropZone/context.tsx
+++ b/packages/core/components/DropZone/context.tsx
@@ -44,10 +44,10 @@ export type DropZoneEditContext<
   setHoveringZone: (zone: string | null) => void;
   hoveringComponent?: string | null;
   setHoveringComponent: (id: string | null) => void;
-  registerZoneArea?: (areaId: string) => void;
+  registerZoneArea: (areaId: string) => void;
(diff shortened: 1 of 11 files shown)


chrisvxd (changes requested):
Thanks for putting this together! All seem like sensible refactors so I'm pretty happy for this to go in once the latest conflicts are resolved.

4leite:
> Thanks for putting this together! All seem like sensible refactors so I'm pretty happy for this to go in once the latest conflicts are resolved.

Yeah you had to go and do all that work on the dnd iframe compat!

Got a bunch on and then it's Easter, so likely won't get back to this for a couple of weeks.

chrisvxd:
Closing this until you're ready @4leite

PR #357, opened 2024-02-28
fix: puck context missing from preview render

4leite:


packages/core/components/Render/index.tsx
```diff
@@ -4,9 +4,13 @@ import { rootDroppableId } from "../../lib/root-droppable-id";
 import { Config, Data } from "../../types/Config";
 import { DropZone, DropZoneProvider } from "../DropZone";
 
-export function Render<
-  UserConfig extends Config<any, any, any> = Config<any, any, any>
->({ config, data }: { config: UserConfig; data: Data }) {
+export function Render<UserConfig extends Config = Config>({
```
chrisvxd, line 7: These types are very deliberate, and I have found that removing the generics causes unexpected issues.

Fixed [here](https://github.com/measuredco/puck/pull/257/commits/82ed63a737e3f9ace114b1a287144ce639776dc7) as part of #257.

packages/core/components/Puck/components/Preview/index.tsx
```diff
@@ -27,7 +37,7 @@ export const Preview = ({ id = "puck-preview" }: { id?: string }) => {
       }}
       style={{ zoom: disableZoom ? 1 : 0.75 }}
     >
-      <Page dispatch={dispatch} state={state} {...rootProps}>
+      <Page {...rootProps}>
```
chrisvxd, line 40: `dispatch` and `state` were kept here to retain backwards compatibility with [plugins from v0.12.0](https://puckeditor.com/v/0.12.0/docs/api-reference/plugins/functions#render-args), before it was changed in v0.13.0.

I think I'd rather retain these for now, but we should probably mark them as `// DEPRECATED` so we remember to pull them out during a breaking change release.

packages/core/components/Puck/components/Preview/index.tsx
```diff
@@ -2,15 +2,25 @@ import { DropZone, dropZoneContext } from "../../../DropZone";
...
+    (pageProps: DefaultRootProps) => {
+      return config.root?.render
+        ? config.root?.render({
+            ...pageProps,
+            id: "puck-root",
+            editMode: true,
+            puck: { renderDropZone: DropZone },
+          })
+        : pageProps.children;
+    },
```
chrisvxd, line 23: So to check, this ensures that components built using the renderDropZone functionality can still render in Preview mode following `v0.13.0`?

chrisvxd (changes requested):

packages/core/components/Puck/components/Preview/index.tsx
4leite: Correct

packages/core/components/Render/index.tsx
4leite: The problem is it widens the type (effectively returning any) which is probably what fixed your issue, but it means that issue below snuck in because it wasn't strongly typed any more.

chrisvxd:
@4leite I've reworked the types as part of #385, and also fixed the `renderDropZone` issue you've identified here. Closing this PR in favour of that.

PR #359, opened 2024-03-04
fix: tsup watch mode not refreshing on css change

4leite:
tsup now processes css pre-processes natively
this is also resolves core yarn build --watch not refreshing css changes
see https://github.com/egoist/tsup/issues/536#issuecomment-1752121594


chrisvxd (approved):
Oh that's much cleaner, thank you!

Issue #150, opened 2023-10-11
Support for inline editing text and textareas


xaviemirmon:
Another feature related to #149 a bit is supporting inline editing of text on components.  

@chrisvxd I'm sure there are many considerations when working with beautiful-react-dnd, but I'm adding an issue so it's on the radar.

I'd love to have the ability to edit the text on my component inline from the main canvas, and if it was rich text have the ability to change the text style inline.

chrisvxd:
I don't think this would be _that_ difficult at this point, and the API would probably be similar to DropZones. We'd also probably use context

```tsx
render: () => <p><InlineField name="title" /></p>
```

One question I would have is whether we retain the prop for the field. For example

```tsx
fields: { title: { type: "text"} },
render: () => <p><InlineField name="title" /></p>
```

I guess it should probably be optional, but the inline field data should still be stored as a prop regardless.

BleedingDev:
Would this work with other interactive components (for example custom `contentEditable` or some markdown editor)?

4leite:
I might have a go at tackling this next as it's something I'm quite keen to see. I've implemented an auto-resizing inline text editing input before.
Are we interested in supporting markdown? We could look to leverage something like:
https://github.com/facebook/lexical

xaviemirmon:
@4leite I was thinking of a plain text setup when I raised this.  If we expand it to an RTE then we should probably use the same one as agreed upon for #149

chrisvxd:
I'm a little undecided over whether we want this, or if it would just detract from the field API. Playing around with a demo might help, if anyone's happy to put one together.

BleedingDev:
I would start with inline editing without MD or something, but would support just making the component clickable and interactive (so that you can plug in something like Lexical).

Advanced editors can be build on top of Custom Input for now.

4leite:
I was thinking something like this https://craft.js.org/

xaviemirmon:
100% @4leite, that was the example I had in mind too!

4leite:
Took me a while to get back to this.
#265 added the usePuck hook and I've got a surprisingly usable inline editor using lexical and a very brittle custom hook wrapping usePuck:

```
"use client";

import { usePuck } from "@measured/puck";
import { SerializedEditorState } from "lexical";
import { LexicalClient } from "./client";

const useSelected = (componentId: string) => {
  const {
    appState: {
      ui: { itemSelector },
      data,
    },
    dispatch,
  } = usePuck();

[...]

PR #360, opened 2024-03-04
feature: add component toolbar portal target

4leite:
Note: This also adjusts the position of the toolbar, attaching it to the top border of the component to prevent the component above 'flashing' on hover. This becomes quite intrusive when interacting with the toolbar regularly.

### DIFF
--- a/packages/core/components/DraggableComponent/index.tsx
+++ b/packages/core/components/DraggableComponent/index.tsx
@@ -97,6 +97,7 @@ export const DraggableComponent = ({
               {label && (
                 <div className={getClassName("actionsLabel")}>{label}</div>
               )}
+              <div id={`DraggableComponentToolbarPortal-${id}`} />
               <button className={getClassName("action")} onClick={onDuplicate}>
                 <Copy size={16} />
               </button>
--- a/packages/core/components/DraggableComponent/index.tsx
+++ b/packages/core/components/DraggableComponent/index.tsx
@@ -104,6 +104,7 @@ export const DraggableComponent = ({
               <button className={getClassName("action")} onClick={onDelete}>
                 <Trash size={16} />
               </button>
+              <div className={getClassName("actionsMask")} />
             </div>
           </div>
           <div className={getClassName("contents")}>{children}</div>
--- a/packages/core/components/DraggableComponent/styles.module.css
+++ b/packages/core/components/DraggableComponent/styles.module.css
@@ -28,7 +28,7 @@
 
 .DraggableComponent-overlay {
   display: none;
-  background: color-mix(in srgb, var(--puck-color-azure-08) 30%, transparent);
+  background: green; /*color-mix(in srgb, var(--puck-color-azure-08) 30%, transparent);*/
   cursor: pointer;
   height: 100%;
   width: 100%;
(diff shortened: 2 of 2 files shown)


4leite:
adds support for #150

packages/core/components/DraggableComponent/styles.module.css
```diff
@@ -99,13 +99,12 @@
 
 .DraggableComponent-actions {
   position: absolute;
-  right: 6.5px;
+  right: -1px;
   width: auto;
-  top: -48px;
+  top: -37px;
```
chrisvxd, line 104: I don't think moving the toolbar here is the right solution, as leaving space between the toolbar and the border is a deliberate design choice that gives. I agree the flashing is annoying, but I think the better solution would be to create an invisible spacer between them that prevents the hover from bubbling up.

packages/core/components/DraggableComponent/index.tsx
```diff
@@ -97,6 +97,7 @@ export const DraggableComponent = ({
               {label && (
                 <div className={getClassName("actionsLabel")}>{label}</div>
               )}
+              <div id={`DraggableComponentToolbarPortal-${id}`} />
```
chrisvxd, line 100: What's the purpose of this portal? To inject new actions into the toolbar?

If so, I think a first party API might be a cleaner way of doing this instead of asking the user to inject via a portal. Something like `<Puck toolbarActions={<div />} />`

chrisvxd (changes requested):
Thanks @4leite! Couple of questions

packages/core/components/DraggableComponent/styles.module.css
4leite: Yup happy to implement that approach instead.

packages/core/components/DraggableComponent/index.tsx
4leite: I want to be able to inject from the inline editor. eg https://inline-rich-text.vercel.app/edit (click on the heading).
In the above example they are at the end of toolbar, but I think the middle is better (that was just the easiest for me to target with getElement / css Query hack).
These are dynamically changing based on lexical plugins.
An alternative would be to set them via setUI, but I do think the portal option is much cleaner for my particular use case.

chrisvxd: I think this API needs a bit more thought before we move forwards:

1. I don't think we should be using portals for some elements of custom UI, and props for others.
2. Users may wish to remove existing overlay controls (possibly a separate, but related API)

I'm not tracking this here: https://github.com/measuredco/puck/issues/377

packages/core/components/DraggableComponent/styles.module.css
4leite: fixed

packages/core/components/DraggableComponent/index.tsx
4leite: I would still like the (amended) UI fix for the dropdown, which I can put as a separate PR.

#377 is good, but... I would still at the least like a stable data-* attribute I can target in the mean time, even if it's an experimental, undocumented solution.

chrisvxd:
Closing as out of date. Conversation to continue in #377.

PR #366, opened 2024-03-14
Docs site SEO

monospaced:
Hygiene and best practise pass on meta data, favicons, etc. to bring in line with [Measured](https://measured.co).

* Favicons [best practise](https://dev.to/masakudamatsu/favicon-nightmare-how-to-maintain-sanity-3al7)
* Add canonical links
* Additonal `og:` and `twitter:` tags
* Add JSON LD `WebSite` entry to support [Google Search Results site name](https://developers.google.com/search/docs/appearance/site-names#website)
* Redirect [/docs/api-reference/plugins](https://puck-docs-git-seo-bits-measured.vercel.app//docs/api-reference/plugins) (fix 404 in Google Search Console)
* Add [sitemap](https://puck-docs-git-seo-bits-measured.vercel.app/sitemap.xml)


apps/docs/next-sitemap.config.js
```diff
@@ -0,0 +1,5 @@
+/** @type {import('next-sitemap').IConfig} */
+module.exports = {
+  siteUrl: "https://puckeditor.com",
+  generateIndexSitemap: false,
+};
```
chrisvxd, line 5: How can I test this? When I visit `/sitemap.xml` I get a 404.

apps/docs/public/manifest.webmanifest
```diff
@@ -0,0 +1,21 @@
+{
+  "name": "Puck",
+  "short_name": "Pck",
```
chrisvxd, line 3: Interesting. Does this have to be 3 characters?

apps/docs/theme.config.tsx
```diff
@@ -8,36 +8,63 @@ import { Viewport } from "./components/Viewport";
...
+      <meta name="twitter:description" content={description} />
+      <meta name="twitter:image:alt" content="Puck" />
+      <meta name="twitter:image:height" content="675" />
+      <meta name="twitter:image:type" content="image/png" />
+      <meta name="twitter:image:width" content="1200" />
+      <meta name="twitter:site" content="@puckeditor" />
+      <meta
+        name="twitter:title"
+        content={title !== defaultTitle ? `${title} - Puck` : defaultTitle}
+      />
```
chrisvxd, line 49: All pages already have the `- Puck` suffix - I suspect this might cause duplicate suffixes? i.e. `Introduction - Puck - Puck`

chrisvxd (changes requested):

apps/docs/public/manifest.webmanifest
monospaced: Nope, it doesn't have to be anything. Could just be Puck again I guess

apps/docs/theme.config.tsx
monospaced: I don't think so, as `title` in this context (coming from `useConfig`) does not have the suffix, hence why I had to add it manually here.

apps/docs/public/manifest.webmanifest
monospaced: It's only used if there is not enough space to display the `name`, so probably moot in the case of "Puck"

https://developer.mozilla.org/en-US/docs/Web/Manifest/short_name

apps/docs/next-sitemap.config.js
monospaced: Reployed and it's back 🤷🏻‍♂️

https://puck-docs-git-seo-bits-measured.vercel.app/sitemap.xml

apps/docs/theme.config.tsx
monospaced: Screenshot 2024-03-21 at 12 54 21

apps/docs/next-sitemap.config.js
monospaced: I think when I force pushed 3 days ago – only to change commit msgs to `docs:` – that must have wiped the sitemap out (maybe by not doing a full build/postbuild).

chrisvxd (approved):

PR #367, opened 2024-03-14
Feature: add label params to component config

bwat-dev:
Provides the option to add a custom label to a component. The label is optional; by default, it will display the component type.


chrisvxd (approved):
This is great, thanks @bwat-dev! I'll add the necessary documentation before releasing this.

PR #368, opened 2024-03-15
fix: Correct inference of objectFields in ObjectField type

hjbrand:
Adjusted the ObjectField type definition to ensure that properties present in Props are required and typed in the objectFields property. This resolves an issue where properties were not being inferred, leading to untyped fields.

Before:
image

After:
image


chrisvxd (approved):
I did wonder why that wasn't working, thanks @hjbrand!

PR #370, opened 2024-03-18
Update multi-column-layouts.mdx

mfakhrys:
Fix categories instead of config attribute


chrisvxd (approved):
Good spot, thanks @mfakhrys!

PR #375, opened 2024-03-23
fix: "Enter" keypress when using ExternalInput together with ArrayFields

### LINKED ISSUE #345: Unexpected behavior for 'Enter' keypress in focused input
https://github.com/measuredco/puck/assets/39839237/ddb5ad50-2408-40d3-87af-e0e994df3fe4

I noticed while editing some components that if I pressed the enter key, it seemed to have a different element focused than the input that I was typing within. In particular, hitting enter when adjusting column span number for columns component, it would delete one of the columns.

jonduarte:
Closes #345

ExternalInput and ArrayFields have button elements which assumed the default type "submit". This causes the buttons to be automatically triggered when "Enter" keypress happens on other inputs of the form.

Setting their type as "button" instead, fixes the behaviour.

### DIFF
--- a/packages/core/components/ExternalInput/index.tsx
+++ b/packages/core/components/ExternalInput/index.tsx
@@ -96,6 +96,7 @@ export const ExternalInput = ({
     >
       <div className={getClassName("actions")}>
         <button
+          type="button"
           onClick={() => setOpen(true)}
           className={getClassName("button")}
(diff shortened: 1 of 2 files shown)


chrisvxd (approved):
Nice one, thanks for the contribution @jonduarte!

PR #380, opened 2024-03-25
chore: fix Button color on dark mode

chrisvxd:
On dark color modes, the secondary Button is currently rendering with dark borders and text, resulting in poor contrast.

Using currentColor addresses this, especially since the main Puck editor does not yet have a dark mode, but the docs do.

Note, this bug is not yet released so a `chore` is the suitable commit type.


monospaced (approved):
Good solution 👍🏻

PR #419, opened 2024-04-04
fix: Fix support for boolean values with select field.

estephan:
When using a boolean value with the select field, the value always evaluates to true (a similar implementation works as expected with the radio field).

puck-select-boolean

https://codesandbox.io/p/sandbox/sleepy-fast-x7cdwn

[Working Radio field code.](https://github.com/measuredco/puck/blob/main/packages/core/components/InputOrGroup/fields/RadioField/index.tsx#L40-L46)


chrisvxd (approved):
Nice catch! Thanks for the contribution.

PR #424, opened 2024-04-08
fix: Reflect value changes made in resolveData in radio field.

estephan:
When a prop change is made in `resolveData`, that value is not reflected when using a radio field but reflected in the render of the component.

In this example, when a change is made to `children` the expected behaviour is that the `align` prop reverts to `"center"`.
Notice that even though the text is aligned in the centre and the prop value has changed, the radio button shows either "left" or "right" as the selected value.

radio-bug

[Bug reproduction ](https://codesandbox.io/p/sandbox/elastic-sutherland-x963k4).

Example with this PR's code change.
radio-working


chrisvxd (approved):
Great catch, thanks @estephan

PR #425, opened 2024-04-09
feat: export InputOrGroup component

dinodeSimon:
Allows the use of puck InputOrGroup component within a custom field


chrisvxd:
Thanks @dinodeSimon, and sorry for the slow delay whilst I've been getting a grip on iframes.

This is on our radar, but is not something we're ready for just yet. Just opened #463 with reasoning.

PR #429, opened 2024-04-10
Add iframe spinner during slow loads

chrisvxd:
If loading takes over 500ms, show a spinner before rendering the iframe.


packages/core/components/Puck/components/Canvas/styles.module.css
```diff
@@ -55,5 +55,24 @@
 
 .PuckCanvas--ready .PuckCanvas-root {
   opacity: 1;
-  transition: opacity 150ms ease-out;
+}
```
chrisvxd, line 58: Not sure we should remove this transition

PR #438, opened 2024-04-14
fix: Adjusted .DropZone--hasChildren CSS class to fix child element

Copystrike:
This change fixes an issue where the layout, a flexbox, was disrupted by the `dropzone` class modifying the children's CSS properties. As a result, the child element, meant to be flexible, was forced to take the entire space. This PR aims to rectify this by restoring the intended behavior. Tested on demo and personal project; no issues so far detected.

322285985-3deddcec-611d-4e2c-adcf-0126124b58c1

### DIFF
--- a/packages/core/components/DropZone/styles.module.css
+++ b/packages/core/components/DropZone/styles.module.css
@@ -49,6 +49,15 @@
   ) !important;
 }
 
+.DropZone--hasChildren {
+  margin-left: initial;
+  margin-right: initial;
+  height: initial;
+  outline-offset: initial;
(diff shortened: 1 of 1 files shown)


Copystrike:
I did actually end up finding an issue where the popup hides under the iframe & what looks like also under the layout?

Need some assistant with this one..

image

chrisvxd:
Hey @Copystrike! Thanks for the PR! Would you mind opening a codepen that reproduces the underlying issue so I can verify I'm understanding correctly?

PR #452, opened 2024-04-17
fix: unified type of render value with onChange value for custom field

AbboDev:
Trying to pass to CustomField any type as props give error because it required an object. The documentation, otherwise, show a working example with a plain input that return a string
with error
suggested


chrisvxd (commented):
Thanks @AbboDev. You've correctly identified some type issues, but this fix just sidesteps the underlying issue by using `any`. I'm going to close this in favour of #462, which should fully infer the type of `value` based on the provided data shape.

PR #468, opened 2024-05-02
Add condition to fields

slb235:
To configure some complexer components, it would be nice to render some fields only when a condition is met.

```
...
    Button: {
      render: (props) => {
         ...
      },
      fields: {
        type: {
          type: 'radio',
          label: 'Type',
          options: [
            { label: 'Text', value: 'text' },
            { label: 'Icon', value: 'icon' },
            { label: 'Icon and Text', value: 'icon-text'}
          ]
        },
        text: {
          type: 'text',
          label: 'Text',
          condition: ({ type }) => type !== 'icon'
        },
        icon: {
          type: 'text',
          label: 'Icon',
          condition: ({ type }) => type !== 'text'
        }
      },
      defaultProps: {
        type: 'text',
        text: 'Button',
        icon: 'arrow-right'
      }
    }
...
```

### DIFF
--- a/packages/core/components/Puck/components/Fields/index.tsx
+++ b/packages/core/components/Puck/components/Fields/index.tsx
@@ -184,6 +184,8 @@ export const Fields = () => {
 
           if (!field?.type) return null;
 
+          if (field.condition && selectedItem && !field.condition(selectedItem?.props)) return null;
+
           const onChange = (value: any, updatedUi?: Partial<UiState>) => {
             let currentProps;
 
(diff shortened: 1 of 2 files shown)


chrisvxd:
Thanks @slb235! We actually have an unreleased API that will cover this use-case in:

https://puckeditor.com/v/canary/docs/integrating-puck/dynamic-fields

See [0.15.0-canary.dd7051e](https://www.npmjs.com/package/@measured/puck/v/0.15.0-canary.dd7051e)

slb235:
That makes even more sense, thank you

PR #469, opened 2024-05-03
Export fields as AutoField

### LINKED ISSUE #463: Consider exporting InputOrGroup (or something with a better name)
Users want to be able to reuse Puck's internal fields (InputOrGroup) in custom fields.

This makes sense, but we're not ready to do that yet:

1. InputOrGroup is a terrible name
2. It massively increases the surface area for breaking API changes
3. We will need to provide documentation

chrisvxd:
Export a new component for rendering Puck style fields inside custom components.

* Renamed InputOrGroup to AutoField
* Tidied up and refactored API
* Fix an iframe style loading issue

Closes #463

## Notes

* 49abe77c101fd0c6188c059231adbc413f940b9e introduces a breaking change
* AutoField is dependant on the Puck context and state, so is not useful for building components _outside_ of `<Puck>`. It is most suited for adding Puck components to custom fields, or other UI within the `<Puck>` context.


apps/docs/pages/docs/extending-puck/custom-fields.mdx
```diff
@@ -128,6 +128,71 @@ const config = {
...
+              <input
+                defaultValue={value}
+                name={name}
+                onChange={(e) => onChange(e.currentTarget.value)}
+                style={{
+                  background: "white",
+                  border: "1px solid black",
+                  padding: 4,
+                }}
+              />
```
chrisvxd, line 181: This example needs updating

packages/core/components/AutoField/index.tsx
```diff
@@ -1,5 +1,6 @@
 import getClassNameFactory from "../../lib/get-class-name-factory";
```
chrisvxd, line 1: This implementation is currently causing a React re-mount whenever the input data is changed when used inside a custom field. Need to investigate further.

packages/core/components/AutoField/index.tsx
```diff
@@ -102,7 +102,10 @@ export type InputProps<F = Field<any>> = {
   readOnlyFields?: Record<string, boolean | undefined>;
 };
 
-export const InputOrGroup = ({ onChange, ...props }: InputProps) => {
+export function AutoField<FieldType extends Field>({
```
monospaced: Change to `function` here deliberate?

apps/docs/pages/docs/api-reference/components/auto-field.mdx
```diff
@@ -0,0 +1,96 @@
...
+};
+```
+
+## Props
+
+| Prop                    | Example                      | Type                                | Status   |
+| ----------------------- | ---------------------------- | ----------------------------------- | -------- |
+| [`field`](#field)       | `{ type: "text" }`           | [Field](/docs/api-reference/fields) | Required |
+| [`onChange`](#onchange) | `onChange("Goodbye, world")` | Function                            | Required |
+| [`value`](#value)       | `"Hello, world"`             | Any                                 | Required |
```
monospaced, line 65: Can `value` be `""`, `null` or `undefined`? Would assume it could need to be, but showing as `Required` here.

packages/core/components/AutoField/index.tsx
chrisvxd: Yes, to support generic types! Newer TS compilers do actually support generic types for arrow functions now, but I still do this by force of habit

monospaced (approved):

PR #479, opened 2024-05-12
wip: add field validation

### LINKED ISSUE #362: Field validation
Currently, there isn't any way to inform or restrict a user from adding undesirable content.  We have multiple fields where we'd like to enforce soft/hard limits e.g. maximum character limits.

We'd need at least four states inform, warn, error, and success where we can change and style different messages such as "Oops you've added 35 characters but only 30 are allowed."

image

Also, it would be great to have the ability to disable the publish button if any field currently is in the error state preventing a user from publishing erroneous content.

chrisvxd:
WIP implementation of field validation.

Usage

```tsx
{
  type: "text",
  validate: (value) => {
    return { type: "error", message: "Something went wrong!" }
  }
}
```

This is non-trivial because it's not clear _when_ to trigger a validation message.

## Tasks

- [ ] Figure out when the validation message should be shown, if at all
- [ ] Abstract out validation from fields
- [ ] Block publish if invalid data, and figure out how to show which data is invalid?

Closes #362


IMythx:
could use shadcn Form component and let the user provide his own zod validation or something and trigger the messages on publish...so the whole editor would be like a form and the publish button would trigger the form submition

chrisvxd:
Stale, closing.

PR #487, opened 2024-05-28
Replace chevrons with undo/redo icons

goynang:
Use undo/redo icons rather than chevrons for moving back/forth in history.


chrisvxd (approved):
🎉  Thanks @goynang!

PR #505, opened 2024-06-07
feat: add initialHistories prop to Puck

mkilpatrick:
Addresses https://github.com/measuredco/puck/issues/492

PR #514, opened 2024-06-20
Streamline history APIs and docs

chrisvxd:
Streamline history APIs and docs, including:

1. Exposing more useful props under the `usePuck` history API, so the user doesn't need to reach into `history.historyStore`
2. Documenting the `usePuck` history API, which should have been done when the feature was implemented
3. Renaming the new `initialHistories` API to `initialHistory` to match the usePuck history API
4. Make references from the new `initialHistory` API to the usePuck `histories` API

Tagging @mkilpatrick for review since you implemented `initialHistories` in #505.

### DIFF
--- a/apps/docs/pages/docs/api-reference/components/puck.mdx
+++ b/apps/docs/pages/docs/api-reference/components/puck.mdx
@@ -25,22 +25,22 @@ export function Editor() {
 
 ## Props
 
-| Param                                   | Example                                           | Type                                               | Status       |
-| --------------------------------------- | ------------------------------------------------- | -------------------------------------------------- | ------------ |
-| [`config`](#config)                     | `config: { components: {} }`                      | [Config](/docs/api-reference/configuration/config) | Required     |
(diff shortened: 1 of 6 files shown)


apps/docs/pages/docs/api-reference/components/puck.mdx
```diff
@@ -182,16 +182,16 @@ Render the Puck preview within iframe. Defaults to `true`.
 
 Disabling iframes will also disable [viewports](#viewports).
 
-### `initialHistories`
+### `initialHistory`
 
-Sets the undo/redo Puck history state.
+Sets the undo/redo Puck history state when using the usePuck [history API](/docs/api-reference/functions/use-puck#history).
```
monospaced, line 187: Should usePuck be `backticked` i.e. `usePuck`? Seems like this would be consistent with other code references in docs.

monospaced (commented):
Apart from minor comment, this looks good to merge ✅

Issue #519, opened 2024-06-25
Expose `setHistories` and `setHistoryIndex`


mkilpatrick:
It would be nice to be able to reset the history so that Puck rerenders with the change. My use case is that I have a clear history button which wipes localStorage/our DB-saved history and I want Puck to also reset fresh with no history. There's currently no way to do that because setHistories() isn't exposed. I can only set initial and then hard refresh the page.

It's worth noting that even though my setup is causing Puck to rerender, it isn't remounted, so even though initialHistory is updated, the underlying histories aren't.

mkilpatrick:
Closed with https://github.com/measuredco/puck/pull/525

PR #525, opened 2024-07-01
feat: expose setHistories and setHistoryIndex

mkilpatrick:
This addresses https://github.com/measuredco/puck/issues/519.

This change exposes the history store's setHistories and setHistoryIndex functions. When they are called the data is dispatched so that Puck rerenders. The use case this supports is when a Puck consumer wants to change or reset the history state manually without requiring the page to be refreshed.

### DIFF
--- a/apps/docs/pages/docs/api-reference/functions/use-puck.mdx
+++ b/apps/docs/pages/docs/api-reference/functions/use-puck.mdx
@@ -129,7 +129,7 @@ const { setHistories } = usePuck();
 setHistories([]); // clears the current history
 ```
 
-### `sethistoryindex`
+### `setHistoryIndex`
 
 A function to set current history index.
 
--- a/apps/docs/pages/docs/api-reference/functions/use-puck.mdx
+++ b/apps/docs/pages/docs/api-reference/functions/use-puck.mdx
@@ -120,7 +120,7 @@ console.log(selectedItem);
 // { type: "Heading", props: {id: "my-heading"} }
 ```
 
-### `sethistories`
+### `setHistories`
 
 A function to set the history state.
 
--- a/apps/docs/pages/docs/api-reference/functions/use-puck.mdx
+++ b/apps/docs/pages/docs/api-reference/functions/use-puck.mdx
@@ -28,12 +28,14 @@ export function Editor() {
 
 ## Returns
 
-| Param                           | Example                                          | Type                                                |
-| ------------------------------- | ------------------------------------------------ | --------------------------------------------------- |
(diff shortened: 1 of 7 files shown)


apps/docs/pages/docs/api-reference/functions/use-puck.mdx
```diff
@@ -117,3 +119,21 @@ The currently selected item, as defined by `appState.ui.itemSelector`.
...
 // { type: "Heading", props: {id: "my-heading"} }
 ```
+
+### `sethistories`
+
+A function to set the history state.
+
+```tsx
+const { setHistories } = usePuck();
+setHistories([]); // clears the current history
```
chrisvxd, line 129:
```suggestion
setHistories([]); // clears all history
```

packages/core/lib/use-puck-history.ts
```diff
@@ -37,6 +39,27 @@ export function usePuckHistory({
...
+  const setHistoryIndex = (index: number) => {
+    if (historyStore.histories.length > index) {
+      dispatch({
+        type: "set",
+        state: historyStore.histories[index]?.data || initialAppState,
+      });
+    }
+
+    historyStore.setHistoryIndex(index);
+  };
```
chrisvxd, line 61: Can you add some unit tests to `__tests__/use-puck-history.spec.ts` for this?

chrisvxd (changes requested):
Great stuff again! Left a few comments.

packages/core/lib/use-puck-history.ts
mkilpatrick: Added, which caught some things 👍

chrisvxd (approved):
Bravo

packages/core/lib/__tests__/use-history-store.spec.tsx
```diff
@@ -89,6 +89,38 @@ describe("use-history-store", () => {
...
+    expect(renderedHook.result.current.currentHistory.data).toBe("Oreo");
+    expect(renderedHook.result.current.index).toBe(0);
+  });
+
+  test("should update index on setHistoryIndex", () => {
+    act(() => renderedHook.result.current.record("Apples"));
+    act(() => renderedHook.result.current.record("Oranges"));
+    act(() => renderedHook.result.current.setHistoryIndex(0));
+
+    expect(renderedHook.result.current.hasPast).toBe(true);
```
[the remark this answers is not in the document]
mkilpatrick: -1 is the no history index

Issue #496, opened 2024-06-05
Making defaultProps an asynchronous function that returns an object?


Roshan54321:
I don't exactly know how I should populate the fields at first render (from external api). I don't want to call an api everytime the appState changes, rather call the api in defaultProps.

Is there a better way that exists?

chrisvxd:
You could probably do something with resolveData and check if any data exists before deciding whether or not to call your API, but perhaps not ideal.

Wonder if a `resolveDefaultProps` API could be a good idea to match `resolveData` and `resolveFields` 🤔

Roshan54321:
> You could probably do something with resolveData and check if any data exists before deciding whether or not to call your API, but perhaps not ideal.
> 
> Wonder if a `resolveDefaultProps` API could be a good idea to match `resolveData` and `resolveFields` 🤔

yes, but I guess an async function that returns the desired object is enough for this.

Roshan54321:
@chrisvxd I was wondering if this is in your priority list. Would be great to have in coming release.

chrisvxd:
It's not on my radar right now, but should be easy to implement with the existing `resolveData` functionality.

chrisvxd:
Closing in favour of #557

Roshan54321:
Hi @chrisvxd. Sorry to mention again. 

Regarding #557, i was thinking about the situation when you would want to give initial data to the component but using async requests. What i understand about resolveData is that it triggers everytime the data changes. I mean, it would be like i want to get 10 items for array type initially and let the user delete, add or edit according to his/her needs, but definitely don't want it to fetch 10 items again and again. 

My concept is to get the initial data from sources and let the user do anything he wants after inserting and dont go with the overhead of resolveData.

I'm not very good with these things, but please correct me.

Roshan54321:
> @Roshan54321 I believe you could configure the resolveData API to only trigger once:
> 
> ```js
> const config = {
>   components: {
>     HeadingBlock: {
>       fields: {
>         title: {
>           type: "text",
>         },
>       },
>       resolveData: async ({ props }) => {
>         if (props.loaded) return;
>        
>         return {
>           props: {
>             title: "Hello, world",
>             loaded: true
>           },
>         };
>       },
>       render: ({ title }) => {
>         return <h1>{title}</h1>;
>       },
>     },
>   },
> };
> ```

[...]

PR #533, opened 2024-07-21
Feature/resolve default props

Roshan54321:
This is for the issue #496 (feature).

How it works?
1. Added `resolveDefaultProps` method in config.  
2. While dispatching `insert` action (i.e. onDragEnd - made it asynchronous too - which probably will need a loader later on), called resolveDefaultProps, passed it to the InsertAction and used it. 

Ps: I did even worse solution in closed PR #532

### DIFF
--- a/apps/docs/pages/docs/api-reference/configuration/component-config.mdx
+++ b/apps/docs/pages/docs/api-reference/configuration/component-config.mdx
@@ -26,14 +26,15 @@ const config = {
 
 ## Params
 
-| Param                                          | Example                                         | Type     | Status   |
-| ---------------------------------------------- | ----------------------------------------------- | -------- | -------- |
-| [`render()`](#renderprops)                     | `render: () => <div />`                         | Function | Required |
-| [`fields`](#fields)                            | `fields: { title: { type: "text"} }`            | Object   | -        |
-| [`defaultProps`](#defaultprops)                | `defaultProps: { title: "Hello, world" }`       | Object   | -        |
-| [`label`](#label)                              | `label: "Heading Block"`                        | String   | -        |
-| [`resolveData()`](#resolvedatadata-params)     | `resolveData: async ({ props }) => ({ props })` | Object   | -        |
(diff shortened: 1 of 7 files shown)


apps/docs/pages/docs/api-reference/configuration/component-config.mdx
```diff
@@ -26,14 +26,15 @@ const config = {
...
+| Param                                                | Example                                         | Type     | Status   |
+| ---------------------------------------------------- | ----------------------------------------------- | -------- | -------- |
+| [`render()`](#renderprops)                           | `render: () => <div />`                         | Function | Required |
+| [`fields`](#fields)                                  | `fields: { title: { type: "text"} }`            | Object   | -        |
+| [`defaultProps`](#defaultprops)                      | `defaultProps: { title: "Hello, world" }`       | Object   | -        |
+| [`label`](#label)                                    | `label: "Heading Block"`                        | String   | -        |
+| [`resolveData()`](#resolvedatadata-params)           | `resolveData: async ({ props }) => ({ props })` | Object   | -        |
+| [`resolveFields()`](#resolvefieldsdata-params)       | `resolveFields: async ({ props }) => ({})`      | Object   | -        |
+| [`resolveDefaultProps()`](#resolvedefaultpropsprops) | `resolveDefaultProps: async (props) => ({})`    | Object   | -        |
```
chrisvxd, line 37: This should be alphabetical, please

chrisvxd (commented):
Thanks for the contribution @Roshan54321!

I think this needs further consideration before we can move ahead with it. There's a lot of overlap between this API and the `resolveData` API, but the `resolveData` API also lets you set fields as read-only.

I'm wondering if we:

1. Reuse the existing `resolveData` API and implementation, triggering it on `insert` somehow
2. or reuse the existing API shape under a new API called `resolveInitialData`

What do you think?

chrisvxd:
Closing this in favour of #557, which should allow the use of resolveData when a new component is inserted.

Issue #543, opened 2024-08-07
Improve resiliency of iframe CSS loading


chrisvxd:
Test Puck's iframe behaviour across multiple CSS frameworks and improve resiliency:

* Tailwind
* Mantine
* ...

Related: #409

chrisvxd:
@xaviemirmon

xaviemirmon:
I've been investigating this issue using the Mantine library and have found two problems that prevent it from rendering all of the styles correctly.  

1. The `AutoFrame` component only mirrors the styles from the document head. Mantine is also injecting styles into the `body`, and these styles aren't getting mirrored. 
2. Mantine is also relying on data attributes in the `html` tag for theme/color mode selection, but this isn't getting copied across.

xaviemirmon:
PR #544 is ready for review.

chrisvxd:
Mantine now working. @xaviemirmon to review further Tailwind setups before closing.

henning-f:
Hello @chrisvxd,

we really appreciate your work and love the puck editor. As we would like to work with serverless infrastructure we started a new project with the remix recipe and added Chakra UI and eventually realized that stylings from the Iframe are not generated and rendered in the editor route. The published site will be rendered as expected.

We tried to enforce client rendering and some other hacks, but did not manage to solve it from the outside. We think this issue is connected with the iframe.

I created a minimal sample for this: https://github.com/henning-f/puck-remix-chakra-issue
I followed the instructions of the [Chakra UI documentation for Remix](https://v2.chakra-ui.com/). 

Are these issues connected? 

I am looking forward to your response.

MiniMarPaz:
Hello.

In our project we are using **styled-components** library for React to style our components. With iframe enabled , those styles are not being picked up by editor. However inline stylings works as expected. Sharing some code example so maybe You can try to reproduce that issue. 

**Editor Config :**

```jsonc
import BlockQuote from './BlockQuote
const config = {
  components: {
    Quote:{
      fields: {
        quote: {
          type: 'text',
        },
      },
      defaultProps: {
        quote: 'Hello, quote',
      },
      render: ({ quote }) => {
        return <BlockQuote quote={quote} />;
      },
    }
  },
};
```

**Component Code**

```jsonc
import styled from 'styled-components';

export default function BlockQuote(props) {
  const { quote } = props;
  return (
    <QuoteWrapper>
      <LineDiv />
      <div className='quote-text'>
        <span>{quote}</span>
      </div>
[...]

PR #544, opened 2024-08-08
#543 improve resiliency of iframe css loading

xaviemirmon:
Fixing the issue where not all styles are synced to the iFrame. Adding sync functionality of attributes from HTML and Body tags to iFrame.

### DIFF
--- a/packages/core/components/AutoFrame/index.tsx
+++ b/packages/core/components/AutoFrame/index.tsx
@@ -37,6 +37,16 @@ const getStyles = (styleSheet?: CSSStyleSheet) => {
   return "";
 };
 
+// Sync attributes from parent window to iFrame
+const syncAttributes = (sourceElement: Element, targetElement: Element) => {
+  const attributes = sourceElement.attributes;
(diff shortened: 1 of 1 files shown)


chrisvxd (approved):
🚀 This looks much more resilient! Thanks @xaviemirmon!

packages/core/components/AutoFrame/index.tsx
```diff
@@ -205,6 +215,14 @@ const CopyHostStyles = ({
     const hrefs: string[] = [];
     let stylesLoaded = 0;
 
+    // Sync attributes for the HTML tag
+    const parentHtml = parentDocument.getElementsByTagName("html")[0];
+    syncAttributes(parentHtml, doc.documentElement);
+
+    // Sync attributes for the Body tag
+    const parentBody = parentDocument.getElementsByTagName("body")[0];
+    syncAttributes(parentBody, doc.body);
```
chrisvxd, line 224: Worth noting that these attributes might change, but this feels like an edge-case that we shouldn't cover just yet.

Issue #537, opened 2024-07-30
TS issues in demo


chrisvxd:
## What happened

The following error was shown in VSCode when cloning the project and running the demo using TS 5.5.4.
```
type `UserConfig` does not satisfy the constraint `Config`
```

Screenshot:

![image](https://github.com/user-attachments/assets/0f35b709-18a4-4fc9-b84c-9d21541a74bc)

## What I expected to happen

No error would be shown.

## Steps to reproduce

1. Clone app
2. Open VSCode
3. Check for TS errors

## Notes

* Could be related to mismatched TS versions (4 vs 5)
* Have not tested build
* User reported issue - I have not attempted to reproduce

giorgio324:
I have similar issue, but instead of passing generic to the component I pass it to the Coinfig Type.
This is my config

```
import { DropZone, type Config } from '@measured/puck';

type Components = {
  HeadingBlock: { title: string };
};
type RootProps = {
  backgroundColor: string;
  fullScreen: '100vh' | '275px';
};

type Categories = 'Text';

export const config: Config<Components, RootProps, Categories> = {
  components: {
    HeadingBlock: {
      fields: {
        title: {
          type: 'text',
        },
      },
      render: ({ title }) => {
        return <h1>{title}</h1>;
      },
    },
  },
  root: {
    fields: {
      backgroundColor: {
        type: 'custom',
        render: ({ name, onChange, value }) => {
          console.log(name, onChange, value);
          return <div></div>;
        },
      },
      fullScreen: {
        label: 'Full Screen',
        type: 'radio',
        options: [
          { label: 'Yes', value: '100vh' },
          { label: 'No', value: '275px' },
        ],
      },
    },
    defaultProps: {
      backgroundColor: '#333333',
      fullScreen: '100vh',
    },
    render: ({ backgroundColor, fullScreen }) => {
      return (
        <main
          style={{
            backgroundColor: backgroundColor,
            minHeight: fullScreen,
          }}
        >
          <DropZone zone='body' style={{ minHeight: fullScreen }} />
        </main>
      );
    },
  },
};

export default config;
```
when I import this config and use it in either Puck or Render i get error

Type 'Config<Components, RootProps, "Text">' is not assignable to type 'Config'.
  Types of property 'components' are incompatible.
[...]

PR #545, opened 2024-08-08
Help out the generics by ensuring that objectFields is never an array.

xaviemirmon:
I narrowed the TypeScript errors to this PR #368

Because the ObjectField is `any` objectFields _could_ be an array here:

 [SubPropName in keyof Props]: Field<Props[SubPropName]>;
 
 I've added a check so that it can only be an object which gets rid of the the Config mismatch errors and keeps typing within the objectFields

### DIFF
--- a/packages/core/types/Fields.ts
+++ b/packages/core/types/Fields.ts
@@ -52,9 +52,11 @@ export type ObjectField<
   Props extends { [key: string]: any } = { [key: string]: any }
 > = BaseField & {
   type: "object";
-  objectFields: {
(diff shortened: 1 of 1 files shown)


chrisvxd (approved):
This works great for #537! However, I can't verify it reproduces related issues [from Discord](https://discord.com/channels/1153376562259951687/1153376562830397563/1270860176215244870) as I cannot reproduce. This is likely a separate issue, and so we should ask the users to test this fix once merged and open a ticket if the issue persists.

Also opened #546 as we need to upgrade TS to actually reproduce the #537.

PR #550, opened 2024-08-12
Adding a default favicon to the demo app

### LINKED ISSUE #259: Favicon getting caught by catch-all in next recipe
The favicon is getting caught by the catch-all route in the next recipe.

We should add a placeholder favicon, which will prevent the catch-all route from getting triggered.

xaviemirmon:
Closes #259


chrisvxd (changes requested):
Thanks @xaviemirmon! The original issue actually relates to the favicon conflicting in the `next` _recipe_, rather than the demo app. I imagine this solution would fix that, too.

xaviemirmon:
I got it, @chrisvxd. I'll also update the branch to account for this in the recipe.

xaviemirmon:
@chrisvxd I've added this to the recipe also!

chrisvxd (approved):
🎉

PR #551, opened 2024-08-12
Removing the stroke from the Lucide icon

### LINKED ISSUE #330: Undo/redo buttons do not inherit interactive state colors from `IconButton`
Because [color styles are being applied at the SVG stroke level](https://github.com/measuredco/puck/blob/22a41823559d36fd06842496d59788004b316797/packages/core/components/MenuBar/index.tsx#L64), the undo/redo buttons in `MenuBar` do not consistently apply the default interactive `:hover`/`:active` color styles (which should be inherited from `IconButton`). 

See [sidebar toggle buttons](https://github.com/measuredco/puck/blob/22a41823559d36fd06842496d59788004b316797/packages/core/components/Puck/index.tsx#L430) for the correct implementation (styles applied to a parent element). This way, we are able to manage different default colors (dark/light grey, depending on state), whist still allowing the interactive states to be inherited.

xaviemirmon:
Removing the stroke from the Lucide icon as the fill can be inherited by the parent icon button CSS. 

Fixes #330

### DIFF
--- a/packages/core/components/MenuBar/index.tsx
+++ b/packages/core/components/MenuBar/index.tsx
@@ -60,21 +60,12 @@ export const MenuBar = ({
           <IconButton title="undo" disabled={!hasPast} onClick={back}>
             <Undo2Icon
               size={21}
-              stroke={
-                hasPast
-                  ? "var(--puck-color-black)"
-                  : "var(--puck-color-grey-08)"
-              }
+              
             />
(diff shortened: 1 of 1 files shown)


xaviemirmon:
Yep, it's my fault, sorry! This is fixed now @chrisvxd

chrisvxd (approved):

PR #538, opened 2024-07-31
Add on action prop

nova4u:
Addresses https://github.com/measuredco/puck/issues/534

This change adds an `onAction` callback to the Puck editor. When an action occurs in the editor, this callback is called with the `action`, `appState`, and `prevAppState`. Enables tracking changes, performing side effects, or syncing with external systems.

PR #552, opened 2024-08-12
type: feature Add `onSelectionChange` callback that emits the currently selected component

ProGM:
Hello there!
This is my first PR on puck, I hope you'll find it valuable ;)

This pull request adds a `onSelectionChange` callback to the Puck component that emits the id and the type of the currently selected component in the editor.

When no component is selected (i.e. you click outside of the editor), it returns `null`.

It's very useful to extend puck!

Example usage:
```typescript
<Puck
  ...
  onSelectionChange={(item) => { console.log(item) }}
  ...
/>
```

### DIFF
--- a/packages/core/components/Puck/index.tsx
+++ b/packages/core/components/Puck/index.tsx
@@ -78,7 +78,9 @@ export function Puck<UserConfig extends Config = Config>({
   ui?: Partial<UiState>;
   onChange?: (data: Data) => void;
   onPublish?: (data: Data) => void;
-  onSelectionChange?: (itemSelector: { id: string, type: string } | null) => void;
+  onSelectionChange?: (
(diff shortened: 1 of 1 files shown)


chrisvxd (commented):
Thanks for the contribution @ProGM!

I suspect you might actually be able to achieve this with the new [`onAction` API](https://puckeditor.com/v/canary/docs/api-reference/components/puck#onactionaction-appstate-prevappstate) that just got merged (#538). Although there's no explicit action for selecting an item, you can figure it out by comparing the current and previous state.

Please test out the [0.16.0-canary.2c52d27](https://www.npmjs.com/package/@measured/puck/v/0.16.0-canary.2c52d27) release and let me know if it solves your use case.

Issue #461, opened 2024-04-23
Add permissions API and actionBar override


chrisvxd:
This ticket covers two features:

1. Adding an API to override the action bar and inject new items
2. Adding a new permissions API to control key parts of functionality for:
  * `drag`: whether the component can be dragged
  * `edit`: whether or not fields are editable (read-only if not)
  * `delete`: whether the component can be deleted
  * `duplicate`: whether the component can be duplicated
  * `insert`: whether the component can be inserted
  * ... TBD

## Tasks

- [x] Export Action Bar #563 
- [x] Add actionBar override #action bar #570
- [x] Implement permissions API #572
- [x] Configure default permissions from (2) above #574
- [x] Implement resolvePermissions API  #581 
- [ ] #579 
- [ ] #580
- [ ] #584 
- [ ] #582
- [ ] Documentation

### Injecting new items via actionBar override

```tsx
import { ActionBar, Puck } from "@measured/puck";

const data = {};

export function Editor() {
  return (
    <Puck
      data={data}
      overrides={{
        // Enable override of entire action bar, not just internal actions
        actionBar: ({ children, item, itemSelector }) => {
          return (
            // <ActionBar> is the default UI
            <ActionBar>
              <div
                style={{
                  display: "flex",
                }}
              >
                <ActionBar.Item>My custom item</ActionBar.Item>
                {children}
            </ActionBar>
          );
        },
      }}
    />
  );
}
```

### Adding `permissions` and `resolvePermissions` APIs

```tsx
import { usePuck } from "@measured/puck";

const data = {};

export function Editor() {
  const permissions = {
    duplicate: false,
    edit: true,
    // A custom permission for the user
    party: false,
  };
  
  return (
    <Puck
      // Specify global permissions
      permissions={permissions}

      config={{
        components: {
          MyComponent: {
            // Basic usage
            permissions: { edit: false },

            // Advanced usage
            resolvePermissions: async ({
              props,
              lastPermissions,
              initialPermissions,
            }) => {
              // Do some async task
              const isAdmin = await getAdmin();
              
              return {
                party: isAdmin
               };
            },
          },
        },
      }}
      data={data}
    />
  );
}
```

sagarchoudhary96:
@chrisvxd  do we have any updates on this one ? 

[...]

PR #555, opened 2024-08-12
Add `dnd.isDisabled` option in the Puck component to disable drag and drop

ProGM:
Hi there!

Here I am again with a new PR.

This adds the option to disable the drag and drop feature of the editor.

Usage:
```typescript
export function Editor() {
  return (
    <Puck
       dnd={{ isDisabled: true }}
       data={{ ... }}
        /*...*/
    />
  );
}
```

### DIFF
--- a/packages/core/components/DraggableComponent/index.tsx
+++ b/packages/core/components/DraggableComponent/index.tsx
@@ -109,7 +109,7 @@ export const DraggableComponent = ({
           style={{
             ...style,
             ...provided.draggableProps.style,
-            cursor: isModifierHeld ? "initial" : "grab",
+            cursor: isModifierHeld || isDragDisabled ? "initial" : "grab",
           }}
           onMouseOver={onMouseOver}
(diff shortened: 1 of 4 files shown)


ProGM:
Auto-bug report:
when I disable drag and drop using the attribute I've added, the side outline section starts to break with this error:

Screenshot 2024-08-14 at 17 41 03

The problem is here:
https://github.com/measuredco/puck/blob/main/packages/plugin-heading-analyzer/src/HeadingAnalyzer.tsx#L167

When the drag and drop is disabled, the `[${dataAttr}="${props.analyzeId}"]` selector returns a broken item without the style set. Weird, it looks like the heading analyzer is strictly dependent to the drag and drop library.

chrisvxd:
Hey @ProGM! Thanks again for another wonderful contribution. However, this one is just about to be added as part of the broader `permissions` API in #461!

Appreciate the support

PR #561, opened 2024-08-14
feat: Allow styles and markup to be injected into the Puck iframe

### LINKED ISSUE #409: iframe style injection
## Description

Users want to be able to inject styles into iframes for some use-cases.

We should leverage the iframe API to support this.

### Option 1

Consider leveraging the `initialContent` API from [`auto-frame-component`](https://github.com/measuredco/auto-frame-component):

```tsx
<Puck iframe={{
  initialContent="<!DOCTYPE html><html><head><style>body { background: hotpink; }</style></head><body><div></div></body></html>"
}} {/* ... */} />
```

### Option 2

Use a stripped back version of `initialContent` API from [`auto-frame-component`](https://github.com/measuredco/auto-frame-component):

```tsx
<Puck iframe={{
  head="<style>body { background: hotpink; }</style>"
}} {/* ... */} />
```

### Option 3

Add a dedicated parameters

```tsx
<Puck iframe={{
  style="body { background: hotpink; }"
  styleUrl="https://www.example.com/styles.css"
}} {/* ... */} />
```

xaviemirmon:
Closes #409


packages/core/types/IframeConfig.tsx
```diff
@@ -1,3 +1,8 @@
+import { ReactNode } from "react";
+import { RenderFunc } from "./Overrides";
+
 export type IframeConfig = {
   enabled?: boolean;
+  initialContent?: string;
+  wrapper?: RenderFunc<{ children: ReactNode }>;
```
chrisvxd, line 7: I reckon this should be called `render` and we should pass in the `document`

```suggestion
  render?: RenderFunc<{ children: ReactNode; document: Document }>;
```

packages/core/types/IframeConfig.tsx
```diff
@@ -1,3 +1,8 @@
+import { ReactNode } from "react";
+import { RenderFunc } from "./Overrides";
+
 export type IframeConfig = {
   enabled?: boolean;
+  initialContent?: string;
```
chrisvxd, line 6: Is there still a valid use-case for this if we're exposing a render function?

chrisvxd (changes requested):
I'm still not totally sure why the styles aren't syncing from the host, but it's likely some combination of factors. On balance, I think the custom render method seems to work well and enable emotion cache injection when necessary, so nice one.

Left a few thoughts inline.

chrisvxd:
Closing in favour of #587

PR #562, opened 2024-08-14
fix: initialHistory not reflected in appState

mkilpatrick:
`initialHistory` was recently added which applies a set of histories on top of the Puck `data`. However, the appState was not set to show the data at the `index` of `initialHistory.histories`. This meant that a user would have to click the "redo" button to get the Puck state back to where they previously left off in the histories array.


chrisvxd:
Closing in favour of #600

PR #564, opened 2024-08-15
refactor: move override logic to lib hooks

xaviemirmon:
Linked to #461


packages/core/lib/use-default-render.tsx
```diff
@@ -0,0 +1,11 @@
...
+
+export const useDefaultRender = () => {
+  return useMemo<React.FunctionComponent<{ children?: ReactNode }>>(() => {
+    const PuckDefault = ({ children }: { children?: ReactNode }) => (
+      <>{children}</>
+    );
+
+    return PuckDefault;
+  }, []);
+};
```
chrisvxd, line 11: I can't remember why I used a memo here and didn't just move this to the global scope... possibly just lazy, but possibly some reason.

It might be easier to just have this as a component

```ts
import { ReactNode, useMemo } from "react";

export const DefaultRender = () => {
  const PuckDefault = ({ children }: { children?: ReactNode }) => (
    <>{children}</>
  );

  return PuckDefault;
};
 ```

chrisvxd (commented):
Looks good but left one (non-blocking) comment

packages/core/lib/use-default-render.tsx
xaviemirmon: That should be moved to the component directory?

chrisvxd: That's fine, but maybe it needs a better name... `DefaultOverride`?

xaviemirmon: I've moved to a component and tested custom ui and couldn't see any problems.

chrisvxd (approved):

PR #570, opened 2024-08-19
feat: add actionBar override for adding component controls

xaviemirmon:
Linked to: #461


packages/core/types/Overrides.ts
```diff
@@ -27,6 +27,11 @@ type OverridesGeneric<Shape extends { [key in OverrideKey]: any }> = Shape;
 export type Overrides = OverridesGeneric<{
   fieldTypes: Partial<FieldRenderFunctions>;
   header: RenderFunc<{ actions: ReactNode; children: ReactNode }>;
+  actionBar: RenderFunc<{
+    itemSelector: ItemSelector | null;
```
chrisvxd, line 31: Doesn't like this is used anymore

packages/core/components/DraggableComponent/index.tsx
```diff
@@ -86,6 +89,22 @@ export const DraggableComponent = ({
...
+  const loadedOverrides = useLoadedOverrides({
+    overrides: overrides,
+    plugins: plugins,
+  });
+
+  const DefaultActionBar = () => (
+    <ActionBar label={label}>
+      <DefaultOverride />
+    </ActionBar>
+  );
```
chrisvxd, line 101: Can this be pulled out to the top of the file, with `label` passed in as a prop?

packages/core/components/ActionBar/ActionBar.tsx
```diff
@@ -0,0 +1,33 @@
...
+}: {
+  children: ReactNode;
+  onClick: (e: SyntheticEvent) => void;
+}) => (
+  <button className={getClassName("action")} onClick={onClick}>
+    {children}
+  </button>
+);
+
+ActionBar.Action = Action;
```
chrisvxd, line 33: Not sure why this is showing up on the diff - suspect you might need a rebase

packages/core/components/DraggableComponent/index.tsx
```diff
@@ -86,6 +89,22 @@ export const DraggableComponent = ({
...
+
+  const CustomActionBar = useMemo(
+    () => loadedOverrides.actionBar || DefaultActionBar,
+    [loadedOverrides]
+  );
+
```
chrisvxd, line 107: Could you add an example to the `custom-ui` implementation? It's useful for both reviewing, but also checking this functionality builds

chrisvxd (changes requested):

packages/core/components/DraggableComponent/index.tsx
xaviemirmon: Done! I've added a debug info action

packages/core/components/DraggableComponent/index.tsx
xaviemirmon: Done

packages/core/components/ActionBar/ActionBar.tsx
xaviemirmon: Fixed.  It was an old file that didn't get removed

packages/core/types/Overrides.ts
xaviemirmon: Removed

apps/demo/app/custom-ui/[...puckPath]/client.tsx
```diff
@@ -276,6 +276,35 @@ const CustomPuck = ({ dataKey }: { dataKey: string }) => {
...
+  } else {
+    data =
+      appState.data.zones[appState.ui.itemSelector.zone][
+        appState.ui.itemSelector.index
+      ];
+  }
+
+  const onClick = () => {
+    alert(
+      `ID: ${appState.ui.itemSelector.index} \nZone: ${
```
Comment (chrisvxd) [inline apps/demo/app/custom-ui/[...puckPath]/client.tsx:293]: Tiny, tiny point.. but this is the `index` and not the `ID` 😄 

```suggestion
      `Index: ${appState.ui.itemSelector.index} \nZone: ${
```

apps/demo/app/custom-ui/[...puckPath]/client.tsx
```diff
@@ -276,6 +276,35 @@ const CustomPuck = ({ dataKey }: { dataKey: string }) => {
   );
 };
 
+const CustomActionBar = ({ children, label }) => {
+  const { appState } = usePuck();
```
Comment (chrisvxd) [inline apps/demo/app/custom-ui/[...puckPath]/client.tsx:280]: You can use `selectedItem` from usePuck instead of reaching into the state yourself

```suggestion
  const { appState, selectedItem } = usePuck();
```

chrisvxd (changes requested):
This is fantastic! One a couple of tiny comments

PR #572, opened 2024-08-20
feat: adding permissions to config and getPermission function

xaviemirmon:
Linked to #461


packages/core/lib/use-puck.ts
```diff
@@ -13,6 +15,14 @@ export const usePuck = () => {
     appState,
     config,
     dispatch,
+    getPermissions: (permissions: string[]) => {
```
chrisvxd, line 18: I think we should tweak this API a bit:

1. Drop the `permissions` param and just return all permissions by default
2. Accept an optional `itemSelector` param in case the user wants to get permissions for a non-selected item

packages/core/lib/get-permissions.ts
```diff
@@ -0,0 +1,48 @@
...
+  let componentPermissions = { ...state.ui.globalPermissions };
+
+  if (config.components[selectedItem.type] !== undefined) {
+    componentPermissions = {
+      ...componentPermissions,
+      ...config.components[selectedItem.type].permissions,
+    };
+  }
+
+  return componentPermissions;
```
chrisvxd, line 47: You don't need this conditional check. Suggest this could be simplified -

```suggestion
  return componentPermissions = {
    ...globalPermissions,
    ...config.components[selectedItem.type]?.permissions
  };
   ```

File: packages/core/lib/get-permissions.ts [hunk 3]
```diff
@@ -0,0 +1,48 @@
...
+        .reduce((obj, key) => {
+          obj[key] = componentPermissions[key] ?? null;
+          return obj;
+        }, {} as { [key: string]: boolean | null })
+    : null;
+
+  return computedPermissions;
+};
+
+export const initialPermissions = ({
```
chrisvxd, line 29: Prefer to keep functions as verbs -

```suggestion
export const getInitialPermissions = ({

File: packages/core/lib/get-permissions.ts [hunk 4]
```diff
@@ -0,0 +1,48 @@
...
+  const computedPermissions = componentPermissions
+    ? Object.keys(componentPermissions)
+        .filter((key) => permissions.includes(key))
+        .reduce((obj, key) => {
+          obj[key] = componentPermissions[key] ?? null;
+          return obj;
+        }, {} as { [key: string]: boolean | null })
+    : null;
+
+  return computedPermissions;
```
chrisvxd, line 26: I feel like this can be simplified a fair bit if we drop the `permissions` array parameter

chrisvxd (changes requested):

apps/demo/app/custom-ui/[...puckPath]/client.tsx
```diff
@@ -309,6 +314,10 @@ export function Client({ path, isEdit }: { path: string; isEdit: boolean }) {
         data={data}
         iframe={{ enabled: false }}
         headerPath={path}
+        permissions={{
+          debug: true,
+          test: true,
+        }}
```
Comment (chrisvxd) [inline apps/demo/app/custom-ui/[...puckPath]/client.tsx:320]: Might be nice to showcase this working fully by disabling the `debug` permission on a specific component

packages/core/components/Puck/index.tsx
```diff
@@ -370,6 +378,22 @@ export function Puck<UserConfig extends Config = Config>({
     setMounted(true);
   }, []);
 
+  // Add global permissions to component config
```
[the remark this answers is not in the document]
xaviemirmon: This is fixed now

packages/core/lib/get-permissions.ts
xaviemirmon: Yeah deffo.

apps/demo/app/custom-ui/[...puckPath]/client.tsx
Comment (xaviemirmon) [reply apps/demo/app/custom-ui/[...puckPath]/client.tsx:320]: I didn't only on the basis that it would be in the config that is shared for custom-ui and demo

packages/core/types/Config.tsx
```diff
@@ -172,3 +173,5 @@ export type UiState = {
 };
 
 export type AppState = { data: Data; ui: UiState };
+
+export type Permissions = Record<string, boolean>;
```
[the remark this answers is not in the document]
xaviemirmon: Should we also pass these into the context?

apps/demo/app/custom-ui/[...puckPath]/client.tsx
Comment (xaviemirmon) [reply apps/demo/app/custom-ui/[...puckPath]/client.tsx:330]: Created an override for this

packages/core/lib/get-permissions.ts
xaviemirmon: Tidied this up

packages/core/lib/use-puck.ts
xaviemirmon: I changed this to componentType as that was all that was required by the logic.

packages/core/lib/get-permissions.ts
```diff
@@ -0,0 +1,38 @@
...
+  globalPermissions: Partial<Permissions>;
+  config: Config;
+}) => {
+  let componentPermissions = getInitialPermissions({
+    componentType,
+    config,
+    globalPermissions,
+  });
+
+  return componentPermissions;
```
chrisvxd, line 18: I guess this is a `let` because it's going to be enhanced to add the resolver?

packages/core/lib/get-permissions.ts
```diff
@@ -0,0 +1,38 @@
...
+  config: Config;
+}) => {
+  let componentPermissions: Partial<Permissions> = { ...globalPermissions };
+
+  componentPermissions = {
+    ...componentPermissions,
+    ...config.components[componentType].permissions,
+  };
+
+  return componentPermissions;
```
chrisvxd, line 37: Feel like there's no need to use a `let` here and you could just return the object directly

```suggestion
   return {
     ...componentPermissions,
     ...config.components[componentType].permissions,
   };
```

packages/core/types/Config.tsx
```diff
@@ -147,6 +148,7 @@ export type ItemWithId = {
 export type ArrayState = { items: ItemWithId[]; openId: string };
 
 export type UiState = {
+  globalPermissions?: Partial<Permissions>;
```
chrisvxd, line 151: This is no longer needed

chrisvxd (changes requested):

packages/core/lib/get-permissions.ts
xaviemirmon: Correct.

chrisvxd (approved):
🚀

Issue #556, opened 2024-08-13
New drag-and-drop engine


chrisvxd:
Introduce a new drag-and-drop engine to resolve numerous issues:

1. #123
2. #389
3. #455
4. #520

### Proposal 1 - dnd-kit (preferred)

The new experimental dnd-kit release, whilst in early alpha, may resolve all issues 1 - 4. This is the solution currently being explored.

Being implemented here: https://github.com/measuredco/puck/pull/598

### Proposal 2 - custom drag-and-drop library

Creating a custom drag-and-drop library may be more suitable for Puck's needs, and a POC has been created, but significant additional effort is needed to support 3 and 4.

## Cross-cutting concerns

* iframe support should be retained
* retain existing animation curves

chrisvxd:
@Anishali2 I appreciate that, but we're pretty far down the road now in #598 and just ironing out the last few bits with the dnd-kit team!

If you have any expertise in the `experimental` brand of dnd-kit, I may have some questions for you in the Discord!

Sarfraz-droid:
You can also consider using react-dnd. I had some issues while using dndkit between  iframes. Since it does not uses HTML Dnd Api

chrisvxd:
@Sarfraz-droid we're sticking with dnd-kit now, but we're using the experimental build and actively contributing to it.

We require a lot of very complex behaviour that most dnd libraries just don't solve for (iframes is obviously one, but there's a ton more like nested dragging, animations, inline CSS, transforms, placeholders, etc).

PR #574, opened 2024-08-21
feat: adding permissions to core actions and disabling on Hero

xaviemirmon:
Linked to #461

### DIFF
--- a/apps/demo/app/custom-ui/[...puckPath]/client.tsx
+++ b/apps/demo/app/custom-ui/[...puckPath]/client.tsx
@@ -279,7 +279,7 @@ const CustomPuck = ({ dataKey }: { dataKey: string }) => {
 const CustomActionBar = ({ children, label }) => {
   const { appState, getPermissions, selectedItem } = usePuck();
 
-  const { debug } = getPermissions();
+  const { debug } = getPermissions({});
 
   const onClick = () => {
     alert(
@@ -340,6 +340,7 @@ export function Client({ path, isEdit }: { path: string; isEdit: boolean }) {
             <CustomActionBar label={label}>{children}</CustomActionBar>
           ),
           components: () => {
+            const { getPermissions } = usePuck();
             return (
               <Drawer direction="horizontal">
                 <div
@@ -351,23 +352,29 @@ export function Client({ path, isEdit }: { path: string; isEdit: boolean }) {
                   }}
                 >
                   {Object.keys(config.components).map(
-                    (componentKey, componentIndex) => (
-                      <Drawer.Item
-                        key={componentKey}
-                        name={componentKey}
-                        index={componentIndex}
-                      >
-                        {({ children }) => (
-                          <div
-                            style={{
-                              marginRight: 8,
-                            }}
-                          >
-                            {children}
-                          </div>
-                        )}
-                      </Drawer.Item>
-                    )
+                    (componentKey, componentIndex) => {
+                      const canInsert = getPermissions({
+                        type: componentKey,
+                      }).insert;
+                      return (
+                        <Drawer.Item
+                          key={componentKey}
+                          name={componentKey}
+                          index={componentIndex}
+                          isDragDisabled={!canInsert}
+                        >
+                          {({ children }) => (
+                            <div
+                              style={{
+                                marginRight: 8,
+                              }}
+                            >
+                              {children}
+                            </div>
+                          )}
+                        </Drawer.Item>
+                      );
+                    }
                   )}
                 </div>
               </Drawer>
--- a/apps/demo/app/custom-ui/[...puckPath]/client.tsx
+++ b/apps/demo/app/custom-ui/[...puckPath]/client.tsx
@@ -301,13 +301,52 @@ const CustomActionBar = ({ children, label }) => {
   );
 };
 
+const CustomDrawer = () => {
+  const { getPermissions } = usePuck();
+
+  return (
+    <Drawer direction="horizontal">
+      <div
+        style={{
+          display: "flex",
+          pointerEvents: "all",
+          padding: "16px",
+          background: "var(--puck-color-grey-12)",
+        }}
+      >
+        {Object.keys(config.components).map((componentKey, componentIndex) => {
+          const canInsert = getPermissions({
+            type: componentKey,
+          }).insert;
+          return (
+            <Drawer.Item
+              key={componentKey}
+              name={componentKey}
+              index={componentIndex}
+              isDragDisabled={!canInsert}
+            >
+              {({ children }) => (
+                <div
+                  style={{
+                    marginRight: 8,
+                  }}
+                >
+                  {children}
+                </div>
+              )}
+            </Drawer.Item>
+          );
+        })}
+      </div>
+    </Drawer>
+  );
+};
+
 export function Client({ path, isEdit }: { path: string; isEdit: boolean }) {
   const { data, resolvedData, key } = useDemoData({
     path,
     isEdit,
   });
-
-  const { getPermissions } = usePuck();
 
   const configOverride: UserConfig = {
     ...config,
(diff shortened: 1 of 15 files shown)


packages/core/components/DraggableComponent/index.tsx
```diff
@@ -18,6 +18,7 @@ import { Loader } from "../Loader";
 import { ActionBar } from "../ActionBar";
 import { DefaultOverride } from "../DefaultOverride";
 import { useLoadedOverrides } from "../../lib/use-loaded-overrides";
+import { getPermissions } from "../../lib/get-permissions";
```
xaviemirmon, line 21: @chrisvxd, is this the right approach here?  Using the lib not the usePuck hook?

packages/core/components/DraggableComponent/index.tsx
```diff
@@ -109,12 +118,20 @@ export const DraggableComponent = ({
...
+      config,
+    });
+
   return (
     <El
       key={id}
       draggableId={id}
       index={index}
-      isDragDisabled={isDragDisabled}
+      isDragDisabled={!permissions?.drag || isDragDisabled}
```
xaviemirmon, line 134: @chrisvxd This isn't working if another component is selected and I drag the Hero in the custom UI. How can I get the dragging component's state, not the selected one?

packages/core/components/DraggableComponent/index.tsx
```diff
@@ -163,12 +180,17 @@ export const DraggableComponent = ({
                 }}
               >
                 <CustomActionBar label={label}>
```
chrisvxd, line 182: I think we need to adjust the visuals of the ActionBar when there are no actions inside, as the current UI collapses down with poor spacing and an unnecessary separator.

image

Additionally, we should provide some kind of indication that the component cannot be dragged.

I'm happy to do these as a follow up.

chrisvxd (approved):

packages/core/components/DraggableComponent/index.tsx
xaviemirmon: Ahh yes, I did the Action Bar already but I think it was in the original PR we didn't merge..

What are you thinking as an indication that the component cannot be dragged

chrisvxd: I think for now perhaps just changing the pointer to `pointer` instead of `grab` is enough.

I plan on adding grab handles to the action bar in #556, which we could then add a little 🔒 illustration to.

packages/core/components/DropZone/index.tsx
```diff
@@ -237,6 +238,15 @@ function DropZoneEdit({ zone, allow, disallow, style }: DropZoneProps) {
...
                   componentConfig?.["label"] ?? item.type.toString();
 
+                const canDrag = getPermissions({
+                  selectedItem: getItem(
+                    { index: i, zone: zoneCompound },
+                    appContext.state.data
+                  ),
+                  config: appContext.config,
+                  globalPermissions: appContext.globalPermissions || {},
+                }).drag;
```
chrisvxd, line 248: I think we should remove the `drag` permissions based on our conversation yesterday

If you disable drag on an item, you don't disable drag on the other items, so it serves no purpose.

chrisvxd (changes requested):

apps/demo/app/custom-ui/[...puckPath]/client.tsx
```diff
@@ -314,6 +314,9 @@ export function Client({ path, isEdit }: { path: string; isEdit: boolean }) {
         ...config.components.Hero,
         permissions: {
           debug: false,
+          drag: false,
+          delete: false,
+          duplicate: false,
```
Comment (chrisvxd) [inline apps/demo/app/custom-ui/[...puckPath]/client.tsx:319]: Can you add the `insert` permission from #461?

chrisvxd (changes requested):

packages/core/components/Drawer/index.tsx
```diff
@@ -22,22 +27,35 @@ const DrawerDraggable = ({
...
-    disableAnimations
-    className={() => getClassNameItem()}
-  >
-    {() => children}
-  </Draggable>
-);
+  type: keyof DefaultComponentProps;
+}) => {
+  const appContext = useAppContext();
+
```
xaviemirmon, line 38: @chrisvxd it's this code that is causing the docs site to fail build.  Any ideas why?

chrisvxd: I think that's because the Docs doesn't have the app context

Drawer is (sort of) designed to be general purpose and usable outside of the Puck context

I think I would move this logic out of Drawer and into the ComponentList, passing in `isDragDisabled` as a prop when using Drawer.

packages/core/lib/use-puck.ts
```diff
@@ -16,9 +16,13 @@ export const usePuck = () => {
     appState,
     config,
     dispatch,
-    getPermissions: (selectedItem?: ComponentData) => {
+    getPermissions: (
+      selectedItem?: ComponentData,
+      type?: keyof DefaultComponentProps
+    ) => {
```
chrisvxd, line 22: We might wanna tweak this API a bit to so users aren't having to do `getPermissions(null, "Hero")`, but I think let's do that as part of the documentation to keep things moving.

packages/core/lib/use-puck.ts
```diff
@@ -16,9 +16,13 @@ export const usePuck = () => {
     appState,
     config,
     dispatch,
-    getPermissions: (selectedItem?: ComponentData) => {
+    getPermissions: (
+      selectedItem?: ComponentData,
+      type?: keyof DefaultComponentProps
```
chrisvxd, line 21: The `usePuck` hook should probably take a generic type of `UserConfig` (see other examples) so that the `type` can be fully typed against the user's actual components.

packages/core/components/Drawer/index.tsx
```diff
@@ -22,35 +22,42 @@ const DrawerDraggable = ({
...
+  );
+};
 
 const DrawerItem = ({
   name,
   children,
   id,
   label,
   index,
+  canInsert,
```
chrisvxd, line 53: I think we should call this `isDragDisabled` rather than `canInsert`, as this is (sort-of) designed to be used in other contexts that aren't necessarily about _inserting_ elements.

apps/demo/app/custom-ui/[...puckPath]/client.tsx
```diff
@@ -314,6 +314,10 @@ export function Client({ path, isEdit }: { path: string; isEdit: boolean }) {
...
+          insert: false,
```
Comment (chrisvxd) [inline apps/demo/app/custom-ui/[...puckPath]/client.tsx:320]: This seems to be disabling the `insert` behaviour of all items in the custom UI, not just the Hero

packages/core/components/Drawer/styles.module.css
```diff
@@ -20,7 +20,6 @@
   font-size: var(--puck-font-size-xxs);
   justify-content: space-between;
   align-items: center;
-  cursor: grab;
   transition: background-color 50ms ease-in, color 50ms ease-in;
 }
```
chrisvxd, line 25: Happy to do this as a follow-up, but we should probably add a disabled state here, to give the user a visual indication this is not draggable

```css
.DrawItem--disabled {
  background: var(--puck-color-grey-11);
  color: var(--puck-color-grey-05);
  cursor: not-allowed; /** Move this out of inline styles */
}
```

And make sure the hover state doesn't occur when disabled

```css
.Drawer:not(.Drawer--isDraggingFrom) .DrawerItem-draggable:hover:not(.DrawItem--disabled) {
```

image

chrisvxd (changes requested):
Awesome progress! Another round of feedback for ya

packages/core/lib/use-puck.ts
xaviemirmon: @chrisvxd I can move to an object which would help?

```tsx
getPermissions: ({
      selectedItem,
      type,
    }: {
      selectedItem?: ComponentData;
      type?: keyof DefaultComponentProps;
    }) => {
      return getPermissions({
        selectedItem: selectedItem || currentItem,
        type,
        globalPermissions: globalPermissions || {},
        config,
      });
    },
```

apps/demo/app/custom-ui/[...puckPath]/client.tsx
```diff
@@ -279,7 +279,7 @@ const CustomPuck = ({ dataKey }: { dataKey: string }) => {
 const CustomActionBar = ({ children, label }) => {
   const { appState, getPermissions, selectedItem } = usePuck();
 
-  const { debug } = getPermissions();
+  const { debug } = getPermissions({});
```
Comment (chrisvxd) [inline apps/demo/app/custom-ui/[...puckPath]/client.tsx:282]: We should default this so we don't have to provide an empty object

apps/demo/app/custom-ui/[...puckPath]/client.tsx
Comment (chrisvxd) [reply apps/demo/app/custom-ui/[...puckPath]/client.tsx:362]: @xaviemirmon something weird is going on here... steps to reproduce in custom UI demo

1. Drag in a non-disabled item
2. Disabled item (Hero) becomes enabled
3. Drag in Hero
4. All items now become disabled

packages/core/lib/use-puck.ts
chrisvxd: Resolving and tracking in new issue: #579

chrisvxd (changes requested):
So close! One weird bug in the comments.

apps/demo/app/custom-ui/[...puckPath]/client.tsx
Comment (xaviemirmon) [reply apps/demo/app/custom-ui/[...puckPath]/client.tsx]: fixed!

packages/core/components/Drawer/styles.module.css
```diff
@@ -2,6 +2,12 @@
   font-family: var(--puck-font-family);
 }
 
+.DrawerItem--disabled {
```
chrisvxd, line 5: Just realised, this should be applied to the `-draggable`. I'll mark this as a follow-up.

chrisvxd: Follow-up: https://github.com/measuredco/puck/issues/580

chrisvxd (approved):
🎉

PR #578, opened 2024-08-22
feat: adding resolvePermissions support

xaviemirmon:


packages/core/types/Config.tsx
```diff
@@ -63,6 +63,13 @@ export type ComponentConfig<
         props?: Partial<ComponentProps>;
         readOnly?: Partial<Record<keyof ComponentProps, boolean>>;
       };
+  resolvePermissions?: (
+    data: DataShape,
+    params: {
+      lastPermissions: Partial<Permissions> | undefined;
+      initialPermissions: Partial<Permissions>;
```
chrisvxd, line 70: I think we're missing a couple of things off the params that would keep this in sync with the other resolvers

* `changed` - An object describing which props have changed on this component since the last time this function was called. The [`get-changed`](https://github.com/measuredco/puck/blob/main/packages/core/lib/get-changed.ts) lib has your back here.
* `appState` - current app state

packages/core/lib/get-permissions.ts
```diff
@@ -20,6 +27,32 @@ export const getPermissions = ({
     config,
     globalPermissions,
   });
+  const changed =
```
chrisvxd, line 30: See below comment about additional params the resolvePermissions API, but we could use the `get-changed` lib here.

apps/demo/app/custom-ui/[...puckPath]/client.tsx
```diff
@@ -301,23 +306,106 @@ const CustomActionBar = ({ children, label }) => {
...
+  return (
+    <Section
+      padding={padding}
+      className={getClassName({
+        left: align === "left",
+        center: align === "center",
+        hasImageBackground: image?.mode === "background",
+      })}
+    >
+      {isExploding && <ConfettiExplosion />}
```
Comment (chrisvxd) [inline apps/demo/app/custom-ui/[...puckPath]/client.tsx:332]: 🤣 as much as I love this, I think we should introduce a simpler `resolvePermissions` implementation. The easiest thing is probably to set one of the default permissions to false on another component (I vote disabled duplication on <Stats> in the custom-ui demo).

chrisvxd (changes requested):
Works great! Some feedback for ya

PR #581, opened 2024-08-28
feat: adding resolve permissions

xaviemirmon:
Linked to: #461

### DIFF
--- a/apps/demo/app/custom-ui/[...puckPath]/client.tsx
+++ b/apps/demo/app/custom-ui/[...puckPath]/client.tsx
@@ -318,6 +318,7 @@ const CustomDrawer = () => {
           const canInsert = getPermissions({
             type: componentKey,
           }).insert;
+
           return (
             <Drawer.Item
               key={componentKey}
@@ -360,6 +361,13 @@ export function Client({ path, isEdit }: { path: string; isEdit: boolean }) {
           delete: false,
           duplicate: false,
           insert: false,
+        },
+        resolvePermissions: (data) => {
+          if (data.props.title.includes("Puck")) {
+            return { duplicate: true };
+          } else {
+            return { duplicate: false };
+          }
         },
       },
     },
--- a/packages/core/lib/get-permissions.ts
+++ b/packages/core/lib/get-permissions.ts
@@ -1,27 +1,72 @@
 import {
+  AppState,
   ComponentData,
   Config,
   DefaultComponentProps,
   Permissions,
 } from "../types/Config";
+import { getChanged } from "./get-changed";
+
+import { resolvePermissions } from "./resolve-permissions";
+
+const cache: {
+  lastPermissions: Record<string, Partial<Permissions>>;
+  lastSelected: ComponentData | {} | undefined;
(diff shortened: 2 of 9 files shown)


apps/demo/app/custom-ui/[...puckPath]/client.tsx
```diff
@@ -361,6 +362,13 @@ export function Client({ path, isEdit }: { path: string; isEdit: boolean }) {
           duplicate: false,
           insert: false,
         },
+        resolvePermissions: (data) => {
+          if (data.props.title.includes("Puck")) {
+            return { duplicate: true };
+          } else {
+            return { duplicate: false };
+          }
+        },
```
Comment (chrisvxd) [inline apps/demo/app/custom-ui/[...puckPath]/client.tsx:371]: Not a blocking comment, but I'll likely tweak these examples before release as I'd like them in the main demo config with a more real-world use-case

packages/core/lib/get-permissions.ts
```diff
@@ -1,28 +1,73 @@
...
   let componentPermissions = getInitialPermissions({
     componentType,
     config,
     globalPermissions,
   });
 
+  const changed = getChanged(selectedItem, cache.lastSelected);
+
+  if (Object.values(changed).some((el) => el === true)) {
+    const resolvedPermissions = resolvePermissions({
```
chrisvxd, line 42: I think this implementation works for now as it's synchronous

I know we talked a lot about synchronous vs asynchronous here and landed in synchronous to keep it more simple, but when implementing the async resolver, we will likely need to consider when this is called to avoid triggering `resolvePermissions` multiple times in parallel.

I'll open a separate ticket for async resolvers, and will aim to look at it later this week.

packages/core/types/Config.tsx
```diff
@@ -63,6 +63,15 @@ export type ComponentConfig<
         props?: Partial<ComponentProps>;
         readOnly?: Partial<Record<keyof ComponentProps, boolean>>;
       };
+  resolvePermissions?: (
+    data: DataShape,
+    params: {
+      changed: Record<string, boolean>;
```
chrisvxd, line 69:
```suggestion
      changed: Partial<Record<keyof ComponentProps, boolean>>;
```

packages/core/lib/get-permissions.ts
chrisvxd: Tracking: #582

chrisvxd (changes requested):
Awesome, works well! One minor tweak.

packages/core/types/Config.tsx
xaviemirmon: fixed!

chrisvxd (approved):
This is excellent! I'm going to make a slight tweak to the API to further align it with `resolveFields`, but will do that in post as keen to get this merged!

```tsx
  resolvePermissions?: (
     data: DataShape,
     params: {
       changed: Partial<Record<keyof ComponentProps, boolean>>;
       permissions: Partial<Permissions>;
       lastPermissions: Partial<Permissions>;
       lastData: DataShape;
       appState: AppState;
     }
   ) => Partial<Permissions>;
   ```

PR #593, opened 2024-09-01
feat: add syncHostStyles API to disable iframe style sync

chrisvxd:
Follow-on for #409 

CC @nicolas-angelo

### DIFF
--- a/apps/docs/pages/docs/api-reference/components/puck.mdx
+++ b/apps/docs/pages/docs/api-reference/components/puck.mdx
@@ -173,15 +173,22 @@ export function Editor() {
 
 #### iframe params
 
-| Param                 | Example          | Type    | Status |
-| --------------------- | ---------------- | ------- | ------ |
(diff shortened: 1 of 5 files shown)


apps/docs/pages/docs/api-reference/components/puck.mdx
```diff
@@ -173,16 +173,23 @@ export function Editor() {
...
 Render the Puck preview within iframe. Defaults to `true`.
 
 Disabling iframes will also disable [viewports](#viewports).
 
+##### `syncHostStyles`
+
+Sync the host styles to the iframe. Defaults to `true`.
+
+If this isn't working, you may need to inject styles manually using the iframe [override API](/docs/api-reference/overrides/iframe), or make use of a [plugin](/docs/extending-puck/plugins).
+
```
chrisvxd, line 192: This API won't work until we can isolate the Puck styles from the required styles. Setting `syncHostStyles` to false will prevent _any_ Preview styles from loading, making the preview unusable.

PR #598, opened 2024-09-04
Implement new drag-and-drop engine

chrisvxd:
Implement a new drag-and-drop engine using the [experimental dnd-kit release](https://next.dndkit.com). This is a backwards compatible upgrade, focused on two new pieces of functionality:

1. Dragging between nested areas (closes #123)
2. Dragging across advanced CSS layouts (inline, grid, flexbox) via elimination of edit wrappers (#455, #520)

https://github.com/user-attachments/assets/9f73c611-213f-4cd9-ab93-4cb90c6d2d01

This implementation includes additional periphery features:

* reduce DropZone to height of items unless empty
* control empty DropZone height with minEmptyHeight prop
* add DropZone collisionAxis API for forcing collision direction
* support inline Drawers, deprecating unnecessary props

Closes #556, closes #481, closes #123, closes #455, closes #520, closes #718

### Quick start

* Demo: https://puck-demo-git-dndkit-measured.vercel.app/edit
* Updated docs: https://puck-docs-git-dndkit-measured.vercel.app/docs/integrating-puck/multi-column-layouts
* Canary release: `0.17.0-canary.2d82115`

### Testing guide

* Report issues in the PR comments
* Check list of known issues before reporting
* React strict mode currently causes the drag-and-drop to be unresponsive - if you used create-puck-app, you likely have this enabled in your next.config.js and will need to disable it.
* Safari still has a bunch of issues
* Mobile/touch still needs implementing
* When using canary builds, you must explicitly reference the specific version in your package.json, i.e. `"@measured/puck": "0.17.0-canary.XXXXXX"`

### What to test

1. The new [multi-column layout behaviour](https://puck-docs-git-dndkit-measured.vercel.app/docs/integrating-puck/multi-column-layout)
2. Dragging between DropZones
3. General Puck functionality
4. Code reviews welcome

### Known issues

* [x] Latest canary triggers `Invalid instance type` error on `DragDropRegistry.register`
* [x] DragDropContext re-renders on pointer move
* [x] Overlay appearing inconsistently on hover
* [x] Doesn't work with React strict mode
  * [x] Local fixes: 1cc7e896d6059e139425eb878929f7fa0c75b71f
  * [x] Upstream fix: https://github.com/clauderic/dnd-kit/pull/1554
  * [x] Strict mode enabled in demo
* [x] Overlay not accounting for scroll when iframe disabled
* [x] Ghosting of item on drop due to rendering before position synced.
* [x] Zoom changes don't refresh overlays
[...]

PR #603, opened 2024-09-07
fix: always record history on component insert

chrisvxd:
This will result in duplicate history entries if the item has a resolveData method. This is difficult to prevent without a deferredDispatch method, similar to that being explored in #598.


mkilpatrick (approved):

mkilpatrick (commented):
Tests failed

Issue #528, opened 2024-07-11
FieldLabel bugged


IMythx:
hello

i just found that if u put a AutoField component as a child of FieldLabel (which its ele prop default value is "label") then clicking anywhere on an array item would trigger a remove action for one of the array items

it took me much time to figure out that ele prop must be "div" otherewise this bug would happen

chrisvxd:
Hi @IMythx. I don't think this is a bug, as it's HTML functionality, but I think we should consider setting the default for `FieldLabel` to `div` to prevent this kind of behaviour.

I'll update the title and mark as a feature.

IMythx:
> Hi @IMythx. I don't think this is a bug, as it's HTML functionality, but I think we should consider setting the default for `FieldLabel` to `div` to prevent this kind of behaviour.
> 
> I'll update the title and mark as a feature.

hmm idk but if u set it to div then if the child AutoField is of type text or number then clicking on the label wouldnt cuz focus on the input field so maybe add a required prop to the FieldLabel to force the user to provide the input type on which u can determine if the ele should be div or label

but even tho the default value is label still why clicking on the label would trigger a remove action??

IMythx:
> Hi @IMythx. I don't think this is a bug, as it's HTML functionality, but I think we should consider setting the default for `FieldLabel` to `div` to prevent this kind of behaviour.
> 
> I'll update the title and mark as a feature.

ok but still why focusing the array item (via clicking on the label) would trigger a remove action one of the array items(usually the first one)

IMythx:
> Hi @IMythx. I don't think this is a bug, as it's HTML functionality, but I think we should consider setting the default for `FieldLabel` to `div` to prevent this kind of behaviour.
> 
> I'll update the title and mark as a feature.

ok but still why focusing on the array item (via clicking on the label) would trigger a remove action t o one of the array items(usually the first one)

FedericoBonel:
Closing this as a duplicate of #1550, since this seems to be the one that triggers at the moment, and I can't reproduce the deletion.

PR #627, opened 2024-09-25
feat: change default  FieldLabel el from `label` to `div` (breaking change)

IMythx:
fixs [#528](https://github.com/measuredco/puck/issues/528)

### DIFF
--- a/packages/core/components/AutoField/index.tsx
+++ b/packages/core/components/AutoField/index.tsx
@@ -72,7 +72,7 @@ export const FieldLabelInternal = ({
   children,
   icon,
   label,
-  el = "div",
+  el = "label",
   readOnly,
 }: FieldLabelPropsInternal) => {
   const { overrides } = useAppContext();
--- a/packages/core/components/AutoField/index.tsx
+++ b/packages/core/components/AutoField/index.tsx
@@ -31,7 +31,7 @@ export const FieldLabel = ({
   children,
   icon,
   label,
-  el = "label",
+  el = "div",
   readOnly,
   className,
 }: {
@@ -72,7 +72,7 @@ export const FieldLabelInternal = ({
   children,
   icon,
   label,
-  el = "label",
+  el = "div",
   readOnly,
 }: FieldLabelPropsInternal) => {
   const { overrides } = useAppContext();


packages/core/components/AutoField/index.tsx
```diff
@@ -72,7 +72,7 @@ export const FieldLabelInternal = ({
   children,
   icon,
   label,
-  el = "label",
+  el = "div",
```
chrisvxd, line 75: Changing the default is a breaking change. Is there an array-only fix for this?

chrisvxd (changes requested):

packages/core/components/AutoField/index.tsx
IMythx: @chrisvxd for the docs bug yeah there is...but if anyone trying to do custom interface and used `<FieldLabel></FieldLabel>` with `<AutoField><AutoField/>` of type `Array` he must path the `el` prop as `"div"` otherwise he will face the same issue in his app thats why i changed the default value

IMythx: @chrisvxd for the docs bug yeah there is...but if anyone trying to do custom interface and used `<FieldLabel></FieldLabel>` with `<AutoField><AutoField/>` of type `Array` he must pass the `el` prop as `"div"` otherwise he will face the same issue in his app thats why i changed the default value

chrisvxd: I think we should fix the docs example, but not change the default behaviour.

`label` is the correct default, because it introduces correct form semantics.

The correct solution is to update the `array` implementation to include this on the array div (L173). This means the `array` field can be used within the label without causing issues.

```
                onClick={(e) => {
                  e.preventDefault();
                }}
```

IMythx: @chrisvxd tbh i didnt know the differnce between `FieldLable` and `FieldLabelInternal` but i just reverted it

IMythx: @chrisvxd tbh i didnt know the differnce between FieldLable and FieldLabelInternal but i just reverted it and only kept the `FieldLabel` changes

IMythx: @chrisvxd tbh i didnt know the differnce between `FieldLabel` and `FieldLabelInternal` but i just reverted it and only kept the `FieldLabel` changes

chrisvxd (approved):
Looks good! Thanks @IMythx!

However, this will be a feature as it's an API change, so I'm not going to merge it until we're closer to v0.17.0.

chrisvxd:
Closing in favour of a solution based on #686

PR #650, opened 2024-10-08
fix:defaultActionBar actions left border in RTL mode

IMythx:
puck_rtl_bug

 fixs little visual bug where the actions's left border of the default action bar would look weird in RTL mode
 
 ### steps to reproduce
 set the `direction` of the iframe's html element (or the root html if you disabled iframe) to `rtl` and select any of the components in the preview


chrisvxd (approved):
🎉

Issue #640, opened 2024-10-04
Add iframe ref to the puck context


IMythx:
Return iframe ref from the usePuck hook:
```tsx
const { iframeRef } = usePuck()
```
so we can do something like:
```tsx
iframeRef.contentWindow.matchMedia(...).matches
```

PR #652, opened 2024-10-11
feat(puck context):added iframe ref to the puck context

IMythx:
added iframe ref to the puck context for the sake of accessing the `iframe.contentWindow`, `iframe.contentDocument` and listening to events...etc

### potential use cases
1-
```tsx 
iframeRef?.current?.contentWindow?.matchMedia(...)?.matches
```
to give us the ability to create useMediaQuery custom hook inside the iframe which is differnet than doing it on the global window object (personally faced that case)

2-
```tsx
iframeRef?.current?.addEventListener('load', (e)=>{
//update some elements or do something
})
```

and of course many other potential cases

### DIFF
--- a/packages/core/components/Puck/index.tsx
+++ b/packages/core/components/Puck/index.tsx
@@ -105,7 +105,7 @@ export function Puck<
   headerTitle?: string;
   headerPath?: string;
   viewports?: Viewports;
-  iframe?: IframeConfig;
+  iframe?: Omit<IframeConfig, "ref">;
   dnd?: {
     disableAutoScroll?: boolean;
   };


IMythx:
Closes [#640](https://github.com/measuredco/puck/issues/640)

chrisvxd (commented):
Hmm, not sure if we actually need this - the Puck iframe does have a unique ID, so this could easily done by using `#preview-frame` selector.

IMythx:
> Hmm, not sure if we actually need this - the Puck iframe does have a unique ID, so this could easily done by using `#preview-frame` selector.

oh i just noticed that unique id!
but what if it got changed in the future?? i mean there is no explicit indicator that someone is using this id

chrisvxd:
Yeah, that's a valid concern, and why I haven't closed this.

I think this PR requires a bit more consideration because it adds 2 new APIs (`iframe` ref in `usePuck`, and `iframe.ref` prop to Puck component). Hopefully the selector should help for the time-being.

chrisvxd:
Closing this for now due to conflicts. Will keep issue #640.

Issue #565, opened 2024-08-15
Opportunity to interact with content inside of <Preview />


SicParv1sMagna:
It would be cool if you add an opportunity to interact with components , which were added on the page. Now, when I clicking on components inside of Preview, it starts dragging. It will be useful, if I add some components like carousel and want to interact with its filling.

PR #674, opened 2024-11-01
Functionality for switching between editing and interactive mode

SicParv1sMagna:
Recently I was trying to add PR for switching between editing and interactive mode [PR](https://github.com/measuredco/puck/pull/599). So, after a small conversation with Chris Villa in discord thread, he said that you want to add toggle in header for switching modes, I have added new functionality in context, which allows user to interact with components inside preview in "out of box" editor and in custom editor.

Closes issue https://github.com/measuredco/puck/issues/565


packages/core/components/Puck/index.tsx
```diff
@@ -195,19 +195,19 @@ export function Puck<
...
+                [categoryName]: {
+                  title: category.title,
+                  components: category.components,
+                  expanded: category.defaultExpanded,
+                  visible: category.visible,
+                },
+              };
+            },
+            {}
+          )
```
chrisvxd, line 210: I think your local prettier is conflicting with the shared prettier. Can you run `npm run format` and commit the results?

packages/core/components/DraggableComponent/index.tsx
```diff
@@ -177,8 +177,8 @@ export const DraggableComponent = ({
             </div>
           )}
 
-          <div className={getClassName("overlay")} />
-          <div className={getClassName("contents")}>{children}</div>
+          <div className={getClassName("overlay")} style={{ ...(isInteractive && { cursor: 'default', background: 'transparent' }) }} />
+          <div className={getClassName("contents")} style={{ ...(isInteractive && { pointerEvents: 'auto' }) }}>{children}</div>
```
chrisvxd, line 181: These would probably be better as a CSS class, passed to the `getClassName` function [here](https://github.com/measuredco/puck/pull/674/files#diff-d625e3b274e5aa193679313129123d868084442be8eaa76119437f5ceb0f6ebaR124).

packages/core/components/Puck/context.tsx
```diff
@@ -72,6 +72,8 @@ export type AppContext<
   plugins: Plugin[];
   overrides: Partial<Overrides>;
   history: Partial<PuckHistory>;
+  isInteractive: boolean;
```
chrisvxd, line 75: Any ideas for how you could add this to the `custom-ui` demo? I like to use features in demos for easier reviewing, and so we can be sure things build properly.

chrisvxd (changes requested):
Nice one, I really like the idea of keeping this in state! I've left some comments.

My only concern I have is that this is going to clash with #556, which rebuilds DraggableComponent.

If you address my comments, we can figure out the conflicts before the next feature release, which will probably include #556.

packages/core/components/Puck/context.tsx
SicParv1sMagna: Sure, I will try to add this into demo tomorrow

SicParv1sMagna: Is there any information about future release? How soon can I start using this changes?

packages/core/components/DraggableComponent/index.tsx
SicParv1sMagna: Done

packages/core/components/Puck/index.tsx
SicParv1sMagna: done

packages/core/components/DraggableComponent/index.tsx
SicParv1sMagna: done

packages/core/components/Puck/context.tsx
SicParv1sMagna: done

chrisvxd (changes requested):
I've been doing some thinking and this will actually be better on the AppState rather than a separate status. This saves us polluting the usePuck hook with every piece of appState, and introducing a custom dispatcher (`setIsInteractive`)

I also think that `isInteractive` is probably not quite the right name. I would prefer to introduce a new parameter to AppState, either `mode` or `previewMode: "interactive" | "edit"`. Does that make sense?

Having said that, I know you're keen to get this done, so I'm happy to do a canary containing this existing implementation if it would help you out.

SicParv1sMagna:
> I've been doing some thinking and this will actually be better on the AppState rather than a separate status. This saves us polluting the usePuck hook with every piece of appState, and introducing a custom dispatcher (`setIsInteractive`)
> 
> I also think that `isInteractive` is probably not quite the right name. I would prefer to introduce a new parameter to AppState, either `mode` or `previewMode: "interactive" | "edit"`. Does that make sense?
> 
> Having said that, I know you're keen to get this done, so I'm happy to do a canary containing this existing implementation if it would help you out.

I think that you're right, but as I understand, usePuck is allowed either on "out of box" solution (default <Puck/>), either on custom layout side, so I would like to use it from usePuck hook, cus I'am using custom layout.
About naming, got you, I will try to fix this as soon as possible, probably, on the beginning of next week or weekend

chrisvxd:
@SicParv1sMagna the app state is exposed via usePuck, and can be controlled via the dispatcher: https://puckeditor.com/docs/api-reference/functions/use-puck#appstate

I'm not convinced on the naming yet. For now, let's just move `isInteractive` to the app state and worry about the rename later (I can do a follow-up).

SicParv1sMagna:
@chrisvxd Hello, sorry for fixing this pr so long. Removed isInteractive from usePuck hook and renamed it to mode

SicParv1sMagna:
@chrisvxd is my merge request still actual?

chrisvxd (changes requested):
Hey @SicParv1sMagna! Sorry for the delay.

Just reviewing now and this doesn't quite behave as expected. When I switch to interactive mode, the items are still draggable and action bar overlay shows. I think it might be better to modify DropZone to avoid rendering <DraggableComponent> altogether when in `interactive` mode.

SicParv1sMagna:
@chrisvxd Missed that moment, fixed it, check it please)

SicParv1sMagna:
@chrisvxd Can you please check new version

chrisvxd (changes requested):
Hi @SicParv1sMagna - I've addressed merge conflicts, but you still need to update your tests for this to pass CI. Please run `cd packages/core && npm run test -u` and commit the updated snapshot change.

SicParv1sMagna:
> Hi @SicParv1sMagna - I've addressed merge conflicts, but you still need to update your tests for this to pass CI. Please run `cd packages/core && npm run test -u` and commit the updated snapshot change.

Give me few minutes

SicParv1sMagna:
> `cd packages/core && npm run test -u`

Done

chrisvxd:
@SicParv1sMagna this is failing again, and now conflicts with the changes merged from #598. I'm going to close this, take your commits, and spin up another PR to get the last of this through. You'll still get authorship credit, and I wanted to say thank you for your work on this 🙏 

Will ping you in the new PR.

PR #678, opened 2024-11-06
fix: broken link in <Render/> component docs

novacoole:
### Steps to reproduce
1. Go to [https://puckeditor.com/docs/api-reference/components/render#required-props](https://puckeditor.com/docs/api-reference/components/render#required-props)
2. Click "Config docs" link


chrisvxd (approved):
Nice catch, thanks @novacoole!

PR #682, opened 2024-11-06
feat: allow react elements in external field mapRow

### LINKED ISSUE #506: Feature request: Allow mapRow to return more than just string and number types
How about letting the `mapRow` function on `external` fields return more types than just `string` and `number`?

For example, if it also allowed `React Node` (or`React Element`?) you could do things like show thumbnail images for the items listed. This would make the `external` selection feature far more powerful.

Alternatively, maybe allowing an override for the list renderer would allow even more flexibility?

camhammel:
## Description
- Adds support for `mapRow` items of type `ReactElement` (previously just `string | number`)
- Resolves #506 

## Screenshots
image

```
       mapRow: (item) => ({
        img: (
          <img src="https://placehold.co/600x400" height="40" width="60" />
        ),
        title: item.title,
        description: item.description,
      }),
```

### DIFF
--- a/apps/demo/config/blocks/Hero/index.tsx
+++ b/apps/demo/config/blocks/Hero/index.tsx
@@ -77,7 +77,10 @@ export const Hero: ComponentConfig<HeroProps> = {
             }
           });
       },
-      mapRow: (item) => ({ title: item.title, description: item.description }),
+      mapRow: (item) => ({
+        title: item.title,
+        description: <span>{item.description}</span>,
+      }),
       mapProp: (result) => {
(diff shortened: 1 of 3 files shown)


chrisvxd (changes requested):
Thanks @camhammel - this looks good. I try and include examples in the demo, which helps to validate things work and build correctly.

Could you add a superficial example [here](https://github.com/measuredco/puck/blob/main/apps/demo/config/blocks/Hero/index.tsx#L80)? Just wrap the description in a `<span>` or something.

PR #684, opened 2024-11-07
feat: add renderFooter option in external field

camhammel:
## Description
Adds a new optional property to external fields, `renderFooter`. This property accepts a function resolving to a `ReactElement` in place of the default footer, with a prop provided for `items` representing the `mappedData`. 

## Screenshots

CleanShot 2024-11-07 at 12 31 32

## Motivation
We're using the `external` field to allow the user to select an image from a set of images provided by our external service. We'd like to be able to display a "Manage Images" link in the footer to provide the user with a clear path to add/remove any of the image options displayed in the modal.

### DIFF
--- a/apps/docs/pages/docs/api-reference/fields/external.mdx
+++ b/apps/docs/pages/docs/api-reference/fields/external.mdx
@@ -63,7 +63,7 @@ const config = {
 | [`mapRow()`](#maprowitem)                 | `mapRow: async ({ title }) => title`         | Function   | -        |
 | [`placeholder`](#placeholder)             | `placeholder: "Select content"`              | String     | -        |
 | [`showSearch`](#showsearch)               | `showSearch: true`                           | Boolean    | -        |
-| [`renderFooter()`](#renderfooterdata)     | `renderFooter: (props) => <p>Hello</p>`      | Function   | -        |
+| [`renderFooter()`](#renderfooterprops)    | `renderFooter: (props) => <p>Hello</p>`      | Function   | -        |
 
 ## Required params
 
--- a/apps/docs/pages/docs/api-reference/fields/external.mdx
+++ b/apps/docs/pages/docs/api-reference/fields/external.mdx
@@ -63,7 +63,7 @@ const config = {
 | [`mapRow()`](#maprowitem)                 | `mapRow: async ({ title }) => title`         | Function   | -        |
 | [`placeholder`](#placeholder)             | `placeholder: "Select content"`              | String     | -        |
 | [`showSearch`](#showsearch)               | `showSearch: true`                           | Boolean    | -        |
-| [`renderFooter()`](#renderfooterdata)     | `renderFooter: (data) => <p>Hello</p>`       | Function   | -        |
+| [`renderFooter()`](#renderfooterdata)     | `renderFooter: (props) => <p>Hello</p>`      | Function   | -        |
 
 ## Required params
 
@@ -612,7 +612,7 @@ const config = {
 }}
 />
 
-### `renderFooter(data)`
+### `renderFooter(props)`
 
 Customize what will be displayed in the footer of the modal.
 
@@ -629,7 +629,7 @@ const config = {
               { title: "Goodbye, world", description: "Lorem ipsum 2" },
             ];
           },
-          renderFooter: (data) => <span>Hello, footer</span>,
+          renderFooter: ({ items }) => <span>{items.length} results</span>,
         },
       },
       render: ({ data }) => {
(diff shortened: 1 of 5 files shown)


packages/core/types/Fields.ts
```diff
@@ -94,6 +94,7 @@ export type ExternalField<
   mapRow?: (value: any) => Record<string, string | number>;
   getItemSummary?: (item: Props, index?: number) => string;
   showSearch?: boolean;
+  renderFooter?: (data: any[]) => ReactElement;
```
chrisvxd, line 97: The props are probably better as an object, so this can be treated as a React component. Since we then also need to name it, let's use `items` instead of `data`.

```suggestion
  renderFooter?: (props: { items: any[] }) => ReactElement;
```

packages/core/components/ExternalInput/index.tsx
```diff
@@ -265,9 +265,14 @@ export const ExternalInput = ({
               </div>
             </div>
           </div>
-
-          <div className={getClassNameModal("footer")}>
-            {mappedData.length} result{mappedData.length === 1 ? "" : "s"}
+          <div className={getClassNameModal("footerContainer")}>
+            {field.renderFooter ? (
+              field.renderFooter(mappedData)
```
chrisvxd, line 270: Although this will work, it will stop React from being able to optimise correctly, creating a new component signature on each render and forcing a re-render.

I'd recommend moving this to the top of the component and wrapping it in a memo, then using JSX to render it.

Something like this (untested):

```tsx
const Footer = useMemo(() = (data) => (
  field.renderFooter ? (
    field.renderFooter(data)
  ) : (
    <span className={getClassNameModal("footer")}>
      {data.length} result{data === 1 ? "" : "s"}
    </span>
  )
), [field.renderFooter]);
```

```tsx
<Footer data={mappedData} />
```

chrisvxd (changes requested):
Another excellent feature addition, thank you!

Left some comments.

apps/docs/pages/docs/api-reference/fields/external.mdx
```diff
@@ -60,9 +60,10 @@ const config = {
 | [`initialFilters`](#initialfilters)       | `{ "rating": 1 }`                            | Object     | -        |
 | [`initialQuery`](#initialquery)           | `initialQuery: "Hello, world"`               | String     | -        |
 | [`mapProp()`](#mappropitem)               | `mapProp: async ({ title }) => title`        | Function   | -        |
-| [`mapRow()`](#mappropitem)                | `mapRow: async ({ title }) => title`         | Function   | -        |
+| [`mapRow()`](#maprowitem)                 | `mapRow: async ({ title }) => title`         | Function   | -        |
 | [`placeholder`](#placeholder)             | `placeholder: "Select content"`              | String     | -        |
 | [`showSearch`](#showsearch)               | `showSearch: true`                           | Boolean    | -        |
+| [`renderFooter()`](#renderfooterprops)    | `renderFooter: (props) => <p>Hello</p>`      | Function   | -        |
```
chrisvxd, line 66: One minor thing! Can you make this alphabetical? i.e. above `showSearch`?

apps/docs/pages/docs/api-reference/fields/external.mdx
```diff
@@ -611,4 +612,54 @@ const config = {
 }}
 />
 
+### `renderFooter(props)`
```
chrisvxd, line 615: As with my other comment, can you move this above the `### showSearch` heading to make it alphabetical?

chrisvxd (changes requested):
Thanks for addressing the comments! Almost perfect now, just spotted one final nit

apps/docs/pages/docs/api-reference/fields/external.mdx
```diff
@@ -611,4 +612,54 @@ const config = {
...
+      fields: {
+        data: {
+          type: "external",
+          fetchList: async () => {
+            return [
+              { title: "Hello, world", description: "Lorem ipsum 1" },
+              { title: "Goodbye, world", description: "Lorem ipsum 2" },
+            ];
+          },
+          renderFooter: ({ items }) => <span>{items.length} results</span>,
```
chrisvxd, line 632: Just realised it would probably help to demo something that looks different to the default. Not entirely sure what, but something like `<b>Custom footer with {items.length} results</b>`, or better

chrisvxd (approved):
🎉  Think I'm actually gonna go ahead and merge this and your other one, so a canary should be released shortly.

Issue #694, opened 2024-11-08
Strip leading zeros in Number Field


benlife5:
When you delete the value in NumberField, it resets to 0. However, once you type in another number, it keeps the leading 0. Ideally, this would be stripped as the user is typing.

https://github.com/user-attachments/assets/08c77494-0278-4c1f-b2be-a65c7f5dbefa

https://stackblitz.com/edit/vitejs-vite-ncvhwx?file=src%2FEditor.tsx

chrisvxd:
Closed by #695

PR #695, opened 2024-11-08
fix: remove leading zeros in NumberField

benlife5:
Addresses https://github.com/measuredco/puck/issues/694


benlife5:
https://github.com/user-attachments/assets/a2d7296e-991d-4dd9-8005-e6d28bc18770

PR #696, opened 2024-11-12
fix(DropZone): Ensure consistent application of className and style in both edit and render modes

haidv1992:
Problem Description and Purpose of Update
In the current implementation of the DropZone component, there is an inconsistency in how className styling is applied between the edit mode (editMode: true) and render mode (editMode: false). Specifically, when using tailwindcss classes in the className prop, they apply correctly in render mode but not in edit mode due to additional div wrappers within DropZoneEdit. This structural difference leads to layout and styling discrepancies between the two modes.

Purpose of This Update
This update addresses the inconsistent application of className between DropZoneEdit and DropZoneRender by ensuring that className and style props are passed directly to the main container in both modes. By doing so, this change allows consistent usage of CSS classes, including tailwindcss styles, for both editing and rendering without altering the core structure of the component.

### DIFF
--- a/apps/docs/pages/docs/api-reference/components/drop-zone.mdx
+++ b/apps/docs/pages/docs/api-reference/components/drop-zone.mdx
@@ -1,5 +1,5 @@
----
-title: <DropZone>
+Để bổ sung `className` vào tài liệu của `DropZone` cho phép áp dụng các lớp CSS tùy chỉnh vào thành phần, bạn có thể cập nhật phần tài liệu như sau:
+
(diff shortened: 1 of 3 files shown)


IMythx:
wait why am i mentioned here ? xdd

chrisvxd (commented):
Hey @haidv1992 and sorry for the slow review. Thanks so much for your contribution. This is actually covered by #598, which also refactors the DropZone considerably so I'm going to close this out in favour of that.

PR #707, opened 2024-11-17
Fix wording of delete permission documentation

antonva:
Noticed that the word "duplication" had duplicated itself.


chrisvxd (approved):
😏

PR #709, opened 2024-11-19
fix(resolvePermissions): dynamic insert permission

IMythx:
_**actually not sure if this is the right way to fix it but could'nt figure out another way**_


chrisvxd (commented):
Thanks @IMythx! As you noted, this isn't the correct fix. I think there's some confusion about how `resolvePermissions` behaves. `resolvePermissions` is triggered once per component instance, but `insert` is global for that component.

The correct way to dynamically set `insert` is to alter the component config outside of Puck and pass it in.

PR #711, opened 2024-11-21
Support passing rootProps to resolveData

haidv1992:
### Summary
This PR adds support for passing `rootProps` from `newData.root.props` into `resolveData` to enable dynamic configuration based on root-level properties.

### Changes
- Modified `useResolvedData` to extract `rootProps` from `newData.root.props`.
- Updated `resolveComponentData` to accept and pass `rootProps` into `resolveData`.
- Enhanced `Hero` component to demonstrate usage of `rootProps`.
- Updated type definitions in `Config.tsx` to include `rootProps`.

### Testing
- Verified that `rootProps` are correctly passed to components.
- Ensured backward compatibility with existing workflows.

### DIFF
--- a/apps/demo/config/blocks/Hero/index.tsx
+++ b/apps/demo/config/blocks/Hero/index.tsx
@@ -173,7 +173,7 @@ export const Hero: ComponentConfig<HeroProps> = {
    *
    * For example, requesting a third-party API for the latest content.
    */
-  resolveData: async ({ props }, { changed }) => {
+  resolveData: async ({ props }, { changed,rootProps }) => {
     if (!props.quote)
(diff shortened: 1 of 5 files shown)


chrisvxd:
Apologies for delay in reviewing your PRs @haidv1992 - I've clocked them but have some catching up to do. Watch this space!

chrisvxd (changes requested):
Thanks again @haidv1992!

I'm not sure this is the right approach. I think this would be better handled by adding the `appState` option to `resolveData`, to match the other resolvers, like [`resolveFields`](https://puckeditor.com/docs/api-reference/configuration/component-config#paramsappstate).

chrisvxd:
Stale

PR #714, opened 2024-11-29
Enhance ExternalInput: Pass filters to mapProp for custom data mapping

haidv1992:
Enhance ExternalInput: Pass filters to mapProp for custom data mapping

- Updated `mapProp` function to accept filters as an argument.
- Modified `ExternalInput` component to pass current filters when selecting data from the table.
- Enabled dynamic data mapping based on filters for better flexibility.

### DIFF
--- a/packages/core/components/ExternalInput/index.tsx
+++ b/packages/core/components/ExternalInput/index.tsx
@@ -31,7 +31,7 @@ export const ExternalInput = ({
   readOnly?: boolean;
 }) => {
   const {
-    mapProp = (val: any) => val,
+    mapProp = (val: any,filters: object) => val,
     mapRow = (val: any) => val,
     filterFields,
   } = field || {};
@@ -244,7 +244,7 @@ export const ExternalInput = ({
                         style={{ whiteSpace: "nowrap" }}
                         className={getClassNameModal("tr")}
                         onClick={() => {
-                          onChange(mapProp(data[i]));
+                          onChange(mapProp(data[i],filters));
 
                           setOpen(false);
                         }}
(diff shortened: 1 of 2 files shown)


packages/core/components/ExternalInput/index.tsx
```diff
@@ -41,8 +41,8 @@ export const ExternalInput = ({
   const [isLoading, setIsLoading] = useState(true);
 
   const hasFilterFields = !!filterFields;
-
-  const [filters, setFilters] = useState(field.initialFilters || {});
+  // @ts-ignore
```
chrisvxd, line 44: We ideally shouldn't be introducing `@ts-ignore`, if at all possible

packages/core/components/ExternalInput/index.tsx
```diff
@@ -244,7 +244,7 @@ export const ExternalInput = ({
                         style={{ whiteSpace: "nowrap" }}
                         className={getClassNameModal("tr")}
                         onClick={() => {
-                          onChange(mapProp(data[i]));
+                          onChange(mapProp(data[i],filters));
```
chrisvxd, line 247: This needs to be formatted, otherwise it will fail linting. You can run `npm run format` to address it.

packages/core/types/Fields.ts
```diff
@@ -90,7 +90,7 @@ export type ExternalField<
     query: string;
     filters: Record<string, any>;
   }) => Promise<any[] | null>;
-  mapProp?: (value: any) => Props;
+  mapProp?: (value: any,filters:any) => Props;
```
chrisvxd, line 93: The second param would probably be better as an object, and we may as well use the correct type for `filters` -

```suggestion
  mapProp?: (value: any, options: { filters: Record<string, any> }) => Props;
```

chrisvxd (changes requested):
Thanks @haidv1992!

Filters are not really intended to be used in this way, and are designed for filtering the data, not mutating it. I worry that applying them to the mapProp field might make the naming of `filters` a little unclear.

Can you share an example use-case?

chrisvxd:
Stale

Issue #713, opened 2024-11-26
Numbers in select field are converted to string


thopedam:
Using the following configuration the value is being converted to a string

```javascript
{
  fields: {
    numberField: {
      type: "select", // "radio"
      label: "A Select-Field with numbers",
      options: [
        { value: 1, label: "1" },
        { value: 2, label: "2" },
        { value: 3, label: "3" },
        { value: 4, label: "4" },
        { value: 5, label: "5" },
        { value: 6, label: "6" },
        { value: 7, label: "7" },
        { value: 8, label: "8" },
        { value: 9, label: "9" },
        { value: 10, label: "10" },
        { value: 11, label: "11" },
        { value: 12, label: "12" },
      ],
    },
    resolveData: (data, {changed}) {
      console.log(JSON.stringify(data, null, 2));
    }
  }
}
```

This returns:

```
{
...
  "props": {
    "numberField": "6"
  }
...
}
```
This happens because internally the HTML input values are string.

Maybe something like this is helpful for SelectField and RadioField? I am maybe missing something, so this is a bug/discussion and not a MR.  

```
const isNumberField =
    typeof value === "number" ||
    field.options.filter((option) => typeof option.value === "number")?.length >
      0;
```

chrisvxd:
We already support booleans in radio/select, so I'm happy to consider this a bug and extend that implementation to numbers! Marking as ready.

chrisvxd:
Closed by #721

PR #721, opened 2024-12-12
fix: adds condition check for number values in select

jswhisperer:
https://github.com/measuredco/puck/issues/713

- Aims to fix For when select boxes that have number values  this PR adds a check and JSON.parse the value.
- (also sneakily adds a title attr for added accessibility)

An alternative approach might be to safely JSON parse all values and should that fail just return the value as is or an empty string. https://gist.github.com/jswhisperer/ec7c99dab7ca8549f40cbd1e853399ae

### DIFF
--- a/packages/core/components/AutoField/fields/SelectField/index.tsx
+++ b/packages/core/components/AutoField/fields/SelectField/index.tsx
@@ -27,18 +27,20 @@ export const SelectField = ({
     >
       <select
         id={id}
+        title={label || name}
         className={getClassName("input")}
         disabled={readOnly}
-        onChange={(e) => {
+        onChange={({ target: { value } }) => {
           if (
-            e.currentTarget.value === "true" ||
-            e.currentTarget.value === "false"
+          value === "true" ||
+          value === "false" ||
+          typeof value === "number"
           ) {
-            onChange(JSON.parse(e.currentTarget.value));
+            onChange(JSON.parse(value));
             return;
           }
 
-          onChange(e.currentTarget.value);
+          onChange(value);
         }}
         value={value}
       >
--- a/packages/core/components/AutoField/fields/SelectField/index.tsx
+++ b/packages/core/components/AutoField/fields/SelectField/index.tsx
@@ -31,18 +31,7 @@ export const SelectField = ({
         title={label || name}
         className={getClassName("input")}
         disabled={readOnly}
-        onChange={({ target: { value } }) => {
-          if (
-          value === "true" ||
-          value === "false" ||
-          typeof safeJsonParse(value) === "number"
-          ) {
-            onChange(JSON.parse(value));
-            return;
-          }
-
-          onChange(value);
-        }}
+        onChange={({ target: { value } }) => safeJsonParse(value) ? onChange(JSON.parse(value)) : onChange(value)}
         value={value}
       >
         {field.options.map((option) => (
--- a/packages/core/components/AutoField/fields/SelectField/index.tsx
+++ b/packages/core/components/AutoField/fields/SelectField/index.tsx
@@ -2,7 +2,7 @@ import getClassNameFactory from "../../../../lib/get-class-name-factory";
 import styles from "../../styles.module.css";
 import { ChevronDown } from "lucide-react";
 import { FieldPropsInternal } from "../..";
-import {safeJsonParse } from "../../../../lib/safe-json-parse"
+import { safeJsonParse } from "../../../../lib/safe-json-parse";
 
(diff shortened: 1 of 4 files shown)


jswhisperer:
Ah sorry hold up on review needs changes

packages/core/components/AutoField/fields/SelectField/index.tsx
```diff
@@ -27,18 +28,20 @@ export const SelectField = ({
     >
       <select
         id={id}
+        title={label || name}
         className={getClassName("input")}
         disabled={readOnly}
-        onChange={(e) => {
+        onChange={({ target: { value } }) => {
```
chrisvxd, line 34: I guess this is a deliberate switch from `currentTarget` to `target`? Makes sense

packages/core/components/AutoField/fields/SelectField/index.tsx
```diff
@@ -27,18 +28,20 @@ export const SelectField = ({
...
           if (
-            e.currentTarget.value === "true" ||
-            e.currentTarget.value === "false"
+          value === "true" ||
+          value === "false" ||
+          typeof safeJsonParse(value) === "number"
```
chrisvxd, line 38: This is smart. I suppose we could reuse this for the boolean.

In fact, we could consider _always_ returning JSON if it's valid 🤔 that would add support for objects, too.

packages/core/components/AutoField/fields/SelectField/index.tsx
```diff
@@ -27,18 +28,20 @@ export const SelectField = ({
     >
       <select
```
chrisvxd, line 29: We probably should extend this to `radio` field too!

chrisvxd (changes requested):
Great stuff! Some comments

packages/core/components/AutoField/fields/SelectField/index.tsx
jswhisperer: Agreed, that is my preference I would actually prefer always returning if valid and if not returning the value as is, and not doing any typeof check.

packages/core/components/AutoField/fields/SelectField/index.tsx
```diff
@@ -27,19 +28,10 @@ export const SelectField = ({
...
-            e.currentTarget.value === "true" ||
-            e.currentTarget.value === "false"
-          ) {
-            onChange(JSON.parse(e.currentTarget.value));
-            return;
-          }
-
-          onChange(e.currentTarget.value);
-        }}
+        onChange={({ target: { value } }) => safeJsonParse(value) ? onChange(JSON.parse(value)) : onChange(value)}
```
chrisvxd, line 34: Can you use the return value of `safeJsonParse` so we're not running JSON parse twice?

packages/core/lib/safe-json-parse.ts
```diff
@@ -0,0 +1,8 @@
+export const safeJsonParse = <T>(str: string) => {
+  try {
+    const jsonValue: T = JSON.parse(str);
+    return jsonValue;
+  } catch {
+    return undefined;
```
chrisvxd, line 6:
```suggestion
    return str;
```

chrisvxd (changes requested):

packages/core/components/AutoField/fields/SelectField/index.tsx
```diff
@@ -2,6 +2,7 @@ import getClassNameFactory from "../../../../lib/get-class-name-factory";
 import styles from "../../styles.module.css";
 import { ChevronDown } from "lucide-react";
 import { FieldPropsInternal } from "../..";
+import {safeJsonParse } from "../../../../lib/safe-json-parse"
```
chrisvxd, line 5: Can you make sure you're running prettier or run `npm run format` and commit the results? Else CI will fail.

jswhisperer: Ah yeah! weirdly vs-code didn't pick up the prettier config, later (one day) will look at adding a pre-commit hook to auto do this, after the real work is done, but did this manually no problemo was bothering me too

chrisvxd: Please no pre commit hooks 😅 really disrupts my workflow

chrisvxd (approved):
🎉

packages/core/components/AutoField/fields/SelectField/index.tsx
jswhisperer: ok done... or not done... I mean  hahah no problem.

PR #746, opened 2024-12-18
feat: use React 19 in next recipe

chrisvxd:
Next 15 mandates React 19 for App Router, which the recipe uses: https://nextjs.org/blog/next-15#react-19

Although we can't upgrade the entire app to React 19 yet (see #739), we can upgrade the Next app.


chrisvxd:
Closing in favour of #739

PR #748, opened 2024-12-18
fix: booleans type in select/radio fields

thopedam:
`safeJsonParse(false)` returns `false` so `const jsonValue` will be `"false"` and cannot be found in `flatOptions`.

We need to check for `undefined` and not for `false`.


chrisvxd (approved):
D'oh, good catch! Unfortunately this will miss 0.17, which just got released.

thopedam:
Just updated to 0.17.0 and was confused about my booleans not working anymore.. 😀

PR #752, opened 2024-12-22
Add react-router v7 example

matthewlynch:
This PR adds a react-router recipe for Puck


matthewlynch:
@chrisvxd deps are failing to install because we need to bump the minimum engine version for node:
> error @react-router/node@7.1.0: The engine "node" is incompatible with this module. Expected version ">=20.0.0". Got "18.20.5"

.github/workflows/ci.yml
```diff
@@ -19,7 +19,7 @@ jobs:
       - name: Setup Node.js
         uses: actions/setup-node@v3
         with:
-          node-version: 18
+          node-version: 20
```
chrisvxd, line 22: This bump is causing CI to fail for Vercel. Noted your comment on RR7 dependencies, though. Need to probe this a bit further.

recipes/react-router/app/routes/puck-splat.tsx
```diff
@@ -0,0 +1,93 @@
...
+  return (
+    <div>
+      {loaderData.isEditorRoute ? (
+        <Editor />
+      ) : (
+        <Render config={config} data={loaderData.data} />
+      )}
+    </div>
+  );
+}
```
chrisvxd, line 61: Might make sense to move this to the bottom?

recipes/react-router/app/routes/puck-splat.tsx
```diff
@@ -0,0 +1,93 @@
+import { useFetcher, useLoaderData } from "react-router";
+import type { Data } from "@measured/puck";
+import { Puck, Render } from "@measured/puck";
+
+import "@measured/puck/puck.css";
+
+import type { Route } from "./+types/puck-splat";
```
chrisvxd, line 7: What is this `+` import magic?!

recipes/react-router/app/root.tsx
```diff
@@ -0,0 +1,76 @@
...
+    rel: "preconnect",
+    href: "https://fonts.gstatic.com",
+    crossOrigin: "anonymous",
+  },
+  {
+    rel: "stylesheet",
+    href: "https://fonts.googleapis.com/css2?family=Inter:ital,opsz,wght@0,14..32,100..900;1,14..32,100..900&display=swap",
+  },
+  { rel: "stylesheet", href: stylesheet },
+];
```
chrisvxd, line 25: Hm what's the font import here for? Would be good to keep these recipes as minimal as possible.

recipes/react-router/app/puck.config.tsx
```diff
@@ -0,0 +1,23 @@
...
+        title: "Heading",
+      },
+      render: ({ title }) => (
+        <div style={{ padding: 64 }}>
+          <h1>{title}</h1>
+        </div>
+      ),
+    },
+  },
+};
```
chrisvxd, line 23: Can this file move up a directory? I know that the TS compiler might require the files to be in certain directories to run properly, so wasn't sure if this was deliberate or not.

recipes/react-router/app/lib/resolve-puck-path.server.ts
```diff
@@ -0,0 +1,13 @@
+export function resolvePuckPath(path: string = "") {
+  const url = new URL(path, "https://placeholder.com/");
```
chrisvxd, line 2: It's not clear that `placeholder.com` doesn't need to be changed by the user (in fact, I thought I would have to change it after generating).

I understand the technique here, but is there a cleaner way we could do this without requiring the placeholder string?

recipes/react-router/tailwind.config.ts
```diff
@@ -0,0 +1,22 @@
...
+          '"Apple Color Emoji"',
+          '"Segoe UI Emoji"',
+          '"Segoe UI Symbol"',
+          '"Noto Color Emoji"',
+        ],
+      },
+    },
+  },
+  plugins: [],
+} satisfies Config;
```
chrisvxd, line 22: Can we keep Tailwind out of this recipe? The existing recipes tend to be pure in this regard, so I think we should keep it consistent.

recipes/react-router/package.json
```diff
@@ -0,0 +1,34 @@
...
+    "build": "cross-env NODE_ENV=production react-router build",
+    "dev": "react-router dev",
+    "start": "cross-env NODE_ENV=production react-router-serve ./build/server/index.js",
+    "typecheck": "react-router typegen && tsc"
+  },
+  "dependencies": {
+    "@measured/puck": "^0.17.1",
+    "@react-router/node": "^7.0.2",
+    "@react-router/serve": "^7.0.2",
+    "isbot": "^5.1.17",
```
chrisvxd, line 16: Noticed this is in the standard react-router template, but is this dep necessary?

recipes/react-router/README.md
```diff
@@ -0,0 +1,39 @@
+# `react-router` recipe
+
+The `react-router` recipe showcases one of the most powerful ways to implement Puck using to provide an authoring tool for any route in your React Router app.
+
+## Demonstrates
+
+- React Router V7 (framework) implementation
```
chrisvxd, line 7:
```suggestion
- React Router v7 (framework) implementation
```

recipes/react-router/README.md
```diff
@@ -0,0 +1,39 @@
+# `react-router` recipe
+
+The `react-router` recipe showcases one of the most powerful ways to implement Puck using to provide an authoring tool for any route in your React Router app.
```
chrisvxd, line 3: Need to do an overhaul if these recipe READMEs at some point, but this is fine for now.

chrisvxd (changes requested):
Awesome thanks @matthewlynch! Left some comments 🙃

recipes/react-router/app/routes/puck-splat.tsx
matthewlynch: I was following the convention used in route files for RR/Remix but I can update this

recipes/react-router/app/routes/puck-splat.tsx
matthewlynch: This is something new in RR7. Types are auto generated based on file changes for the route. Works really well.

recipes/react-router/app/root.tsx
matthewlynch: It was the default RR template, I can remove it

recipes/react-router/package.json
matthewlynch: Defaults from the base RR template, it can be removed but we'd need to add a new server entry file because RR expects the dependency to be there.

.github/workflows/ci.yml
matthewlynch: I could replicate the issue locally, moving to v22 (LTS) works so will try on CI once I've sorted a monorepo issue.

recipes/react-router/app/lib/resolve-puck-path.server.ts
matthewlynch: I've added a comment saying it isn't important and set the origin as a default parameter value. Let me know if you want to change it to something else.

recipes/react-router/package.json
chrisvxd: Generally I'm okay with the RR defaults rather than removing them. I just wanted to probe it a bit.

recipes/react-router/app/entry.server.tsx
```diff
@@ -0,0 +1,68 @@
+import { PassThrough } from "node:stream";
```
chrisvxd, line 1: Is this the "revealed" file? I'd actually rather not include this since that's not the default behaviour

recipes/react-router/app/root.tsx
```diff
@@ -0,0 +1,72 @@
...
+  return (
+    <main className="pt-16 p-4 container mx-auto">
+      <h1>{message}</h1>
+      <p>{details}</p>
+      {stack && (
+        <pre className="w-full p-4 overflow-x-auto">
+          <code>{stack}</code>
+        </pre>
+      )}
+    </main>
```
chrisvxd, line 70: These are referring to Tailwind classes, but we've removed Tailwind

chrisvxd (changes requested):
Looking good, couple more comments.

Re Node v22, we just need to make sure we're not changing the Puck minimum node version just to support an RR7 template

matthewlynch:
@chrisvxd as discussed, bumped the repo min node version to v20.

chrisvxd (approved):
🚀

PR #834, opened 2025-01-24
fix: override fieldType incorrect type

### LINKED ISSUE #626: onChange provided to override field types has incorrect type
### Overview 
When I try to override any fieldType, I get error on the line where onChange is called. 
Error: Argument of type 'string' is not assignable to parameter of type 'NumberField'.
  Type 'string' is not assignable to type '{ type: "number"; min?: number | undefined; max?: number | undefined; }'.ts(2345)

### Example code snippet  
```tsx
        fieldTypes: {
          number: ({ onChange, field, id, name }) => (
            <div>
              <label
                className="block text-sm font-medium leading-6 text-gray-900"
                htmlFor={name}
              >
                {field.label}
              </label>
              <div className="mt-2">
                <input
                  aria-describedby="number-description"
                  className="block w-full rounded-md border-0 py-1.5 text-gray-900 shadow-sm ring-1 ring-inset ring-gray-300 placeholder:text-gray-400 focus:ring-2 focus:ring-inset focus:ring-indigo-600 sm:text-sm sm:leading-6"
                  id={id}
                  type="number"
                  // eslint-disable-next-line react/jsx-sort-props
                />
              </div>
            </div>
          ),
```

### Expectations
onChange should accept type that is the Value of the field, instead it accepts the FieldType. 

### Potential problematic source code
_Overrides.ts_
```ts
type FieldRenderFunctions = Omit<{
    [Type in Field["type"]]: React.FunctionComponent<FieldProps<Extract<Field, {
        type: Type;
    }>> & {
        children: ReactNode;
        name: string;
    }>;
}, "custom"> & {
    [key: string]: React.FunctionComponent<FieldProps<any> & {
        children: ReactNode;
        name: string;
    }>;
};
```

rv-bparise:
This fixes #626.  Although the use of `any` type here is not ideal, it does address the underlying issue of incorrect usage of the `FieldProps` type.

### DIFF
--- a/packages/core/types/API/Overrides.ts
+++ b/packages/core/types/API/Overrides.ts
@@ -57,7 +57,7 @@ export type Overrides = OverridesGeneric<{
 export type FieldRenderFunctions = Omit<
   {
     [Type in Field["type"]]: React.FunctionComponent<
-      FieldProps<Extract<Field, { type: Type }>> & {
+      FieldProps<any, Extract<Field, { type: Type }>> & {
         children: ReactNode;
         name: string;
       }


chrisvxd (approved):
I held off reviewing this as I wanted to try and use something more specific than `any`, but now realise this is correct and there's not any way to know the intended type of the field (without inferring it from the field `type`, which is assumptive as there's no link between fields and the underlying data type).

Thanks @rv-bparise!

PR #854, opened 2025-01-28
fix: ensure ctrl+i interactive toggle hotkey works on Windows

### LINKED ISSUE #851: Interactive hotkey not working on Windows
The interactive mode hotkey (meta + i) is not functional on windows. It might be conflicting with reserved hotkeys.

Try adding explicit `ctrl+i` or `ctrl+shift+i` in addition to existing hotkey (retain for Mac to avoid breaking change)

chrisvxd:
@FedericoBonel would you mind testing this now I've added an explicit hotkey? If this doesn't work, we can try a new binding.

Closes #851


FedericoBonel:
Hey Chris! 

I can confirm that this works for windows now:

[Editing_ _.webm](https://github.com/user-attachments/assets/8aaad769-dcfa-4778-94f2-38b60a0013d7)

Issue #876, opened 2025-02-04
Bug: resolveData props don't render until the user drags and drops a second time


FedericoBonel:
## Description  
When using an asynchronous `resolveData` function, the resolved props do not render on the canvas until the user interacts with the editor (e.g., drags and drops or changes something a second time).

## Steps to reproduce

1. Use an asynchronous function with awaited props inside `resolveData`:

```jsx
const config = {
  components: {
    SomeAsyncComponent: {
      resolveData: async () => {
        // "Hello World" doesn't render until something changes in the editor
        const result = await new Promise((resolve) => {
          setTimeout(() => {
            resolve("Hello World");
          }, 1000);
        });

        return {
          props: {
            title: result,
          },
        };
      },
      render: ({ title }) => {
        return (
          <div style={{ display: "flex", minHeight: "10rem" }}>
            <h1>{title}</h1>
          </div>
        );
      },
    },
  },
};

const Editor = () => {
  return <Puck config={config} data={{}} onPublish={onPublish} />;
};
```

2. Open the editor
3. Drag and drop the component with the async `resolveData`
4. Wait until spinner resolves without changing anything in the editor

## What happens

The editor doesn't show the resolved props after the promise resolves unless the user interacts with the editor again.

https://github.com/user-attachments/assets/9dc6c712-7703-4157-b1f6-0686d56677e4

## What I expect to happen

The editor should display the resolved props immediately after the spinner/promise resolves, without requiring additional user interaction.

chrisvxd:
Likely to be closed in 0.19

chrisvxd:
@FedericoBonel could you check whether this still occurs in 0.19? If not, please close it.

FedericoBonel:
It's fixed in the latest canary!

Issue #644, opened 2024-10-07
Improve rendering performance


chrisvxd:
Rendering performance can sometimes be slow, with re-renders occurring whenever the app context changes in all components that depend on it.

This is particularly notable when changing field values, which makes multiple changes to the app context (due to advanced features like field resolution, data resolution, etc). This is somewhat masked by a `50ms` debounce.  This works, but is inconsistent and causes a delay between user input and rendered output. `50ms` may also be insufficient for some environments, and shows a significant different between dev and prod environments.

We should improve rendering performance and eliminate the use of debounce entirely.

### Better memoizing

Review current memoization and improve it wherever possible.

### Split contexts

Break context down into smaller chunks.

### Use `use-context-selector`

Make use of the [`use-context-selector`](https://www.npmjs.com/package/use-context-selector) library, which implements [RFC119](https://github.com/reactjs/rfcs/pull/119).

### Replace app context with another state solution

Various state management libraries use selectors to avoid unnecessary re-renders.

### Virtualize components

Virtualize component rendering when out of view. 

## Considerations

- Impact on `usePuck` hook
- #556 will change how items are rendered
- #549 
- DropZone component relies on context to know structure of tree, causing entire tree re-renders when any item in the tree changes. Would be solved by dropzone/slot field.

IMythx:
i think [react-tracked](https://www.npmjs.com/package/react-tracked) could help with context unwanted renders issue

chrisvxd:
We might be able to go straight to [use-context-selector](https://github.com/dai-shi/use-context-selector)

Sarfraz-droid:
Hey, I was working inside puck and I found the underlying issue. Main issue is. On every drag, the entire page rerenders. Because it's currently using a nested architecture. I shifted it to storing objects in a global store. And rendering only the changed elements using zustand & useSyncExternalStore hook.

Could be something to consider but would require an entire state overhaul

Performance got improved by a lot! I guess using profiler puck had around 1500 rerenders on drag. It got reduced by 150.

chrisvxd:
@Sarfraz-droid State overhaul is definitely something we could consider as part of this. Did you do that via a fork? Would be interested to see it.

Sarfraz-droid:
I tried out those changes mostly on a private repo. By cloning puck editor to libs in a startup. But I would be happy to detach the same in a fork. Will share as soon as it's done.

FedericoBonel:
[...]

Issue #255, opened 2023-11-24
More portable data payload


chrisvxd:
## Problem

The current [data model](https://puckeditor.com/docs/api-reference/data) is hard to reason about, which makes it less portable and harder to build custom renderers on top of. Some concerns:

* `zones` are weird
* The Root component is treated differently to the other components
* DropZones rely on [context](https://puckeditor.com/docs/api-reference/components/drop-zone) or [render props](https://puckeditor.com/docs/api-reference/configuration/component-config#propspuckrenderdropzone), which are React-only paradigms, making it hard to implement #302 and negatively impacting performance

Example of a custom renderers:

* XML string (i.e. to produce React code)
* Non-React outputs

### Current data model

```json
{
  "content": [
    {
      "type": "Columns",
      "props": {
        "id": "Columns-1234"
      }
    }
  ],
  "root": { "props": { "title": "Page" } },
  "zones": {
    "Columns-1234:left": [
      {
        "type": "Heading",
        "props": { "title": "Hello, world" }
      }
    ]
  }
}
```

## Proposal

[react-from-json]( http://github.com/measuredco/react-from-json) implements a predictable AST based on, but not tied to, the React tree. It would allow for custom renders and increased portability.

### Proposed data model

```json
{ 
  "root": {
    "type": "Root",
    "props": {
      "title": "Page",
      "children": [
        {
          "type": "Columns",
          "props": {
            "left": [
              {
                "type": "Heading",
                "props": { "title": "Hello, world" }
              }
            ]
          }
        }
      ]
    }
  },
  "content": [], // Deprecated, but still supported
  "zones": {}, // Part of DropZone API which will be deprecated, but still supported
}
```

### New `slot` API

The `children` prop in the above will be powered by the `slot` field API, which implements a DropZone-like API but as a field. This makes data much more portable, as it's part of the same payload, whilst also taking advantage of the numerous field APIs (`defaultProps`, `resolveData`, etc).

Current proposal:

```tsx
const config = {
  fields: {
    leftColumn: { type: "slot" }
  },
  defaultProps: {
    leftColumn: {} // Unlike DropZones, you can set defaultProps for slots
  },
  render: ({ leftColumn }) => (
    <div>{leftColumn()}</div>
  )
}
```

DropZone's will be deprecated but retained to avoid breaking changes, but user's will be directed to the `slot` API instead.

### Considerations

* Implement migration using `migrate()` API
[...]

PR #877, opened 2025-02-04
fix: unable to update content with async resolveData function

do4Mother:
this is first time i made contribution in open source project. please let me know if i made some mistake. 😄 

so i have an issue regarding  #876, when i tried to debugging. `contentWithPreview` variable didn't update correctly if inserting component with async resolveData function.


chrisvxd (commented):
Hey @do4Mother! Thank you so much for your contribution (and flattered that it's your first ever).

Unfortunately this function is quite intentionally setup to avoid race conditions caused by Zustand and React rendering out of sync. I think your fix is likely to cause other issues, so I'm going to have to debug #876 myself further.

I'll leave this PR open, and if it's the correct fix, go with your contribution 👍

chrisvxd:
Can confirm this will be resolved in #644 and #255. Closing!

Issue #798, opened 2025-01-14
Metadata API


chrisvxd:
Add a metadata API to enable the user to inject their own data down the Puck tree. This can provide a viable alternative to state management solutions, and should work for server components.

### Proposal 1

Add a `metadata` prop to the Puck component:

```tsx
<Puck metadata={{ foo: "bar" }} />
```

Pass into components via the `puck.metadata prop:

```tsx
{
  render: ({ puck }) => <div>{puck.metadata.foo}</div>
}
```

### Proposal 2

Add a `mapMetadata` function to component config, with the result passed to all children

```ts
{
  fields: {
    example: { type: "text" }
  },
  mapMetadata: (metadata, props) => ({ ...metadata, example: props.example }),
}
```

Then, in a child component

```tsx
{
  render: ({ puck }) => <div>{puck.metadata.example}</div>
}
```

## Considerations

* Consider generic typing of metadata, especially when user is using Config or ComponentConfig types
* If making reactive, consider how to isolate rerenders

chrisvxd:
Tagging @jsjexpert so I can assign

chrisvxd:
Proposal 1 implemented in #882. Proposal 2 spun out into #910.

Issue #302, opened 2023-12-20
Multi-framework support


chrisvxd:
Puck should explore support for multi-framework support, including:

* Vue.js
* Svelte
* -> Insert other JS framework here <-

## Considerations

* The host application is React
* The components API is React
* The DropZone API uses context, which is _very_ React

## Related

* #255

## Proposals

No formal proposal yet, just some ideas:

* Keep core as React, and use web components or similar to wrap the host app
* Wrap component render methods using a compatibility layer
* Reconsider DropZone API to avoid context (see #255)
* Consider web components

peoray:
@chrisvxd any updated plan on this? I'm interested in creating a port for Vue

chrisvxd:
@peoray I think we might need to tackle #255 first, which I'm planning to explore within the next couple of months.

Interesting to hear your exploring a Vue port. Are you considering an approach similar to the ones outlined above?

peoray:
@chrisvxd I'm not familiar with Web Component, but after reading about it, it sounds like a good idea. My initial thought was to rewrite in Vue

tbinna:
Might be useful for this work: Atlassian has released a [framework agnostic drag and drop library](https://github.com/atlassian/pragmatic-drag-and-drop) (I believe after stopping their efforts on the [React-specific dnd library](https://github.com/atlassian/react-beautiful-dnd)).

chrisvxd:
Thanks @tbinna! We're not considering pragmatic-drag-and-drop as it doesn't support some of the functionality we require (previews, animations)

jamesjulich:
Hey! Just wanted to give a +1 for this idea. I briefly toyed around with the editor. I've got two suggestions related to this issue:

## Allow devs to export some intermediate representation of the UI (tree made up of JSON?) and let developers create their own render methods independent of Puck.

...now this...this would be powerful. Giving developers the ability to export their UI as an intermediate representation (ex: tree in JSON) would allow Puck to be used for rendering UIs on any platform. Puck could limit its "preview" window to web UIs. This seems like it would be the quickest way to start supporting multiplatform or non-React frameworks.

## Support multiple frameworks ALA Astro Islands and stick to web
(kinda spitballing here based off of some limited research, take with a grain of salt)

[...]

PR #882, opened 2025-02-06
Add metadata API

jsjexpert:
This PR is based on my own need but I hope it can be useful to others.

Puck (and Render) can now accept an externalData prop, ostensibly populated from a backend query such as on a Next.JS page. 
image
image

A new beforeRender param is exposed in Component Config allowing for render-time transformations without mutating the saved data,
image

This allows for template substitution using shared global data and provides a live preview of the transformed data
image
image
image


chrisvxd:
This is awesome @jsjexpert. However, I think this might be a similar implementation to the proposed #798 - what do you think?

jsjexpert:
Actually, yes! I did search issues before making the PR, but i didn't look for "metadata" and missed it. Feels like a good match for it, though.

I can rename it to metadata if you want.

Also,  my implementation only exposes the new input into the beforeRender function, but it could also be useful in others. The key issue sticking point was sending the variable name and not the value to be saved. My demo uses handlebars, but could use traverse and dot notation as preferred.

The metadata api, instead of being built in, is user generated within the beforeRender call, which allows for implementation flexibility.

Next steps or changes you need?

chrisvxd:
Okay, cool. There are still some gaps between this implementation and that proposed in #798. I think the solution in #798 is preferable, so if we're going to move forwards with this we should probably:

1. Rename `externalData` to `metadata`
2. Remove the handlebars details, as they're implementation-specific
3. Avoid introducing the new `beforeRender` API - I think you can achieve this with `resolveData`, please give it a go and see if it works
4. Add the `mapMetadata` API from the proposal (this can also be a subsequent PR)
5. Document all the things

If you don't want to rework this PR then no problem, I'll pick it up later.

jsjexpert:
@chrisvxd Sounds good. I'll try my hand at this. You can assign #798 to for tracking if needed.

chrisvxd:
Tried to assign but couldn't. Have self assigned to prevent anyone else working on it.

jsjexpert:
@chrisvxd That works :)

I've started refactoring, but i'm bumping against a core issue blocking the removal of the beforeRender API and handling all within resolveData, at least for the use case I have in mind:

resolveData mutates the saved content to substitute values, saving the final versions, so with 
`const metadata = {
  title: "hello world"
}`

instead of saving the link to metadata.title, the component prop is being changed to "hello world" and published with the substitution in place. 

The idea i had for the beforeRender APIis to apply a non-permanent transformation to the output, without mutating the input. This allows the use of a single Puck published template to be used across multiple pages. Is that what you see mapMetadata doing?

Alternatively, renderData could be extended to return 
`{
   props?: Partial<FieldProps>;
   readOnly?: Partial<Record<keyof FieldProps, boolean>>;
   renderProps: : Partial<FieldProps>;
 }`

This would give the flexibility to mutate the data and save by returning to props or to apply a render-time, non-destructive change to the output by returning renderProps.

The renderer would then take
`{
   ...props,
   ...renderProps
}`
as input for the final display.

Thoughts?

jsjexpert:
@chrisvxd ignore my last, I think i have a good approach. More to come

jsjexpert:
@chrisvxd Changes applied.

externalData is now metadata across the board.
metadata is exposed to resolveData and render.
Docs updated.
Demo updated to remove handlebars & to support metadata. A simple useMetadata function has been added to demonstrate with static data, and useDemoData has been updated to accept metadata input.

I think this is good to go.

packages/core/types/Data.tsx
```diff
@@ -52,3 +52,5 @@ export type Data<
   content: Content<Props>;
   zones?: Record<string, Content<Props>>;
 };
+
+export type MetaData = { [key: string]: any };
```
chrisvxd, line 56:
```suggestion
export type Metadata = { [key: string]: any };
```

I also think you need to run `npm run format` and commit the results to fix various formatting issue.

yarn.lock
```diff
@@ -7965,7 +7965,7 @@ hachure-fill@^0.5.2:
   resolved "https://registry.yarnpkg.com/hachure-fill/-/hachure-fill-0.5.2.tgz#d19bc4cc8750a5962b47fb1300557a85fcf934cc"
   integrity sha512-3GKBOn+m2LX9iq+JC1064cSFprJY4jL1jCXTcpnfER5HYE2l/4EfWSGzkPa/ZDBmYI0ZOEj5VHV/eKnPGkHuOg==
 
-handlebars@^4.4.3, handlebars@^4.7.7:
+handlebars@^4.4.3, handlebars@^4.7.7, handlebars@^4.7.8:
```
chrisvxd, line 7968: Nit: Can you revert the `yarn.lock` as we're not touching handlebars now?

apps/demo/config/index.tsx
```diff
@@ -107,6 +107,22 @@ export const initialData: Record<string, UserData> = {
...
+        },
+      },
+      {
+        type: "Space",
+        props: {
+          size: "96px",
+          id: "Space-1687298109537",
+          direction: "vertical",
+        },
+      },
```
chrisvxd, line 125: Can you drop these additions? No need to update the demo data here.

packages/core/lib/resolve-all-data.ts
```diff
@@ -16,12 +17,13 @@ export async function resolveAllData<
 >(
   data: Partial<Data>,
   config: Config,
+  metadata: MetaData = {},
```
chrisvxd, line 20: Slight bit of nuance needed here. This is technically a breaking change, but `onResolveStart` and `onResolveEnd` should not actually be public (see #899). I'm not sure there's any action on you here, but will think about it.

packages/core/types/Config.tsx
```diff
@@ -38,15 +38,18 @@ export type ComponentConfig<
...
   ) =>
     | Promise<{
         props?: Partial<FieldProps>;
         readOnly?: Partial<Record<keyof FieldProps, boolean>>;
+        renderProps?: Partial<FieldProps>;
       }>
     | {
         props?: Partial<FieldProps>;
         readOnly?: Partial<Record<keyof FieldProps, boolean>>;
+        renderProps?: Partial<FieldProps>;
```
chrisvxd, line 52: What are these `renderProps` type changes for?

packages/core/types/Config.tsx
```diff
@@ -58,6 +61,7 @@ export type ComponentConfig<
       lastData: DataShape | null;
     }
   ) => Promise<Partial<Permissions>> | Partial<Permissions>;
+  // mapMetadata?: (metadata:MetaData, props:Partial<FieldProps> | FieldProps) => {}
```
chrisvxd, line 64: Let's drop this comment if we're not implementing it now

apps/docs/pages/docs/api-reference/components/puck.mdx
```diff
@@ -113,6 +114,21 @@ export function Editor() {
 }
 ```
 
+### `metadata`
+
+Configure drag-and-drop behavior.
```
chrisvxd, line 119:
```suggestion
Provide additional data to the [`puck.metadata`](/docs/api-reference/configuration/component-config#puckmetadata) render prop provided to all components.
```

apps/docs/pages/docs/api-reference/components/render.mdx
```diff
@@ -16,10 +16,11 @@ export function Example() {
...
 
-| Param               | Example                      | Type                                 | Status   |
-| ------------------- | ---------------------------- | ------------------------------------ | -------- |
-| [`config`](#config) | `config: { components: {} }` | [Config](/docs/api-reference/config) | Required |
-| [`data`](#data)     | `data: {}`                   | [Data](/docs/api-reference/data)     | Required |
+| Param                   | Example                      | Type                                     | Status   |
+| ----------------------- | ---------------------------- | ---------------------------------------- | -------- |
+| [`config`](#config)     | `config: { components: {} }` | [Config](/docs/api-reference/config)     | Required |
+| [`data`](#data)         | `data: {}`                   | [Data](/docs/api-reference/data)         | Required |
+| [`metadata`](#metadata) | `metadata: {}`               | [Metadata](/docs/api-reference/metadata) | Optional |
```
chrisvxd, line 23: Optional status is denoted by `-`

```suggestion
| [`metadata`](#metadata) | `metadata: {}`               | [Metadata](/docs/api-reference/metadata) | - |
```

apps/docs/pages/docs/api-reference/components/render.mdx
```diff
@@ -73,3 +74,21 @@ export function Example() {
   );
 }
 ```
+
+### `metadata`
+
+Global metadata to be used in conjunction with data to render against the provided config. See the [`Metadata` docs](/docs/api-reference/metadata) for a full reference.
```
chrisvxd, line 80:
```suggestion
Provide additional data to the [`puck.metadata`](/docs/api-reference/configuration/component-config#puckmetadata) render prop provided to all components.
```

apps/docs/pages/docs/api-reference/configuration/component-config.mdx
```diff
@@ -56,18 +56,37 @@ const config = {
...
+| [`...props`](#props)                            | `{}`          | Object   |
 
 ##### `id`
 
 A unique ID generated by Puck for this component. You can optionally apply this, or use your own ID.
 
+##### `puck.metadata`
+
+Global metadata passed in to the `<Puck>` component. Can be used along with component props for rendering.
+Using metadata within the `render` API is non-destructive and will conserve the values saved in the component props.
```
chrisvxd, line 75: Think this can be simplified

```suggestion
Global metadata that is be passed in via the Puck [`metadata` prop](/docs/api-reference/components/puck#metadata).
```

apps/docs/pages/docs/api-reference/configuration/component-config.mdx
```diff
@@ -56,18 +56,37 @@ const config = {
...
-| -------------------------------------------- | ------------- | -------- |
-| [`id`](#id)                                  | `button-1234` | String   |
-| [`puck.dragRef`](#puckdragref)               | `null`        | Function |
-| [`puck.isEditing`](#puckisediting)           | `false`       | Boolean  |
-| [`puck.renderDropZone`](#puckrenderdropzone) | `() => {}`    | Function |
-| [`...props`](#props)                         | `{}`          | Object   |
+| Arg                                             | Example       | Type     |
+| ----------------------------------------------- | ------------- | -------- |
+| [`id`](#id)                                     | `button-1234` | String   |
+| [`puck.metadata`](/docs/api-reference/metadata) | `{}`          | Object   |
```
chrisvxd, line 62: Can you make alphabetical please?

apps/docs/pages/docs/api-reference/configuration/component-config.mdx
```diff
@@ -56,18 +56,37 @@ const config = {
...
+| [`puck.dragRef`](#puckdragref)                  | `null`        | Function |
+| [`puck.isEditing`](#puckisediting)              | `false`       | Boolean  |
+| [`puck.renderDropZone`](#puckrenderdropzone)    | `() => {}`    | Function |
+| [`...props`](#props)                            | `{}`          | Object   |
 
 ##### `id`
 
 A unique ID generated by Puck for this component. You can optionally apply this, or use your own ID.
 
+##### `puck.metadata`
```
chrisvxd, line 72: Can you make alphabetical please?

apps/docs/pages/docs/api-reference/configuration/component-config.mdx
```diff
@@ -358,6 +377,20 @@ const resolveData = async ({ props }, { changed }) => {
...
+```tsx copy {2-4} /changed/1 filename="Example only updating 'resolvedTitle' when 'title' changes"
+const resolveData = async ({ props }, { metadata }) => {
+  return {
+    props: { title: metadata.title || props.title },
+  };
+};
+```
+
+Note that returning metadata in the resolveData API will overwrite the prop values. For non-destructive use of metadata, refer to the [`render(puck.metadata)` docs](#puckmetadata).
+
```
chrisvxd, line 393: Hm I hadn't thought that we might need to expose this in `resolveData`, but I think it makes sense 🤔

packages/core/lib/__tests__/use-metadata.spec.tsx
```diff
@@ -0,0 +1,172 @@
...
+            prop: true,
+          },
+        };
+      },
+      render: () => <div />,
+    },
+  },
+};
+
+describe("use-metadata", () => {
```
chrisvxd, line 60: There is no `use-metadata` hook implemented, so I suspect these would be better added to the `useResolveData` tests

chrisvxd (changes requested):
Looks fantastic! Thank you!

I've left some comments. I might do another pass on the docs once this is merged.

Additionally, I may hold off merging until we're closer to a feature release.

packages/core/lib/resolve-all-data.ts
jsjexpert: Adding Metadata here was required based on the types but it could be changed from `metadata` to `{}` to discourage use.

packages/core/types/Config.tsx
jsjexpert: Removed. Artifact from a previous implementation attempt.

apps/docs/pages/docs/api-reference/configuration/component-config.mdx
jsjexpert: Since metadata can be anything, a value contained within could be used in resolveData as a conditional operator. For example, `if (metadata.product.type === "water") return { props: { description: "It's very wet" } };`

packages/core/lib/__tests__/use-metadata.spec.tsx
jsjexpert: I've updated the tests but an additional review would be best.

apps/docs/pages/docs/api-reference/components/puck.mdx
```diff
@@ -29,6 +29,7 @@ export function Editor() {
 | ----------------------------------------------------- | -------------------------------------------------- | -------------------------------------------------- | ------------ |
 | [`config`](#config)                                   | `config: { components: {} }`                       | [Config](/docs/api-reference/configuration/config) | Required     |
 | [`data`](#data)                                       | `data: {}`                                         | [Data](/docs/api-reference/data)                   | Required     |
+| [`metadata`](#metadata)                               | `metadata: {}`                                     | [Metadata](/docs/api-reference/metadata)           | -            |
```
chrisvxd, line 32: Should be alphabetical

apps/docs/pages/docs/api-reference/components/puck.mdx
```diff
@@ -113,6 +114,21 @@ export function Editor() {
 }
 ```
 
+### `metadata`
+
+Provide additional data to the [`puck.metadata`](/docs/api-reference/configuration/component-config#puckmetadata) render prop provided to all components.
+
```
chrisvxd, line 120: Should be alphabetical

chrisvxd (commented):
Hey @jsjexpert! This is great! I'm _provisionally_ approving it, but have had some further thoughts...

#897 rewrites much of Puck's internal state management to reduce the number of re-renders. This will likely introduce conflicts, but more importantly it has changed the lens I'm looking at this PR with: I now realised that a `puck.metadata` prop will cause a re-render of all components if the metadata changes.

Perhaps this is fine, but I can't help but wonder if there is a better way to handle it, so components can opt-in to receive metadata so they don't re-render by default.

## Proposals

I don't think we should jump on addressing this just yet, but I think we should discuss to make sure we're not implementing an API that will cause issues down the road. To understand the best implementation, it might be better to hear more use-cases for the metadata (i.e. how often does the metadata change).

### Proposal 1 - Do nothing

Just accept that all components re-render if metadata changes. If the user needs more fine-grained rendering control, they can implement their own state solution.

### Proposal 2 - Add an opt-in param to the config 

Add something to enable the user to opt-in to receiving the prop

```ts
{
  hasMetadata: true, // Option 1
  getMetadata: (metadata) => metadata, // Option 2, alternatively call it `resolveMetadata` and make async
  render: () => {}
}
```

### Proposal 3 - Switch to a hook

Use a hook to access metadata, although this is a React-only paradigm and will hamper multi-framework efforts (#302)

```tsx
{
  render: () => {
    const metadata = useMetadata();

    return <div />;
  }
}
```

apps/demo/lib/use-demo-data.ts
```diff
@@ -36,7 +38,9 @@ export const useDemoData = ({
 
   useEffect(() => {
     if (data && !isEdit) {
-      resolveAllData<Props, RootProps>(data, config).then(setResolvedData);
+      resolveAllData<Props, RootProps>(data, config, metadata).then(
+        setResolvedData
+      );
     }
   }, [data, isEdit]);
```
chrisvxd, line 45: Linting failing - metadata needs adding to this hook.

jsjexpert: Corrected in [1df8b99](https://github.com/measuredco/puck/pull/882/commits/1df8b99806ef1895d495d0a97008a26c1c5eed3b)

jsjexpert:
@chrisvxd 

My thinking behind this whole implementation was to provide a global data set (from a CMS, database, etc) that can be used within Puck at the component and field levels. The metadata is fetched from source at the start of the request, then fed into Puck for use. I would expect that if the underlying data changes, users would expect those changes to trickle down to the components.

If freezing the metadata is required, it's probably best handled like the initial Data is and added to the appState. An additional reducer action for setMetadata would then be required to trickle down the updates. 

Personally, I tend to think that the "do nothing" approach is the correct one, leaving it to the developer implementation to determine whether they need to freeze metadata state or not.

jsjexpert:
@chrisvxd just fixed the formatting issues that caused the puck-demo check to fail

Issue #886, opened 2025-02-09
Expose Header & Canvas as Puck components


shannonhochkins:
While building a custom UI, I encountered limitations with the current `Puck` component. Specifically, the `Header` component (which includes the undo/redo/save buttons) and the `Canvas` component (containing viewport controls and preview device buttons) are not exposed under the `Puck` namespace, restricting their usage in custom implementations.

**Proposed Changes:**

1. **Expose `Header` Component:**

- **Refactor**: Move the `Header` component out of the main `Puck` module into an isolated component.
- **Styling**: Share CSS styles from the parent style module (note: this may be subject to future refactoring).
- **Namespace Exposure**: Expose the Header component under the Puck namespace.

2. **Expose `Canvas` Component:**

- Make the Canvas component accessible under the Puck namespace to allow direct usage and customization.

3. Deprecate `renderHeader` and `renderHeaderActions`:

- Introduce a deprecation flag for `renderHeader` and `renderHeaderActions` instead of using inline comments.
- These flags provide a clearer indication of deprecation and can be removed if deemed unnecessary.

4. Refactor PuckProps:

- Move PuckProps to an internal interface to promote reusability.
- Utilize this internal interface within the Header component for consistency and maintainability.

**Motivation:**

Exposing the `Header` and `Canvas` components under the `Puck` namespace will:

- Enhance Flexibility: Allow developers to utilize and customize these components directly in their UIs.
- Improve Developer Experience: Provide better access to essential UI elements, facilitating more seamless integrations.
- Promote Reusability: Refactoring PuckProps enables consistent prop usage across components, reducing redundancy.

shannonhochkins:
You can close this if you like Chris 👍

PR #887, opened 2025-02-09
feat: Exposing header/canvas components

shannonhochkins:
This PR aims to fix [issue](https://github.com/measuredco/puck/issues/886).

**Next Steps:**

Before proceeding with detailed documentation and further refinements, I seek feedback on the proposed changes to ensure alignment with the project's goals and to assess the likelihood of this PR being merged.

I did test this with the demo app, both components render, obviously because the Canvas component isn't rendered in the standard UI, it does require additional css to stretch the Canvas element to fill the viewport, again something we can clean up after feedback!

### DIFF
--- a/packages/core/components/Puck/components/Header/index.tsx
+++ b/packages/core/components/Puck/components/Header/index.tsx
@@ -1,0 +1,226 @@
+import {
+  useCallback,
+  useMemo,
+  useState,
+  useEffect,
+} from "react";
+
+import type {
+  Overrides,
+  UserGenerics,
+  Config,
+} from "../../../../types";
+import { Button } from "../../../Button";
+import {
+  ChevronDown,
+  ChevronUp,
+  Globe,
+  PanelLeft,
+  PanelRight,
+} from "lucide-react";
+import { Heading } from "../../../Heading";
+import { IconButton } from "../../../IconButton/IconButton";
+import getClassNameFactory from "../../../../lib/get-class-name-factory";
+import { MenuBar } from "../../../MenuBar";
+import styles from "../../styles.module.css";
+import { DefaultOverride } from "../../../DefaultOverride";
+import { useAppContext } from "../../context";
+import { type PuckProps } from "../..";
+
+const getLayoutClassName = getClassNameFactory("PuckLayout", styles);
+
+
+export function Header<
+  UserConfig extends Config = Config,
+  G extends UserGenerics<UserConfig> = UserGenerics<UserConfig>
+>({
+  renderHeader,
+  onPublish,
+  headerPath,
+  headerTitle,
+  renderHeaderActions,
+  onMenuStateChange,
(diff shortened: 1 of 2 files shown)


shannonhochkins:
I just noticed the loadedOverrides in the main Puck component too, which is obviously a bug with this PR as the header will not be overridable via a plugin currently

shannonhochkins:
Hey @chrisvxd Thanks for the reply!

## Canvas

I did look into the `ViewportControls` component first as this was one I actually wanted to change the UI of the most and expose the scaling functionality/css, however that simply seemed like more work than simply exposing the Canvas component itself. It's not impossible, however the scaling / zoom settings is used for the DraggableComponent so it makes it a tough one to move to an isolated component because of the callbacks it's using, may be able to use a HOC to help with this, happy to give it a crack.

I did get this to work out of the box with some style changes, granted, not ideal to have to get users to do this on their side, this is something I whipped together quite quickly, I'm sure any capable software engineer is able to make these style changes if they want to use the Canvas :) 

```ts
<Puck {...props}>
    <div className="canvas-wrapper">
        <Puck.Canvas />
    </div>
</Puck>
```

With this simple example, I did have to add the following styles, #root is my react entry point, the rest should be self explanitory!

```css
body, html {
  width: 100%;
  height: 100%;
}
* {
  box-sizing: border-box;
}

body {
  height: 100%;
  margin: 0;
}

#root, .Puck {
  width: 100%;
  height: 100%;
}

.Puck > div:has(.puck-editor-wrapper) {
  width: 100%;
  height: 100%;
}

[class*="_PuckCanvas_"] {
  width: 100%;
}

.canvas-wrapper {
  display: flex;
  width: 100%;
  height: 100%;
  align-items: stretch;
  justify-content: stretch;
}
```

## Header

I did re-create some of the components on my side, the header itself isn't wildly customizable, so I think it's fine to let users re-create their own header if need be, but also the main reason I suggsted this change is when users want to create their own interface, the default puck props allow you to hide/show sidebar controls, however it only hides them visually, so if someone doesn't want the outline for example, or they want to render it elsewhere, it's technically going to be rendered twice, so I figured having individual parts of the existing UI exportable may be useful, after diving a little deeper into my current UI I think it may be best to simply leave it up to the user to re-create their own Header as it really doesn't do too much.

```ts
import { IconButton, usePuck } from '@measured/puck';
import { Undo2Icon } from "lucide-react";

export function Undo() {
  const { history } = usePuck();
  const { hasPast, back } = history;
  return <IconButton title="undo" disabled={!hasPast} onClick={back}>
    <Undo2Icon size={21} />
  </IconButton>;
}
```

shannonhochkins:
I'll close this an the related ticket, unfortunately the zoom settings are the problem here for me as the draggable components use this value internally as well and there's no way for me to update the apostate, I'll have to implement my own responsive functionality for this

PR #916, opened 2025-02-25
Small feature to introduce step prop for number fields

shannonhochkins:
This introduces a very small change to add a `step` attribute for number field types.

Usage - I have a field for opacity, it's value should be between 0-1, the default step value for number fields is 1, so I can't use my keyboard / arrow buttons on the field to increment a smaller amount without typing the value myself.

This is a low level feature change - unsure if i've missed anything else in the repo that needs to be updated!


chrisvxd (changes requested):
Looks great! Thanks @shannonhochkins!

Just has some conflicts, and needs formatting (`npm run format` + commit).

shannonhochkins:
> Looks great! Thanks @shannonhochkins!
> 
> Just has some conflicts, and needs formatting (`npm run format` + commit).

Of course there was a change to the same files I was working on since i pushed :D - Fixed and ran format

chrisvxd (approved):

Issue #926, opened 2025-03-07
Nested Arrays Fields Can't Be Reordered


FedericoBonel:
## Description  
Nested array items (arrays inside arrays) are no longer draggable for reordering. This functionality worked in previous versions but is now broken.  

## Steps to Reproduce  
1. Create a basic Puck application with the following component config:  

    ```jsx
    const config = {
      components: {
        TestNestedArray: {
          fields: {
            array: {
              type: "array",
              arrayFields: {
                nested: {
                  type: "array",
                  arrayFields: {},
                },
              },
            },
          },
          render: () => {
            return <h1>Test array</h1>;
          },
        },
      },
    }
    ```

2. Navigate to the editor.  
3. Drag and drop a `TestNestedArray` component into the canvas.  
4. Add items to the array field, and add sub-items to each item.  
5. Try to drag and drop the nested items to reorder them.  

## What Happens  
The nested items are not draggable.  

## What I Expect to Happen  
The nested items should be draggable to reorder—this used to work in previous versions.  

## Video  
[https://github.com/user-attachments/assets/e86ae96b-7d53-4175-a7ff-768f4a049481](https://github.com/user-attachments/assets/e86ae96b-7d53-4175-a7ff-768f4a049481)

PR #927, opened 2025-03-07
fix sorting nested arrays

jcbbb:
Closes: https://github.com/measuredco/puck/issues/926


chrisvxd (changes requested):
Thanks @jcbbb. Unfortunately this change results in other fields being draggable (try selecting text in a text field), which is not intentional. Is it possible to isolate this further?

chrisvxd:
Closing in favour of #970

Issue #879, opened 2025-02-05
Access breadcrumbs in custom interfaces


chrisvxd:
Make breadcrumbs accessible for custom interfaces.

### Proposal 1

Expose a <Puck.Breadcrumbs> component

### Proposal 2

Expose the useBreadcrumbs hook

### Proposal 3

Expose another hook related to paths

## Considerations

* How will data portability and the slot API impact this?

shannonhochkins:
Whilst having components like Puck.Breadcrumbs are useful for very quick UI layouts - Personally, I'd rather have the data available to render my own UI which i realise opens a can of worms in terms of complexity, but either way I appreciate the ticket!

PR #939, opened 2025-03-18
Breadcrumb label use block label if available

shadowlik:
# Breadcrumb: Use Block Label if Available  

## 🙏 Huge Thanks to the Puck Team!  
First off, I just want to say that I **love** this project—big thanks to the Puck team for building such an awesome tool! ❤️  

## Summary  
This change enhances the breadcrumb functionality by allowing it to use the **block label** when available, instead of defaulting only to the block type.  

## Why?  
Currently, the breadcrumb always displays the block type, which limits flexibility in certain scenarios. While Puck **doesn't natively support i18n**, I'm working on a **workaround** to introduce some level of localization. This change makes my workaround a bit more **robust** by allowing the breadcrumb to display a translated block label instead of just the block type.  

## Changes  
- Modified breadcrumb logic to prefer the **block label** over the **block type** if available.  
- Ensured fallback to the block type when no label is present.  

## Impact  
- Enables more **flexibility** for custom implementations.  
- Makes my **i18n workaround** more reliable.  

## Testing  
- ✅ Verified that block label is used when present.  
- ✅ Ensured fallback to block type when no label exists.  

## Screenshots
Bellow a usage example, the component type is Flex with a label in Portuguese (Linha)

Screenshot 2025-03-18 at 8 53 09 PM

### DIFF
--- a/packages/core/lib/use-breadcrumbs.ts
+++ b/packages/core/lib/use-breadcrumbs.ts
@@ -1,10 +1,10 @@
 import { useContext, useMemo } from "react";
-import { dropZoneContext, PathData } from "../components/DropZone/context";
+import { DropZoneContext, dropZoneContext, PathData } from "../components/DropZone/context";
 import { useAppContext } from "../components/Puck/context";
 import { getZoneId } from "./get-zone-id";
 import { rootDroppableId } from "./root-droppable-id";
 import { ItemSelector } from "./get-item";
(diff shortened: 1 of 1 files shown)


FedericoBonel:
Might be related with #879

shadowlik:
Definitely related, will close this PR and follow that issue, let me know if I can be of any help! Appreciate you!

Issue #656, opened 2024-10-15
Viewport Control is in front of Action Bar


theophilhenry:
<img width="936" alt="image" src="https://github.com/user-attachments/assets/2a02e3b2-3549-4918-8a6a-0dbf0dbab71b">

Hi, I'm currently in Puck version "@measured/puck": "^0.16.1"

In the top current of my items, the viewport control is in front of my action bar (to delete and etc). I can't see the action bar. Is there any workaround for this? Thanks!

bakikucukcakiroglu:
I'm having the same issue!

chrisvxd:
This is an issue when the component is the first item on the page, and smaller than action bar.

There's not any good fix yet, but I can look at using some JS to account for this, similar to what I'm doing for the new drag-and-drop engine if the item is too narrow and the action bar clips off the left of the screen.

jwartofsky-yext:
I am running into the same issue as-well

PR #946, opened 2025-03-19
fix: actionbar clipping past top bound

camhammel:
closes https://github.com/measuredco/puck/issues/656

(removed "Header" content in screen recordings below to replicate issue)

before screen recording:
https://github.com/user-attachments/assets/024c51da-b942-4a4d-b1bb-42537c8c4bb2

after screen recording:
https://github.com/user-attachments/assets/56004b5f-a857-401b-b689-6f48e1bf163d

### DIFF
--- a/packages/core/components/DraggableComponent/index.tsx
+++ b/packages/core/components/DraggableComponent/index.tsx
@@ -445,6 +445,7 @@ export const DraggableComponent = ({
 
           // Modify position if it spills over frame vertically
           if (exceedsBoundsTop) {
+            el.style.transformOrigin = "left top";
             el.style.top = "0px";
           }
         }
--- a/packages/core/components/DraggableComponent/index.tsx
+++ b/packages/core/components/DraggableComponent/index.tsx
@@ -439,8 +439,8 @@ export const DraggableComponent = ({
 
           // Modify position if it spills over frame
           if (exceedsBoundsLeft) {
-            el.style.left = "0px";
             el.style.transformOrigin = "left top";
+            el.style.left = "0px";
           }
           
           if (exceedsBoundsTop) {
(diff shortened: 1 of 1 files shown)


packages/core/components/DraggableComponent/index.tsx
```diff
@@ -434,12 +434,20 @@ export const DraggableComponent = ({
...
-          // Modify position if it spills over frame
+          // Modify position if it spills over frame horizontally
           if (exceedsBoundsLeft) {
             el.style.transformOrigin = "left top";
             el.style.left = "0px";
           }
+
+          // Modify position if it spills over frame vertically
+          if (exceedsBoundsTop) {
+            el.style.transformOrigin = "left top";
```
chrisvxd, line 448: Should probably be `right top`, but we now also need to account for the scenario where this exceeds both left and top

chrisvxd (changes requested):
Thanks @camhammel! Unfortunately this isn't achieving the desired effect for me - see screenshot.

image

packages/core/components/DraggableComponent/index.tsx
camhammel: updated! seems to handle each case correctly now from what I can tell.

packages/core/components/DraggableComponent/index.tsx
```diff
@@ -434,12 +434,21 @@ export const DraggableComponent = ({
...
+          const exceedsBoundsTop = diffTop < 0;
 
           // Modify position if it spills over frame
           if (exceedsBoundsLeft) {
             el.style.transformOrigin = "left top";
             el.style.left = "0px";
           }
+
+          if (exceedsBoundsTop) {
+            el.style.top = "0px";
```
chrisvxd, line 447: Could you try setting this as `12px` so the action label doesn't butt-up against the component?

I also thought about moving it up to the top of the component, but then there's a risk the overlay awkwardly covers the contents, especially with smaller items.

Without 12px:

image

With 12px:

image

chrisvxd (changes requested):
Thanks @camhammel. This now works for me.

One further comment.

PR #957, opened 2025-03-24
fix: input file on component type array not working

### LINKED ISSUE #933: Input file on component type array not working
I have a need to upload more than 1 image, therefore I use the array component, but what happens is that the file input doesn't work, I've tried all versions.
The desktop window for selecting file does not appear. but when it is done outside the component array it works

```
 file: {
            type: 'array',
            arrayFields: {
                url: {
                    type: 'custom',
                        const handleFileChange = (event) => {
                            const file = event.target.files?.[0];
                            if (file) {
                                console.log("file name:", file.name);
                            }
                        }
                        return (
                            <input
                                type="file"
                                onChange={handleFileChange}
                                accept="image/*"
                                className="p-5"
                            />
                        )
                    }
                }
            }
        }
```

Image

christian-tchaikovsky:
This PR is intended to fix a bug with the file-type field.

It's possible that `e.preventDefault()` in onClick was addressing some issue, but during local testing, I didn't notice any changes in the behavior of the array-type field. Clicking on the file-type field works as expected.

Closes #933

### DIFF
--- a/packages/core/components/AutoField/fields/ArrayField/index.tsx
+++ b/packages/core/components/AutoField/fields/ArrayField/index.tsx
@@ -160,9 +160,6 @@ export const ArrayField = ({
             hasItems: Array.isArray(value) && value.length > 0,
             addDisabled,
           })}
-          onClick={(e) => {
-            e.preventDefault();
-          }}
         >
           <div className={getClassName("inner")} data-dnd-container>
             {localState.arrayState.items.map((item, i) => {


christian-tchaikovsky:
If `e.preventDefault()` was addressing some issue. This code also works as expected:

```js
onClick={(e) => {
  if (e.target === e.currentTarget) {
    e.preventDefault();
  }
}}
```

feryrmdhn:
+1

feryrmdhn:
Is there any team from _measured/puck_ that can confirm for this merge? I need this feature fixed ASAP please

chrisvxd (approved):
Just catching up, thanks @christian-tchaikovsky!

PR #958, opened 2025-03-24
fix: reordering items in array fields

### LINKED ISSUE #940: Issue with Reordering Items in Array Fields
In the latest version (`0.18.2`), an issue occurs when changing the order of items in a field with the `array` type. If there are multiple fields of this type, the behavior breaks for all except the first one. For example, when trying to reorder an item, it jumps to the first array field.

Can't reproduce in `0.18.1`

The issue persists in the latest available version: `0.19.0-canary.0ea6ce4`

Reordering Bug in Array Fields Occurs on the Website: https://puckeditor.com/v/0.18.2/docs/api-reference/fields/array

MacOS 15.3.1
Browser: Chrome 134.0.6998.89

https://github.com/user-attachments/assets/2611450b-826b-4693-94f7-758f8633eb1f

christian-tchaikovsky:
This PR fixes an issue with sorting elements in an array-type field, where a moved element was always bound to the first container.

Closes #940


chrisvxd (approved):
Thanks for tidying up my messy array field refactor! 🙏

Issue #963, opened 2025-03-31
`params.changed` flags all fields as `true` on initialization in `resolveData` method


christian-tchaikovsky:
In version `0.18.2` and the latest version (`0.19.0-canary.0a3e56e`), during component initialization in the `resolveData` method, all fields in `params.changed` are set to `true`.

Here’s an example:

```jsx
import React from "react";

import type { ComponentConfig } from "@measured/puck";

import { delay } from "@/helpers/delay/delay";
import { Media as _Media } from "@/components/Media/Media";
import { uploader } from "@/configs/puck/_fields/uploader/uploader";

import type { Props as MediaProps } from "@/components/Media/Media";

export type Props = MediaProps;

export const Media: ComponentConfig<Props> = {
  label: "Media File",
  fields: {
    type: {
      type: "radio",
      label: "File",
      options: [
        { label: "Image", value: "image" },
        { label: "Video", value: "video" },
      ],
    },
    src: uploader,
  },
  defaultProps: {
    type: "image",
    src: null,
  },
  resolveFields: (data, params) => {
    return {
      ...params.fields,
      src: data.props.type === "image"
        ? uploader({ accept: "image/*" })
        : uploader({ accept: "video/*" }),
    };
  },
  resolveData: async (data, params) => {
    console.log(data, params.changed);

    if (params.changed.type && data.props.src) {
      // TODO: Remove file on change type
      await delay(3000);

      return {
        props: { ...data.props, src: null },
      };
    }

    return data;
  },
  render: ({ type, src }) => {
    return (
      <_Media
        src={src}
        type={type}
      />
    );
  },
};
```

In the example above, the condition triggers immediately, even though the fields haven't been modified yet. This results in unintended code execution during initialization.

chrisvxd:
I'm unsure if this is a bug. We want `resolveData` to trigger on insertion, and at that point it's fair to say all data has changed.

If you don't want to trigger on insertion, you could check if `id` has changed. If it has, you can assume the component is new and skip your resolver.

christian-tchaikovsky:
@chrisvxd Thank you for your feedback!

The `resolveData` method is definitely supposed to be called when inserting/creating a component, but for some reason, at the moment of creation, all fields in `params.changed` have a state of `true` (including `id`), even though the fields haven't actually changed. I believe that in `params.changed`, a field should be `true` only when that specific field has been modified.

I also noticed that if you duplicate a component, the `resolveData` method does not trigger during creation.

[...]

PR #964, opened 2025-03-31
fix: prevent all fields in `params.changed` from being `true` on initialization in `resolveData`

christian-tchaikovsky:
In this PR, the issue where all fields in `params.changed` were incorrectly set to `true` during initialization in the `resolveData` method has been fixed. This prevented unnecessary code execution on component mount.

Closes: #963


chrisvxd (commented):
Unfortunately this prevents resolveData from firing on initial load, and breaks the render path - see missing Hero. I'm going to close this for now as resolveData is being reworked in 0.19.

image

PR #984, opened 2025-04-08
Add slots API

chrisvxd:
Edit: this PR was refusing to merge. Now in #999.


b-barry:
I've been testing the PR and have a couple of observations and suggestions:

### Issue: resolveData not executing on initial drop

When first dropping a component into a slot, I notice that `resolveData` isn't being executed automatically. We need to trigger a field change to make it run.  I think `resolveData` should execute immediately when a component is first added to a slot to ensure we start with properly initialized data.

### Enhancement: Support for different rendering modes

Currently, slot content is always rendered in edit mode. I think we should extend this to support read-only rendering modes.

I'd suggest adding a `mode` property to the slot field definition:

```javascript
fields: {
  children: { 
    type: "slot", 
    mode: 'edit' | 'render' | 'server-render' 
  }
},
```

Where:
- `edit`: Current behavior (fully editable)
- `render`: Uses the `Render` component exported by Puck
- `server-render`: Uses the `ServerRender` component

This would give component authors more flexibility in how slots are presented while still keeping the field editable in the right sidebar. 

Let me know what you think about these suggestions. The slots API is a powerful addition to Puck's capabilities, I love it

PR #985, opened 2025-04-10
feat: include context data in config

christian-tchaikovsky:
Added `context`, which can be used during component creation. Currently, there is no easy way to pass data from the API into `ComponentConfig`.

A `context` property has been added to `Config`, which can accept any data type. Now, a component can be declared in two ways:
1. Using `ComponentConfig`
2. Using `ComponentConfigFunction`

Example:

```tsx
export type Context = {
  package: string;
};

const HeadingBlock: ComponentConfigFunction<HeadingBlockProps, Context> = (context) => ({
  fields: {
    title: {
      type: "text",
    },
  },
  defaultProps: {
    title: `${context.package} v${context.version}`,
  },
});

const config = {
  context: {
    package: "@measured/puck",
    version: "x.x.x",
  },
  components: {
    HeadingBlock,
  },
};
```

### DIFF
--- a/apps/demo/app/custom-ui/[...puckPath]/client.tsx
+++ b/apps/demo/app/custom-ui/[...puckPath]/client.tsx
@@ -10,6 +10,7 @@ import { useDemoData } from "../../../lib/use-demo-data";
 import { IconButton, createUsePuck } from "@/core";
 import { ReactNode, useEffect, useRef, useState } from "react";
 import { Drawer } from "@/core/components/Drawer";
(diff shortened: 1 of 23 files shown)


chrisvxd:
Hey @christian-tchaikovsky! I think the new `metadata` API will probably cover you here: https://puckeditor.com/v/canary/docs/api-reference/components/puck#metadata

One notable difference - metadata exists as a prop, instead of the root of the config.

I'm going to close this out.

PR #990, opened 2025-04-25
Prep 0.19 release

chrisvxd:
Very early draft PR to start capturing features going out in 0.19. Subject to change.


FedericoBonel:
Hey @jedrzejowski,

You're right. Initially we thought we could bring support for React 17 back, but later realized that some of our main dependencies had dropped it completely.  
Because of that, we had to roll back on supporting React 17. Sorry for the confusion!

PR #1014, opened 2025-05-07
fix: typing issue in UsePuckData

lukstei:
Fixes a small typing issue in `UsePuckData` where the non typed `Config` is returned for `config` instead of the typed `UserConfig`


chrisvxd (approved):
Good catch, thanks @lukstei!

PR #1016, opened 2025-05-08
fix: Iframe selection collision

### LINKED ISSUE #880: Iframe selection collision
## Description  
The selection of the preview iframe is causing issues because it is too vague. Currently, it uses `document.querySelector("iframe")`, which can lead to collisions and unexpected behavior if the user has other iframes on the same page where the editor is rendered.  

## Steps to Reproduce  
1. Add an iframe with an external domain as the `src`  on the same page where the editor is being rendered.  
2. Navigate to the editor in the browser and open the console.  
3. Attempt to drag and drop a component.  

## What Happens  
The component cannot be dragged and dropped, and multiple error messages appear when moving the mouse across the screen.  

## Expected Behavior  
The editor should not collide with other iframes and should allow users to drag and drop components or use it seamlessly, regardless of other content rendered on the same page.  

**Screenshot:**  
Image

FedericoBonel:
Closes #880


chrisvxd (approved):
Thanks for sorting this!

PR #1031, opened 2025-05-12
Fix various bugs with slots

### LINKED ISSUE #1024: Bug: Root level object values become undefined in custom fields
## Description

When using custom fields at the root level with object values, the value becomes `undefined` while selecting any nested component, resulting in unexpected undefined pointer errors.

This does not happen in version 0.18.3.

## Environment

- Puck version: "0.19.0-canary.f6398bca"

## Steps to reproduce

1. Render the Puck editor with the following config:
```tsx
{
  root: {
    fields: {
      rootObject: {
        type: "custom",
          console.log(value);
          return (
            <>
              <input
                value={value.field1}
                onChange={(e) => {
                  onChange({
                    ...value,
                    field1: e.target.value,
                  });
                }}
              />
              <input
                value={value.field2}
                onChange={(e) => {
                  onChange({
                    ...value,
                    field2: e.target.value,
                  });
                }}
              />
            </>
          );
        },
      },
    },
    defaultProps: {
      rootObject: {
        field1: "Field 1",
        field2: "Field 2",
      },
    },
    render: (fields) => {
      return (
        <div>
          <h1>{fields.rootObject.field1}</h1>
          <h2>{fields.rootObject.field2}</h2>
          <div>{fields.children}</div>
        </div>
      );
    },
  },
  components: {
    Text: {
      fields: {
        text: {
          type: "text",
        },
      },
      defaultProps: {
        text: "Hello World",
      },
      render: (fields) => {
        return <div>{fields.text}</div>;
      },
    },
  },
}
```

2. Navigate to the editor  
3. Drag and drop a component into the canvas

## What happens

The editor crashes because the root object value becomes `undefined` in the custom field.

## What I expect to happen

The object value should not become `undefined`, since it has default values.

## Additional Media

https://github.com/user-attachments/assets/b386080e-7698-48a4-b0c3-74b0888cc83c

### LINKED ISSUE #1025: Bug: onChange callback is called excessively in 0.19 canary when interacting with the editor
## Description

In the canary version, the `onChange` callback is being triggered multiple times when interacting with the editor. This behavior does not occur in `0.18.3`, where the callback is only fired once per interaction.

This excessive firing could cause performance issues or unintended side effects, especially when `onChange` updates state or triggers re-renders.

## Environment

- Puck version: 0.19.0-canary.f6398bca

## Steps to reproduce

1. Render the `<Puck />` component using the following setup:

```tsx
import { Puck } from "@measured/puck";
import "@measured/puck/puck.css";

export default function Editor() {
  return (
    <div>
      <Puck
        config={{
          components: {
            Text: {
              fields: {
                text: {
                  type: "text",
                },
              },
              defaultProps: {
                text: "Hello World",
              },
              render: (fields) => {
                return <h2>{fields.text}</h2>;
              },
            },
          },
        }}
        onChange={(data) => {
          console.log("data");
        }}
        data={{}}
      />
    </div>
  );
}
```

2. Run the application in development mode.
3. Navigate to the editor and open devtools.
4. Drag and drop a component into the canvas.

## What happens

In version 0.18.3, the `onChange` callback is triggered once per interaction.  
In the canary version, the same action triggers the callback 39 times.

## What I expect to happen

The `onChange` callback should only be triggered once per user interaction, matching the behavior in version 0.18.3.

## Additional Media

https://github.com/user-attachments/assets/1b09e967-3af1-46bf-9a7b-b1efab3fa79d

### LINKED ISSUE #1027: Bug: Deprecation warning for DropZone is triggered when dropping a slot field
## Description

A deprecation warning for `DropZone` is shown even when dragging and dropping a `slot` field, which should not happen. This occurs even if no actual `DropZone` is dropped into the editor.

In the config below, dragging a `DropZone` correctly triggers the warning (expected), but dragging a `slot` does as well (unexpected).

## Environment

- Puck version: 0.19.0-canary.f6398bca

## Steps to reproduce

1. Render the Puck editor using the following config:

```tsx
{
  components: {
    Container: {
      render: () => {
        return <DropZone zone="Zone" />;
      },
    },
    Slot: {
      fields: {
        Zone: { type: "slot" },
      },
      render: ({ Zone }) => {
        return <Zone />;
      },
    },
  },
}
```

2. Run the app in development mode.
3. Navigate to the editor and open devtools.
4. Drag and drop a `Slot` component into the editor canvas.

## What happens

The deprecation warning for `DropZone` is shown even though only a `slot` field is used.

## What I expect to happen

The deprecation warning should only be triggered when a `DropZone` is actually dragged into the canvas, not when using `slot` fields.

## Additional Media

https://github.com/user-attachments/assets/ddd08dbb-6070-43a7-ad04-6305da28aa35

### LINKED ISSUE #1029: Bug: Harder to Drag and Drop component in the middle of 2 dropzones in Canary 0.19
## Description

Hi, I got this issue with canary where it's harder to drag and drop with freshly dragged component.

Previously in 0.18, when a component is dragged, the page inside is moving automatically even if the component haven't reach the iframe yet.

But in 0.19, since dragging component outside the iframe no longer move the component inside, it is now harder to put component in the middle of 2 dropzones. (This also happen in the Puck canary demo 0.19)

Is this intentional? Could this change be reverted like the previous version where it's easier to drag in the middle of 2 dropzones? Thanks!

## Environment

Available in the Package.json below

[package.json.zip](https://github.com/user-attachments/files/20164798/package.json.zip)

## Steps to reproduce

1. Use the page.tsx provided
2. Try to drag outside iframe, it does not move the inside.

## Link to code example: 

[page.tsx.zip](https://github.com/user-attachments/files/20164840/page.tsx.zip)

## Recordings :
https://github.com/user-attachments/assets/7761412a-7a5f-4220-abdb-51f15890a8fe
https://github.com/user-attachments/assets/7fa0f76c-ccdf-4585-8416-49d3f943882d
https://github.com/user-attachments/assets/860080db-5a52-4ed7-b596-92038d68a028

chrisvxd:
Closes #1025, closes #1024, closes #1027, closes #1029

### DIFF
--- a/packages/core/lib/global-position.ts
+++ b/packages/core/lib/global-position.ts
@@ -14,7 +14,7 @@ export class GlobalPosition {
     this.target = target;
     this.original = original;
 
-    this.frameEl = document.querySelector("#preview-frame");
+    this.frameEl = document.querySelector("iframe#preview-frame");
 
     if (this.frameEl) {
       this.frameRect = this.frameEl.getBoundingClientRect();
--- a/package.json
+++ b/package.json
@@ -25,7 +25,7 @@
     "eslint-config-custom": "*",
     "lerna": "^7.1.1",
     "prettier": "^2.5.1",
-    "puppeteer": "^23.11.1",
+    "puppeteer": "^24.8.2",
     "standard-changelog": "^2.0.21",
     "turbo": "^2.3.3"
   },
--- a/packages/core/components/Drawer/styles.module.css
+++ b/packages/core/components/Drawer/styles.module.css
@@ -16,10 +16,7 @@
   bottom: 0;
   left: 0;
   pointer-events: none;
-}
-
-.Drawer-draggableFg {
-  z-index: 1;
+  z-index: -1;
 }
 
 .DrawerItem-draggable {
--- a/packages/core/components/DropZone/index.tsx
+++ b/packages/core/components/DropZone/index.tsx
@@ -294,12 +294,6 @@ export const DropZoneEdit = forwardRef<HTMLDivElement, DropZoneProps>(
     // Register and unregister zone on mount
     useEffect(() => {
       if (!zoneType || zoneType === "dropzone") {
-        if (zoneCompound !== rootDroppableId) {
-          console.warn(
-            "DropZones have been deprecated in favor of slot fields and will be removed in a future version of Puck. Please see the migration guide: https://www.puckeditor.com/docs/guides/migrations/dropzones-to-slots"
-          );
-        }
-
         if (ctx?.registerZone) {
           ctx?.registerZone(zoneCompound);
         }
@@ -309,6 +303,16 @@ export const DropZoneEdit = forwardRef<HTMLDivElement, DropZoneProps>(
             ctx?.unregisterZone(zoneCompound);
           }
         };
+      }
+    }, [zoneType]);
+
+    useEffect(() => {
+      if (zoneType === "dropzone") {
+        if (zoneCompound !== rootDroppableId) {
+          console.warn(
+            "DropZones have been deprecated in favor of slot fields and will be removed in a future version of Puck. Please see the migration guide: https://www.puckeditor.com/docs/guides/migrations/dropzones-to-slots"
+          );
+        }
       }
     }, [zoneType]);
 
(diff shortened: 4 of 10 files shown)


packages/core/lib/global-position.ts
```diff
@@ -14,7 +14,7 @@ export class GlobalPosition {
     this.target = target;
     this.original = original;
 
-    this.frameEl = document.querySelector("#preview-frame");
+    this.frameEl = document.querySelector("iframe#preview-frame");
```
chrisvxd, line 17: FYI @FedericoBonel this was breaking non-iframe use (first noticed in the multi-column layout docs), so I had to use a more specific selector.

FedericoBonel: Ah darn my bad! Thanks for checking it out and letting me know

Issue #1059, opened 2025-05-26
Bug: Draggable component hasn't its own z-index


SicParv1sMagna:
## Description

When I trying to render a table, which has sticky header and z-index: 5, header placed higher, then puck's select overlay, more details on screenshot

## Environment

- Puck version: 0.18.2
- Browser version: Chrome 114

## Steps to reproduce

1. Create a component with z-index higher then 2,
2. Render it in Puck
3. Click on him, to make it selected
4. You can see that parts where z-index is higher then 2 are placed higher then overlay

![Image](https://github.com/user-attachments/assets/cb71e026-cba1-4571-8911-74f7fefe377b)

FedericoBonel:
Hey @SicParv1sMagna! Thank you for the contribution 😁. I’ve already brought it up with the team, and they’ll review it as soon as they can.

chrisvxd:
@SicParv1sMagna Would you be able to provide a repo with reproduction CSS?

I'm unable to reproduce this. Unfortunately the fix in #1060 just increases the z-index to `100`, which isn't a long-term solution (someone could use `101`). We need to reset the stacking context, but without a working reproduction I can't easily identify where or validate a proper fix.

SicParv1sMagna:
I will add example how to reproduce soon

FedericoBonel:
Since we cannot reproduce this in 0.19.3, the fix PR was closed, and there have been no updates in over a month, I am going to close this as stale and non-reproducible.

PR #1060, opened 2025-05-26
feature: increase overlay z-index

SicParv1sMagna:
Solves issue: https://github.com/measuredco/puck/issues/1059


chrisvxd (commented):
Thanks for the contribution! Unfortunately this fix isn't suitable, please see reasons in #1059 comments.

PR #1077, opened 2025-06-04
chore: fix typo in react-router package.json

### LINKED ISSUE #1071: react-router recipe not working
## Description

When using the Puck recipes, you can't select the new `react-router` recipe. The script reports that there's no recipe with that name.

## Steps to reproduce

1. Run the `create-puck-app` script:

```bash
$ npx create-puck-app react-router-puck-app
```

2. Type `react-router` and press Enter:

```bash
? Which recipe would you like to use? react-router
```

## What happens

The script reports that there are no recipes named `react-router`.

## What I expect to happen

The new `react-router` recipe should be used to scaffold a new app with Puck preconfigured.

## Additional Media

Image

chrisvxd:
Closes #1071

### DIFF
--- a/packages/create-puck-app/templates/react-router/package.json.hbs
+++ b/packages/create-puck-app/templates/react-router/package.json.hbs
@@ -25,7 +25,7 @@
     "@types/react-dom": "^19.1.2",
     "typescript": "^5.8.3",
     "vite": "^6.3.3",
-    "vite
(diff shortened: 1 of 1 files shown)


matthewlynch (approved):
Whoops, copy paste fail. I did have this working so must have messed it up moving things around.

PR #1078, opened 2025-06-04
feat: enable resizing the sidebar

### LINKED ISSUE #969: Feat: Resizing of the sidebar width
## Description

The sidebars and preview area in the editor have fixed sizes, which can feel a bit cramped sometimes, especially with fields which have lots of text.

It'd be great if we could resize these panels easily to give the user more room where they need it.

## Proposal

Add draggable dividers between the panels so users can just slide them to resize. Libraries like react-resizable-panels might help here.

tlahmann:
closes #969 

This update adds resizable sidebars to the Puck editor. Users can now drag the borders between the sidebars and the canvas to adjust their width. These changes are saved and restored using localStorage.

Key Changes:

Added ResizeHandle component with mouse drag support

Updated layout grid to use dynamic column sizes

Why It Matters:

Improves flexibility and control over layout

Remembers user preferences across sessions

Maintains a clean and responsive UI


chrisvxd:
@tlahmann epic!

I'm going to pause before I review this as we're just wrapping up 0.19, and this is a significant change.

tlahmann:
Thank you for reviewing.

dj-fiorex:
this would be awesome!

tlahmann:
I just noticed the ResizeHandle is not positioned properly when the left or right sidebar has a scroll bar.  
I will rework the positioning.

tlahmann:
It is working well now

packages/core/components/Puck/components/ResizeHandle/index.tsx
```diff
@@ -0,0 +1,110 @@
...
+    const finalWidth = sidebarRef.current?.getBoundingClientRect().width || 0;
+    onResizeEnd(finalWidth);
+
+    // Trigger auto zoom by dispatching the viewportchange event
+    window.dispatchEvent(
+      new CustomEvent("viewportchange", {
+        bubbles: true,
+        cancelable: false,
+      })
+    );
```
chrisvxd: Would be better to explicitly trigger the auto zoom rather than faking the `viewportchange` event, which could have unintended side-effects.

See [resetAutoZoom](https://github.com/puckeditor/puck/blob/main/packages/core/components/Puck/components/Canvas/index.tsx#L83) for a reference implementation. This could be turned into a hook (`useResetAutoZoom`).

packages/core/components/Puck/components/Sidebar/styles.module.css
```diff
@@ -0,0 +1,35 @@
...
+  grid-area: left;
+}
+
+.Sidebar--right {
+  background: var(--puck-color-white);
+  border-inline-start: 1px solid var(--puck-color-grey-09);
+  grid-area: right;
+}
+
+.ResizeHandleContainer {
```
chrisvxd: Puck uses SUITCSS format, so this should be

```suggestion
.Sidebar-resizeHandle {
```

packages/core/components/Puck/components/Sidebar/styles.module.css
```diff
@@ -0,0 +1,35 @@
...
+  position: absolute;
+  height: 100%;
+  width: 6px;
+  z-index: 11;
+}
+
+.ResizeHandleContainer--left {
```
chrisvxd:
```suggestion
.Sidebar--left .Sidebar-resizeHandle {
```

packages/core/components/Puck/components/Sidebar/styles.module.css
```diff
@@ -0,0 +1,35 @@
...
+  grid-area: left;
+  justify-self: end;
+}
+
+.ResizeHandleContainer--right {
```
chrisvxd:
```suggestion
.Sidebar--right .Sidebar-resizeHandle {
```

packages/core/components/Puck/components/Sidebar/index.tsx
```diff
@@ -0,0 +1,55 @@
...
+      <div
+        ref={sidebarRef}
+        className={`${getClassName()} ${getClassName(`--${position}`)}`}
+      >
+        {children}
+      </div>
+      <div
+        className={`${getResizeHandleContainerClassName()} ${getResizeHandleContainerClassName(
+          `--${position}`
+        )}`}
```
chrisvxd:
```suggestion
        className={getClassName("resizeHandle")}
```

packages/core/components/Puck/components/Sidebar/styles.module.css
```diff
@@ -0,0 +1,35 @@
...
+  background: var(--puck-color-white);
+  border-inline-start: 1px solid var(--puck-color-grey-09);
+  grid-area: right;
+}
+
+.ResizeHandleContainer {
+  position: absolute;
+  height: 100%;
+  width: 6px;
+  z-index: 11;
```
chrisvxd: I'd prefer to avoid random z-index values as these can easily break. Can you reset the stacking context somehow, and set this to `1`?

packages/core/components/Puck/components/ResizeHandle/index.tsx
```diff
@@ -0,0 +1,110 @@
...
+
+  useEffect(() => {
+    document.addEventListener("mousemove", handleMouseMove);
+    document.addEventListener("mouseup", handleMouseUp);
+
+    return () => {
+      document.removeEventListener("mousemove", handleMouseMove);
+      document.removeEventListener("mouseup", handleMouseUp);
+    };
+  }, [handleMouseMove, handleMouseUp]);
```
chrisvxd: Can you only apply these handlers in the `mousedown` callback so they're not always active?

packages/core/components/Puck/components/ResizeHandle/index.tsx
```diff
@@ -0,0 +1,110 @@
...
+
+      const overlay = document.createElement("div");
+      overlay.id = "resize-overlay";
+      overlay.style.position = "fixed";
+      overlay.style.top = "0";
+      overlay.style.left = "0";
+      overlay.style.right = "0";
+      overlay.style.bottom = "0";
+      overlay.style.zIndex = "9999";
+      overlay.style.cursor = "col-resize";
```
chrisvxd: Ideally these styles would be captured in a data attribute and style.css file:

```css
[data-resize-overlay] {
  position: fixed;
  // etc
}

File: packages/core/types/AppState.tsx [hunk 9]
```diff
@@ -12,6 +12,8 @@ export type ArrayState = { items: ItemWithId[]; openId: string };
 export type UiState = {
   leftSideBarVisible: boolean;
   rightSideBarVisible: boolean;
+  leftSidebarWidth?: number | null;
+  rightSidebarWidth?: number | null;
```
chrisvxd: Could you document these?

Source: https://github.com/puckeditor/puck/blob/main/apps/docs/pages/docs/api-reference/data-model/app-state.mdx?plain=1

Example: https://puckeditor.com/docs/api-reference/data-model/app-state#uileftsidebarvisible

packages/core/components/Puck/components/ResizeHandle/index.tsx
tlahmann: I have reworked the event logic into the FrameContext

packages/core/components/Puck/components/Sidebar/styles.module.css
tlahmann: updated ✅

packages/core/components/Puck/components/ResizeHandle/index.tsx
tlahmann: done ✅

packages/core/types/AppState.tsx
tlahmann: I have added documentation to app-state.mdx.

packages/core/components/Puck/components/Sidebar/styles.module.css
tlahmann: the rules are not needed anymore

packages/core/components/Puck/components/ResizeHandle/index.tsx
tlahmann: I have updated the event listener logic

tlahmann:
Hello @chrisvxd,
Thanks for the review. I've updated the code.
I'd be glad to see this feature in a Puck release.

apps/docs/pages/docs/api-reference/data-model/app-state.mdx
```diff
@@ -87,6 +87,39 @@ Whether or not the left side bar is visible.
...
+| Prop          | Type                                  | Description                                                  |
+| ------------- | ------------------------------------- | ------------------------------------------------------------ |
+| `position`    | `"left" \| "right"`                   | Determines which side of the sidebar the handle appears on   |
+| `sidebarRef`  | `{ current: HTMLDivElement \| null }` | React ref to the sidebar element being resized               |
+| `onResize`    | `(width: number) => void`             | Callback function triggered during resize with the new width |
+| `onResizeEnd` | `(width: number) => void`             | Callback function triggered when resizing ends               |
+
+The component creates a draggable handle that updates the corresponding sidebar width in the AppState when dragged, and resets the canvas zoom when the resize operation completes.
+
+---
```
chrisvxd: This is not public so it's not necessary to document. We only need `rightSideBarWidth` and `leftSideBarWidth` added

apps/docs/pages/docs/api-reference/data-model/app-state.mdx
```diff
@@ -98,12 +131,6 @@ Puck does not currently provide the UI to control this value, but it can be togg
 
 ---
 
-### `ui.rightSideBarVisible`
-
-Whether or not the right side bar is visible.
-
----
```
chrisvxd: I understand you reordered these to keep them group, but these docs are alphabetical.

If you would rather not worry about docs, feel free to drop these changes and we can handle in post.

apps/docs/pages/docs/api-reference/data-model/app-state.mdx
```diff
@@ -87,6 +87,39 @@ Whether or not the left side bar is visible.
...
 
+### `ui.rightSideBarVisible`
+
+Whether or not the right side bar is visible.
+
+---
+
+### `ui.leftSideBarWidth`
+
+Width of the left side bar in pixels. This value can be dynamically adjusted by users through the `ResizeHandle` component, which provides a draggable interface for resizing the sidebar.
```
chrisvxd:
```suggestion
Current width of the left side bar in pixels.
```

ResizeHandle is not part of the public API.

apps/docs/pages/docs/api-reference/data-model/app-state.mdx
```diff
@@ -87,6 +87,39 @@ Whether or not the left side bar is visible.
...
+
+---
+
+### `ui.rightSideBarWidth`
+
+Width of the right side bar in pixels. Similar to the left sidebar, this width can be adjusted using the `ResizeHandle` component.
```
chrisvxd:
```suggestion
Current width of the right side bar in pixels.
```

ResizeHandle is not part of the public API.

apps/docs/pages/docs/api-reference/data-model/app-state.mdx
```diff
@@ -87,6 +87,39 @@ Whether or not the left side bar is visible.
...
+The `ResizeHandle` component provides a draggable interface for resizing sidebars in the Puck editor. It appears as a thin vertical handle on the edge of sidebars that users can drag to adjust the width.
+
+| Prop          | Type                                  | Description                                                  |
+| ------------- | ------------------------------------- | ------------------------------------------------------------ |
+| `position`    | `"left" \| "right"`                   | Determines which side of the sidebar the handle appears on   |
+| `sidebarRef`  | `{ current: HTMLDivElement \| null }` | React ref to the sidebar element being resized               |
+| `onResize`    | `(width: number) => void`             | Callback function triggered during resize with the new width |
+| `onResizeEnd` | `(width: number) => void`             | Callback function triggered when resizing ends               |
+
+The component creates a draggable handle that updates the corresponding sidebar width in the AppState when dragged, and resets the canvas zoom when the resize operation completes.
```
chrisvxd: This is not part of the public API.

```suggestion
```

packages/core/components/Puck/components/Sidebar/styles.module.css
```diff
@@ -0,0 +1,33 @@
...
+
+.Sidebar--left + .Sidebar-resizeHandle {
+  grid-area: left;
+  justify-self: end;
+}
+
+.Sidebar--right + .Sidebar-resizeHandle {
+  grid-area: right;
+  justify-self: start;
+}
```
chrisvxd, line 33: This is not technically SUITCSS format, but it works and I'm happy to refactor it in post, so don't worry about it.

If you're interested, to follow SUITCSS your HTML should have a wrapper like so

```html
<div class="Sidebar Sidebar--left">
  <div class="Sidebar-contents />
  <div class="Sidebar-resizeHandle" />
</div>
```

And your CSS selector would use a modifier descendant, rather than sibling selector

```css
.Sidebar--left Sidebar-resizeHandle {}
```

packages/core/components/Puck/index.tsx
```diff
@@ -395,12 +398,48 @@ function PuckLayout<
...
+                rightSidebarWidth: right || null,
+              },
+            });
+          }
+        }
+      } catch (error) {
+        console.error("Failed to load sidebar widths from localStorage", error);
+      }
+    }
+  }, []);
```
chrisvxd: I would like this to be inside the `useSidebarResize` hook too

packages/core/lib/frame-context.tsx
```diff
@@ -0,0 +1,51 @@
+import React, {
```
chrisvxd, line 3: I think this file is causing server components to fail on build: https://github.com/puckeditor/puck/actions/runs/16220036922/job/45798124386?pr=1078

We might need to specify use-client.

```suggestion
"use client";

import React, {
```

packages/core/components/Puck/index.tsx
```diff
@@ -395,12 +398,48 @@ function PuckLayout<
...
+    sidebarRef: leftSidebarRef,
+    handleResizeEnd: handleLeftSidebarResizeEnd,
+  } = useSidebarResize("left", dispatch);
+
+  const {
+    width: rightWidth,
+    setWidth: setRightWidth,
+    sidebarRef: rightSidebarRef,
+    handleResizeEnd: handleRightSidebarResizeEnd,
+  } = useSidebarResize("right", dispatch);
```
chrisvxd, line 418: These hooks should probably be called within the Sidebar component, along with the refs. There is no reason to keep this state in the main Puck component.

packages/core/lib/frame-context.tsx
```diff
@@ -0,0 +1,51 @@
...
+  RefObject,
+  useMemo,
+  useCallback,
+} from "react";
+import { useResetAutoZoom } from "./use-reset-auto-zoom";
+
+interface FrameContextType {
+  frameRef: RefObject<HTMLDivElement | null>;
+  resetAutoZoom: ReturnType<typeof useResetAutoZoom>;
+}
```
chrisvxd: I think this context can be simplified to simply `frameRef`. This could be useful in other places throughout the app.

Bundling `resetAutoZoom` in this context feels duplicative, and it can just as easily be called inline with `frameRef` as needed.

packages/core/components/Puck/components/ResizeHandle/index.tsx
```diff
@@ -0,0 +1,99 @@
...
+    (e: MouseEvent) => {
+      if (!isDragging.current) return;
+
+      const delta = e.clientX - startX.current;
+      const newWidth =
+        position === "left"
+          ? startWidth.current + delta
+          : startWidth.current - delta;
+
+      const width = Math.max(186, newWidth);
```
chrisvxd: Puck uses multiples of 4

```suggestion
      const width = Math.max(192, newWidth);
```

chrisvxd (changes requested):
Some further comments! Can't test due to build issues.

apps/docs/pages/docs/api-reference/data-model/app-state.mdx
tlahmann: removed ✅

apps/docs/pages/docs/api-reference/data-model/app-state.mdx
tlahmann: updated ✅

apps/docs/pages/docs/api-reference/data-model/app-state.mdx
tlahmann: updated ✅

apps/docs/pages/docs/api-reference/data-model/app-state.mdx
tlahmann: removed ✅

packages/core/lib/frame-context.tsx
tlahmann: I think you're right

packages/core/components/Puck/components/ResizeHandle/index.tsx
tlahmann: Updated ✅
I noticed some old values inside packages/core/components/Puck/styles.module.css (rows 51 and 81) that might also need an update

packages/core/components/Puck/index.tsx
tlahmann: moved ✅

apps/docs/pages/docs/api-reference/data-model/app-state.mdx
tlahmann: I have reverted the order to alphabetical

packages/core/components/Puck/components/Sidebar/styles.module.css
tlahmann: Aah, thanks. I see my issue now.

tlahmann: will leave it for now

packages/core/lib/frame-context.tsx
tlahmann: removed the duplicate reference to the resetAutoZoom

packages/core/components/Puck/index.tsx
tlahmann: The main issue I have with moving the hooks is styling the left and right sidebars. Managing the width gets messy.

Do you have any suggestions for styling them?
I'm mainly concerned about this ternary: ```leftWidth ? `${leftWidth}px` : "var(--puck-side-bar-width)"```

tlahmann:
Thanks again for reviewing. I left the sidebar resize hooks in the Puck component’s index.tsx. I need help with one last issue.

When I move the width and event listener into the useSidebarResize hook, the sidebar doesn’t update visually while dragging. The onAfterDragging event also keeps storing the same static width. I need a way to store and update the width during the drag.

packages/core/components/Puck/components/Canvas/index.tsx
```diff
@@ -80,26 +90,25 @@ export const Canvas = () => {
...
-    [frameRef, zoomConfig, viewports]
-  );
-
   // Auto zoom
   useEffect(() => {
-    setShowTransition(false);
-    resetAutoZoom(viewports);
-  }, [frameRef, leftSideBarVisible, rightSideBarVisible]);
+    if (!isResettingZoomRef.current) {
+      isResettingZoomRef.current = true;
```
chrisvxd: I don't quite understand what this code is doing and, if it's realy necessary, why it couldn't be moved within `useResetAutoZoom` rather than being repeated throughout this component

packages/core/components/Puck/index.tsx
chrisvxd: Aha I see. This is fine for now.

chrisvxd (changes requested):
Nearly there, thanks for the continued tweaks! There's still one small piece of code I don't understand that I think might be good to adjust before merge.

packages/core/components/Puck/components/Canvas/index.tsx
chrisvxd: @tlahmann please let me know about this one! I'd like to get this PR merged this week for our 0.20 release

packages/core/styles.css
```diff
@@ -12,3 +12,13 @@
...
+
+[data-resize-overlay] {
+  position: fixed;
+  top: 0;
+  left: 0;
+  right: 0;
+  bottom: 0;
+  z-index: 9999;
+  cursor: col-resize;
+}
```
chrisvxd: This file has moved so now facing a conflict. Can you move to a `styles.css` file under the component it's used and import it into the component like `import "./styles.css"`?

tlahmann: I have moved the styles and imported the css directly inside the Canvas/index.tsx

packages/core/components/Puck/components/Canvas/index.tsx
tlahmann: Hey @chrisvxd. you're absolutely right, the hook can also handle the reference and other options. I have reworked that.
While working on this i had minor issues with the reference object preventing the reset calling itself

tlahmann:
stylus has been unpublished: https://github.com/stylus/stylus/issues/2938
will hopefully be soon resolved

tlahmann:
thank you for the reviews ❤️

chrisvxd (changes requested):
@tlahmann woo looks good! Just needs a format. Please run `npm run format` and commit the results.

chrisvxd (approved):

chrisvxd:
Just going to format separately.

tlahmann:
thank you for your support

PR #1117, opened 2025-06-26
perf: optimize matchedComponents handling using Set for better performance and remove unnecessary type conversions.

ilxqx:
optimize matchedComponents handling using Set for better performance and remove unnecessary type conversions.


chrisvxd:
Thanks for the contribution @ilxqx, but these refactors seem somewhat redundant and unnecessary. If you have performance benchmarks and a valid reason for this change, please share and I'll reopen.

Issue #668, opened 2024-10-24
Feature Request: Adding up/down and add buttons in Action Bar


OsamuBlack:
These features can Puck Actionbar which can be **implemented using the override prop** but it would be better if they are included **by default**.

# 1. Having the ability to move your component up and down using arrows is usually provided with most drag n drop editors.

For example:
## Shopify Actionbar
![Shopify Actionbar](https://github.com/user-attachments/assets/a4406760-ed12-4d44-9a73-fcf4d07ea89c)
## WordPress Actionbbar
![WordPress ActionBar](https://github.com/user-attachments/assets/c182822a-4f1f-4346-b4ab-28475a15174a)

## Puck Implementation
I have implemented it using the override:
![Puck Overrided Actionbar](https://github.com/user-attachments/assets/ab0c4d0a-3a77-49e7-a70f-cf28d6d2c3be)

```tsx
import { rootDroppableId } from "@/puck/lib/root-droppable-id";
import { IconButton, usePuck } from "@measured/puck";
import { ReactNode, memo } from "react";
import { LuChevronDown, LuChevronUp } from "react-icons/lu";
import { ActionBar as PuckActionBar } from "@measured/puck";

export default function ActionBar({
  label,
  children,
}: {
  label?: string;
  children: ReactNode;
}) {
  const {
    appState: {
      ui: { itemSelector },
      data,
    },
    dispatch,
  } = usePuck();

  if (!itemSelector) return null;

  const upDisabled = itemSelector?.index === 0;
  const downDisabled =
    (itemSelector.zone && itemSelector?.zone !== rootDroppableId && data.zones
      ? data.zones[itemSelector.zone]?.length
      : data.content.length) ===
    itemSelector.index + 1; 

  return (
    <PuckActionBar label={label}>
      <IconButton
        onClick={() => {
          dispatch({
            type: "reorder",
            sourceIndex: itemSelector.index,
            destinationIndex: itemSelector.index - 1,
            destinationZone: itemSelector.zone || rootDroppableId,
          });
          dispatch({
            type: "setUi",
            ui: (ui) => ({
              ...ui,
              itemSelector: { ...itemSelector, index: itemSelector.index - 1 },
            }),
          });
        }}
        title="Move Up"
        disabled={upDisabled}
      >
        <LuChevronUp />
      </IconButton>
      <IconButton
        onClick={() => {
          dispatch({
            type: "reorder",
            sourceIndex: itemSelector.index,
[...]

PR #1119, opened 2025-06-29
feat: add move up and move down actions to DraggableComponent

OsamuBlack:
feat: implement getZoneContentIds utility function

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,19 @@
 
 <!--__CHANGELOG_ENTRY__-->
 
+## [0.20.2](https://github.com/measuredco/puck/compare/v0.20.1...v0.20.2) (2025-09-05)
+
+
+### Bug Fixes
+
+* add CommonJS export to emotion cache package.json ([fbf3bd1](https://github.com/measuredco/puck/commit/fbf3bd1038a4c39d4e0c5079c3af33dd515c590b))
+* bypass resolveData cache when trigger is force ([eea53e0](https://github.com/measuredco/puck/commit/eea53e0464c15e7c2f964da973a16a04160224b8))
+* preserve selection when switching viewports and using zoom select ([38514a1](https://github.com/measuredco/puck/commit/38514a169592a3ffbaaedcfcb5b92767ba07b0e7))
+* prevent hidden category components from showing in 'Other' ([8333eaf](https://github.com/measuredco/puck/commit/8333eaf7e27d3225fb7133fe14326e1f8e93776a))
+
+
+
+
 ## [0.20.1](https://github.com/measuredco/puck/compare/v0.20.0...v0.20.1) (2025-08-19)
 
 
--- a/apps/docs/package.json
+++ b/apps/docs/package.json
@@ -1,6 +1,6 @@
 {
   "name": "docs",
-  "version": "0.20.1",
+  "version": "0.20.2",
   "private": true,
   "scripts": {
     "dev": "next dev",
--- a/lerna.json
+++ b/lerna.json
@@ -7,6 +7,6 @@
     "packages/plugin-emotion-cache",
     "packages/plugin-heading-analyzer"
   ],
-  "version": "0.20.1",
+  "version": "0.20.2",
   "npmClient": "yarn"
 }
--- a/package.json
+++ b/package.json
@@ -36,7 +36,7 @@
     "recipes/*",
     "packages/*"
   ],
-  "version": "0.20.1",
+  "version": "0.20.2",
   "engines": {
     "node": ">=20.0.0"
   }
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@measured/puck",
-  "version": "0.20.1",
+  "version": "0.20.2",
   "description": "The open-source visual editor for React",
   "author": "Chris Villa <chris@puckeditor.com>",
   "repository": "measuredco/puck",
--- a/packages/create-puck-app/package.json
+++ b/packages/create-puck-app/package.json
@@ -1,6 +1,6 @@
 {
   "name": "create-puck-app",
-  "version": "0.20.1",
+  "version": "0.20.2",
   "author": "Chris Villa <chris@puckeditor.com>",
   "repository": "measuredco/puck",
   "bugs": "https://github.com/measuredco/puck/issues",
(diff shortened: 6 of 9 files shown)


FedericoBonel:
This PR is attempting to close #668

chrisvxd:
Hey @OsamuBlack! Would you mind reopening this PR? We had to change some deployment settings which means I can't trigger the preview build.

packages/core/components/DraggableComponent/index.tsx
```diff
@@ -14,7 +14,13 @@ import {
 import styles from "./styles.module.css";
 import "./styles.css";
 import getClassNameFactory from "../../lib/get-class-name-factory";
-import { Copy, CornerLeftUp, Trash } from "lucide-react";
+import {
+  ChevronDown,
+  ChevronUp,
```
chrisvxd: I think [`MoveDown`](https://lucide.dev/icons/move-down) and [`MoveUp`](https://lucide.dev/icons/move-up) would be better for this

packages/core/components/DraggableComponent/index.tsx
```diff
@@ -144,6 +161,10 @@ export const DraggableComponent = ({
     [setLocalZones]
   );
 
+  const zoneContentIds = useAppStore(
+    useShallow((s) => getZoneContentIds(zoneCompound, s.state))
+  );
```
chrisvxd: This will cause every item in the drop zone to re-render whenever any item is moved. To avoid this, you should add an `isFirst` or `isLast` const that checks whether the current item is the first or last item in the zone.

```
const isFirst = useAppStore(s => getZoneContentIds(zoneCompound, s.state)[0] === componentId);
// ...
```

packages/core/components/DraggableComponent/index.tsx
```diff
@@ -555,6 +625,26 @@ export const DraggableComponent = ({
     [ctx?.areaId]
   );
 
+  const moveUpAction = useMemo(
+    () =>
+      Boolean(index) && (
```
chrisvxd: I think it would be better to keep these visible but disable them

packages/core/components/DraggableComponent/index.tsx
```diff
@@ -555,6 +625,26 @@ export const DraggableComponent = ({
...
+        <ActionBar.Action onClick={onMoveUp} label="Select Previous">
+          <ChevronUp size={16} />
+        </ActionBar.Action>
+      ),
+    [index]
+  );
+
+  const moveDownAction = useMemo(
+    () =>
+      index < (zoneContentIds?.length || 0) - 1 && (
```
chrisvxd: As above, I think better to keep visible and disable

packages/core/components/DraggableComponent/index.tsx
```diff
@@ -375,6 +396,55 @@ export const DraggableComponent = ({
...
+      destinationIndex: index - 1,
+      destinationZone: zoneCompound,
+    });
+
+    dispatch({
+      type: "setUi",
+      ui: {
+        itemSelector: {
+          zone: zoneCompound,
+          index: index - 1,
```
chrisvxd: This one doesn't seem to be updating the currently selected item on moveUp for me. It works on moveDown.

chrisvxd (changes requested):
👏 Great feature! A few comments.

packages/core/components/DraggableComponent/index.tsx
OsamuBlack: I create a new pull request with the changes?

packages/core/components/DraggableComponent/index.tsx
OsamuBlack: Done, my thoughts exactly but for that we would need to add a prop to action button to disable and I wished to follow the original implementation as with the parent action. 

But you are right that they should be disabled not removed from view.

packages/core/components/DraggableComponent/index.tsx
chrisvxd: It's fine to push to this PR

packages/core/components/DraggableComponent/index.tsx
chrisvxd: I don't see the changes!

packages/core/components/DraggableComponent/index.tsx
OsamuBlack: I believe we only need to check for `isLast`.

The first item is easy to check based on `index == 0`

packages/core/components/DraggableComponent/index.tsx
OsamuBlack: The move up is also working but the selection is not re-rendered. I believe the issue might be with some other component or we need to setup a state to trigger render in the `DraggableComponent.tsx`.


https://github.com/user-attachments/assets/66ca7ec4-ed32-4042-8649-6bdc24010975

OsamuBlack (commented):
- Added disable styles for action bar button.
- Used Move up and Move Down icons for sorting
- Replaced `zoneIds` with `isLast` for better performance.

Need help with the move up selection re-render issue.

packages/core/components/DraggableComponent/index.tsx
FedericoBonel: @chrisvxd @OsamuBlack this could be related to this: https://github.com/puckeditor/puck/issues/955

OsamuBlack: Calling `setUi` for up action reflects changes in the state. When we hover over the element, it triggers a re-render for selection component, updating its position. 

I believe it is a *Re-rendering* issue, in the selection rendering component. The state is not updating when new index < previous index.

FedericoBonel:
Hey @OsamuBlack!  

This is coming along great. I did notice that the up and down actions allow you to bypass permissions, since they remain available even when using `drag: false` in the permissions API.  

For example, if I open the demo Button component (`apps/demo/config/blocks/Button/index.tsx`) and add `permissions: { drag: false }` in the component config, I can still see the up and down actions in Button components, even though I’m not able to drag them around.  

https://github.com/user-attachments/assets/06efa0ab-4ef9-4e4e-864f-b8d21fd97f39

packages/core/components/DraggableComponent/index.tsx
```diff
@@ -555,6 +620,28 @@ export const DraggableComponent = ({
     [ctx?.areaId]
   );
 
+  const moveUpAction = useMemo(
+    () => (
+      <ActionBar.Action onClick={onMoveUp} label="Move Down" disabled={!index}>
```
FedericoBonel: Could you change the label here to **"Move Up"**?

OsamuBlack: On it,

PR #1127, opened 2025-07-08
fix: ensure setHistories resets the appState

### LINKED ISSUE #1101: setHistories not longer updates the editor data
## Description

In Puck `0.19.1`, using `setHistories` no longer updates the internal Puck data. This causes the current histories index to become detached from the actual data in use within the editor. As a result, any changes made after calling `setHistories` will inject the current data (that is no longer in the histories) into the histories stack back unless a `setData` action is also dispatched.

This behavior differs from version `0.18`, where `setHistories` implicitly updated the editor data.

## Environment

- Puck version: 0.19.1

## Steps to reproduce

1. Render the `<Puck />` component with a UI override

 that calls `setHistories` to clear the history:

```tsx
const Editor = () => {
  return (
    <Puck
      config={config}
      data={data}
      overrides={{
        headerActions: ({ children }) => {
          const getPuck = useGetPuck();

          const handleClick = useCallback(() => {
            const {
              history: { setHistories, index, histories },
            } = getPuck();

            setHistories([
              { id: "root", state: { data: histories[0].state.data } },
            ]);
          }, []);

          return (
            <>
              <button onClick={handleClick}>Reset editor</button>
              {children}
            </>
          );
        },
      }}
    />
  );
};
```

2. Navigate to the editor.
3. Open your browser's devtools.
4. Click the "Reset editor" button (or similar) that triggers `setHistories`.

## What Happens

The editor data does not update to reflect the current histories index after resetting.

## What I Expect to Happen

The editor data should update along with the histories when calling `setHistories`, maintaining consistency with the selected index.

## Additional Media

https://github.com/user-attachments/assets/43f5e6bf-94f4-4b83-9b3b-969a65292f10

chrisvxd:
Puck was resetting to the old histories.

Closes #1101

### DIFF
--- a/packages/core/store/slices/__tests__/history.spec.tsx
+++ b/packages/core/store/slices/__tests__/history.spec.tsx
@@ -200,7 +200,7 @@ describe("history slice", () => {
       expect(index).toBe(1);
       expect(appStore.getState().dispatch).toHaveBeenCalledWith({
         type: "set",
-        state: defaultAppState, // from the old store’s last item or initialAppState
+        state: { ...defaultAppState, data: "Two" },
       });
     });
   });
--- a/packages/core/store/slices/history.ts
+++ b/packages/core/store/slices/history.ts
@@ -128,8 +128,7 @@ export const createHistorySlice = (
       dispatch({
         type: "set",
         state:
-          history.histories[history.histories.length - 1]?.state ||
-          history.initialAppState,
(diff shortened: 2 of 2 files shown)


mkilpatrick:
I think you have the wrong issue tagged. Should it be this one? https://github.com/puckeditor/puck/issues/1101

chrisvxd:
@mkilpatrick good catch!

Issue #771, opened 2025-01-07
Issue with Array Field Restriction in Puck Editor


Ashu912:
In version 0.17.1 by using resolveData we are unable to acces readOnly in array fields and inside that  we are using objects.

For example:

When a user has restricted access to an array field, the array should not not editable for its contents. However, the inner objects of the array are still editable , bypassing the restriction.
This behavior compromises the data restriction control and doesn't align with the intended permission logic.

Could you please look into this or let us know if additional configurations are required to apply restrictions to both the array and its inner objects?

chrisvxd:
Hi @Ashu912 - thanks for reporting. Can you share steps to reproduce or a code example?

Ashu912:
Hi @chrisvxd - thanks for responding,
I have used this structure of field and in resolveData what will be the format off passing readOnly to the array of objects.
we are passing readOnly to this in json so are we missing something I have gone throught the docs of the latest version we won't find any thing how to pass readOnly to array please do let us know are we have do some changes?

readonly Json------>
{"Heading.headingText":true,"Heading.paddingY":true,"Heading.color":true,"Heading.textAlign":true,"Heading.fontSize":true,"socialDescriptionText":true,"socialMediaNumber.number":true,"socialMediaNumber.color":true,"SocialMediaButtons":true,"SocialMediaButtons.0.mediaIcons.title":true,"SocialMediaButtons.0.mediaIcons.socialMediaLink":true,"SocialMediaButtons.0.mediaIcons.socialMediaIcon":true,"SocialMediaButtons.1.mediaIcons.title":true,"SocialMediaButtons.1.mediaIcons.socialMediaLink":true,"SocialMediaButtons.1.mediaIcons.socialMediaIcon":true,"SocialMediaButtons.2.mediaIcons.title":true,"SocialMediaButtons.2.mediaIcons.socialMediaLink":true,"SocialMediaButtons.2.mediaIcons.socialMediaIcon":true,"SocialMediaButtons.3.mediaIcons.title":true,"SocialMediaButtons.3.mediaIcons.socialMediaLink":true,"SocialMediaButtons.3.mediaIcons.socialMediaIcon":true,"maxWidth":true,"margin":true}

fields--------->
const fields_data: Fields<Props> = {
  id: {
    type: "custom",
    label: "Id",
    render: ({ id, value, onChange }: any) => (
      <p className="text-xs">
        <b>Id</b> <br />
        {value}
      </p>
    ),
  },
  Heading: {
[...]

PR #1131, opened 2025-07-09
1130 field name propagation

shannonhochkins:
This PR introduces a fix for this [issue](https://github.com/puckeditor/puck/issues/1130), where the name fields which are not being populated correctly for custom fields.

The puck demo app also has this problem in the `Hero` component, which renders a custom field, and within that an image url field, see the attached screenshot where the `name` of the field is rendered in `main`, and in this feature branch

The `image.url` value is simply undefined in main, so it doesn't render at all, this change passes through the correct name context, and uses it within the AutoField component, and additionally renders custom fields within the context provider so custom fields get the name value as expected.

Before (in main):

> Notice - the `image.url` name is not rendered at all, because the `name` field value is undefined.

Screenshot 2025-07-09 at 12 20 31 PM

After (in feature branch)

> Notice the `image.url` value now coming through as expected.

Screenshot 2025-07-09 at 12 21 01 PM


packages/core/components/AutoField/fields/ObjectField/index.tsx
```diff
@@ -37,12 +37,11 @@ export const ObjectField = ({
           {Object.keys(field.objectFields!).map((subName) => {
             const subField = field.objectFields![subName];
 
-            const subPath = `${name}.${subName}`;
-            const localSubPath = `${localName || name}.${subName}`;
+            const subPath = `${localName}.${subName}`;
```
shannonhochkins, line 41: I'm not sure if there was an intentional decision to use `localSubPath` and `subPath` here, but I have a feeling this was a mistake, without this change, `name` is undefined so the `subPath` value === `undefined.subfieldName`

packages/core/components/AutoField/index.tsx
```diff
@@ -216,31 +216,24 @@ function AutoFieldInternal<
...
-
-  const Render = render[field.type] as (props: FieldProps) => ReactElement;
-
   return (
     <NestedFieldContext.Provider
       value={{
         readOnlyFields: nestedFieldContext.readOnlyFields || readOnly || {},
-        localName: nestedFieldContext.localName,
+        localName: nestedFieldContext.localName ?? mergedProps.name,
       }}
```
shannonhochkins, line 237: If the `nestedFieldContext.localName` value is undefined, provide it the field name by default, this is needed so AutoFields within a `custom` field can receive the parent name correctly.

packages/core/components/AutoField/index.tsx
```diff
@@ -216,31 +216,24 @@ function AutoFieldInternal<
...
-        <div className={getClassName()}>
-          <CustomField {...mergedProps} />
-        </div>
-      </div>
-    );
+    FieldComponent = field.render as any;
+  } else {
+    children = defaultFields[field.type](mergedProps);
+    FieldComponent = render[field.type] as (props: FieldProps) => ReactElement;
   }
```
shannonhochkins, line 231: This is a small refactor to switch the FieldComponent to use the `custom` component or the default fields within the NestedFieldContext.Provider

chrisvxd (approved):
I tested this and checked for regressions, particularly around #771 (the PR for which introduced the provider). This seems to do the trick! Thanks!

Issue #1145, opened 2025-07-09
Tidy up types


chrisvxd:
## Description

The types currently have various issues:

1. Some types are exposed that shouldn't be
2. Generics are hard to extend, as they are ordered

The following tidy ups should be made:

- [ ] ~~Replace all generics with a single, extensible generic (i.e. `<Props, RootProps, DataShape>` becomes `<{ Props, RootProps, DataShape }>`~~ (Done in #1147)
- [ ] Unwanted types that should never have been public
- [ ] Collapse duplicate types that are made simpler by simplified generics

## Considerations

- This will be a breaking change

## Proposals

TBD

PR #1147, opened 2025-07-09
fix: support user-defined fields in overrides

### LINKED ISSUE #729: Custom field type overrides result in type error
### Discussed in https://github.com/measuredco/puck/discussions/702

<div type='discussions-op-text'>

<sup>Originally posted by **icedcola** November 14, 2024</sup>
Hello. I'm having some trouble defining new field types with the overrides. Just wondering if I am doing something silly, or if the overrides aren't meant for this.

I have an override like so (just to test):
```ts
const overrides = {
    fieldTypes: {
      slider: ( { name, onChange, value }) => (<div>Slide!</div>),
    },
```

and using it in a component:
```ts
fields: {
        Sliderbar: {label: "Test", type: "slider"}
        },
```

gives an `Uncaught TypeError: defaultFields[field.type] is not a function at AutoFieldInternal` error.

The Docs mention "You can specify a custom render method for each known [field type](https://puckeditor.com/docs/api-reference/fields), **or introduce completely new ones.**"

I replaced with a `text` or `number` override and that seems to work. Assuming it is only for overriding default field types, but hopefully I'm just messing something up :) Thank you for any help.

Best regards,
Matt</div>

chrisvxd:
Closes #729, addressing support for user-defined fields in `overrides` and `fieldTransforms` (#1191). Also types plugins using the same mechanism.

This introduces generic types for `Config` (originally part of #1145) enabling typing via an object:

```ts
type UserConfig = Config<{
  components: { Hero: { background: string } },
  fields: { // Define new fields
    color: {
      type: "color",
      mode: "hex" | "rgb"
    }
  }
}>;
```

And for `ComponentConfig`:

```ts
type HeroConfig = Config<{
  props: { background: string },
  fields: { // Define new fields
    color: {
      type: "color",
      mode: "hex" | "rgb"
    }
  }
}>;
```

It is not possible to introduce automatic typing for field transforms, and the user must do this themselves. For inline text fields, the user must set the node to `string | ReactNode`.

### DIFF
--- a/apps/demo/app/[...puckPath]/client.tsx
+++ b/apps/demo/app/[...puckPath]/client.tsx
@@ -1,10 +1,11 @@
 "use client";
 
-import { Button, Puck, Render } from "@/core";
+import { AutoField, Button, FieldLabel, Puck, Render } from "@/core";
 import headingAnalyzer from "@/plugin-heading-analyzer/src/HeadingAnalyzer";
 import config from "../../config";
 import { useDemoData } from "../../lib/use-demo-data";
 import { useEffect, useState } from "react";
+import { Type } from "lucide-react";
 
 export function Client({ path, isEdit }: { path: string; isEdit: boolean }) {
   const metadata = {
@@ -42,6 +43,22 @@ export function Client({ path, isEdit }: { path: string; isEdit: boolean }) {
             enabled: params.get("disableIframe") === "true" ? false : true,
           }}
           overrides={{
+            fieldTypes: {
+              // Example of user field provided via overrides
+              userField: ({ readOnly, field, name, value, onChange }) => (
+                <FieldLabel
+                  label={field.label || name}
+                  readOnly={readOnly}
+                  icon={<Type size={16} />}
+                >
+                  <AutoField
+                    field={{ type: "text" }}
+                    onChange={onChange}
+                    value={value}
+                  />
+                </FieldLabel>
+              ),
+            },
             headerActions: ({ children }) => (
               <>
                 <div>
--- a/apps/demo/app/[...puckPath]/client.tsx
+++ b/apps/demo/app/[...puckPath]/client.tsx
@@ -42,6 +42,9 @@ export function Client({ path, isEdit }: { path: string; isEdit: boolean }) {
           iframe={{
             enabled: params.get("disableIframe") === "true" ? false : true,
           }}
+          fieldTransforms={{
+            userField: ({ value }) => value, // Included to check types
+          }}
           overrides={{
             fieldTypes: {
               // Example of user field provided via overrides
--- a/apps/demo/app/rsc/page.tsx
+++ b/apps/demo/app/rsc/page.tsx
@@ -1,7 +1,7 @@
 import { Metadata } from "next";
 import config from "../../config/server";
 import { initialData } from "../../config/initial-data";
-import { Props, RootProps } from "../../config/types";
+import { Components, RootProps } from "../../config/types";
 
 import { Config } from "@/core";
 import { Render, resolveAllData } from "@/core/bundle/rsc";
@@ -22,7 +22,7 @@ export default async function Page() {
     example: "Hello, world",
   };
 
-  const resolvedData = await resolveAllData<Props, RootProps>(
+  const resolvedData = await resolveAllData<Components, RootProps>(
     data,
     conf,
     metadata
--- a/apps/demo/config/blocks/Heading/index.tsx
+++ b/apps/demo/config/blocks/Heading/index.tsx
@@ -1,6 +1,6 @@
 import React from "react";
 
-import { ComponentConfig } from "@/core";
+import { ComponentConfig } from "@/core/types";
 import { Heading as _Heading } from "@/core/components/Heading";
 import type { HeadingProps as _HeadingProps } from "@/core/components/Heading";
 import { Section } from "../../components/Section";
--- a/apps/demo/config/blocks/Hero/client.tsx
+++ b/apps/demo/config/blocks/Hero/client.tsx
@@ -1,13 +1,20 @@
 /* eslint-disable @next/next/no-img-element */
 import React from "react";
-import { ComponentConfigWithExtensions } from "@/core/types";
+import { ComponentConfig } from "@/core/types";
 import { quotes } from "./quotes";
 import { AutoField, FieldLabel } from "@/core";
 import { Link2 } from "lucide-react";
 import HeroComponent, { HeroProps } from "./Hero";
-import { Extensions } from "../../types";
 
-export const Hero: ComponentConfigWithExtensions<Extensions, HeroProps> = {
+export const Hero: ComponentConfig<{
+  props: HeroProps;
+  fields: {
+    userField: {
+      type: "userField";
+      option: boolean;
+    };
+  };
+}> = {
   fields: {
     quote: {
       type: "external",
--- a/apps/demo/config/blocks/Hero/client.tsx
+++ b/apps/demo/config/blocks/Hero/client.tsx
@@ -1,12 +1,13 @@
 /* eslint-disable @next/next/no-img-element */
 import React from "react";
-import { ComponentConfig } from "@/core/types";
+import { ComponentConfigWithExtensions } from "@/core/types";
 import { quotes } from "./quotes";
 import { AutoField, FieldLabel } from "@/core";
 import { Link2 } from "lucide-react";
 import HeroComponent, { HeroProps } from "./Hero";
+import { Extensions } from "../../types";
 
-export const Hero: ComponentConfig<HeroProps> = {
+export const Hero: ComponentConfigWithExtensions<Extensions, HeroProps> = {
   fields: {
     quote: {
       type: "external",
@@ -133,7 +134,7 @@ export const Hero: ComponentConfig<HeroProps> = {
         },
       },
     },
-    padding: { type: "text" },
+    padding: { type: "userField", option: true },
   },
   defaultProps: {
     title: "Hero",
--- a/apps/demo/config/blocks/Template/client.tsx
+++ b/apps/demo/config/blocks/Template/client.tsx
@@ -5,15 +5,15 @@ import { ComponentConfig, ComponentDataOptionalId, Slot } from "@/core/types";
 import { withLayout } from "../../components/Layout";
 import { generateId } from "@/core/lib/generate-id";
 import { componentKey } from "../../index";
-import { type Props } from "../../types";
+import { type Components } from "../../types";
 import TemplateComponent, { TemplateProps } from "./Template";
 
 const usePuck = createUsePuck();
 
-async function createComponent<T extends keyof Props>(
+async function createComponent<T extends keyof Components>(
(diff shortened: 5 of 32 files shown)


matthewlynch:
1. I'd say feature, docs were incorrect if the feature was never supported.
2. Can you expand on this point / give an example?

chrisvxd:
On (2), currently the implementation requires that the user will need to let TS know that `{ type: "color" }` is a valid type via the new `ConfigWithExtensions` and `ComponentConfigWithExtensions` types.

```ts
const config: ConfigWithExtensions<
  { Field: { type: "color", mode: "hex" | "rgb" } }
  { Hero: { background: string } }
> = {
  components: {
    Hero: {
      fields: {
        background: { type: "color", mode: "hex" }
      }
    }
  }
}
```

This could be wrapped via a generic type for convenience.

```ts
const config: ConfigWithColorField<{ Hero: { background: string } }> = {
  components: {
    Hero: {
      fields: {
        background: { type: "color" }
      }
    }
  }
}
```

However, this additional typing could be a little cumbersome, especially if loading via a plugin.

If we support TS string values, the user will be able to define the above in their fields config with no further changes.

```ts
const config: Config<{ Hero: { background: string } }> = {
  components: {
    Hero: {
      fields: {
        background: { type: "color" }
      }
    }
  }
}
```

This could be convenient, but means that the user can enter any string into `type` without a build error. It also kills TS autocompletion.

FedericoBonel:
> 1. Should this be a bug, or a feature? This is currently documented, so the lack of the feature is a bug. Although WithExtensions types imply feature, we don't typically do feature releases for types.

I think it should be considered a bug fix. This was supposed to work and was already documented, so the feature effectively existed. The expected behavior would have been that you could define your own custom fields via the overrides, but that wasn’t functioning correctly.

> 2. Should this support untyped string type values? This might make extension easier (via plugins), but breaks type-safety and autocomplete when defining fields.

It would be great to maintain type-safety and autocomplete; otherwise, it feels like using `any`, in my opinion.

> 3. Types are undocumented

We should document them, but this is an issue that applies to most types since many are currently undocumented and can be confusing. I think we should probably finish #1145 before fully documenting each one though, unless it’s very clear they won’t change after the rework.

FedericoBonel (approved):
I tested the API a bit. It’s a little tedious to provide the Extensions, but it’s manageable if you centralize the type export with the extensions already added to it.

chrisvxd:
Holding this as suspect we might need to add some docs

FedericoBonel:
I actually would prefer this new way of typing, since, as you pointed out would allow you to provide types optionally without stacking: 

```tsx
type UserConfig = Config<{
  props: { Hero: { background: string } },
  fields: {
    color: {
      type: "color",
      mode: "hex" | "rgb"
    }
  }
}>;
```

FedericoBonel:
I found the following bugs or differences from the previous implementation of the `Config` type:

1. Categories don't get type checked
I tried what you're doing with arrays, but also with unions as it used to work, but none of them worked, categories don't get autocompleted, nor it requires me to fill them up. You can also add any category and it won't throw an error.
```tsx
type UserConfig = Config<{
  components: { Example: { title: string } };
  categories: ["typography"]; // I also tried "typography" by itself
}>;

const configExample1: UserConfig = {
  components: { ... };
  categories: { }, // Doesn't ask for typography to be defined
};

const configExample2: UserConfig = {
  components: { ... };
  categories: {
    typography: { }, // Doesn't get autocompleted
    random: { } // Doesn't throw an error
  },
};
```

2. Root fields are not typed checked
When providing the following config with a root:
```tsx
type UserConfig = Config<{
  components: { Example: { title: string } };
  root: { title: string, object: { text: string }, array: { text: string }[] };
}>;
```
    1. Root doesn't throw an error for not defined fields
    ```tsx
    const configExample: UserConfig = {
      components: { ... },
      root: {
        fields: { }, // Doesn't throw an error
        render: ({ children }) => {
          return <div>{ children }</div>
        },
      },
    };
    ```
    2. Root doesn't type check field configs, I can provide random fields, or an "array" field and not "arrayItems" and it won't throw an error.
    ```tsx
    const configExample: UserConfig = {
      components: { ... },
      root: {
        fields: {  array: { type: "array" }, object: { type: "asdasd" } }, // Doesn't throw an error
        render: ({ children }) => {
          return <div>{ children }</div>
        },
      },
    };
    ```
    3. Root doesn't type check the defined fields, I can provide random fields that are not defined in the type and it won't throw an error
    ```tsx
    const configExample: UserConfig = {
      components: { ... },
      root: {
        fields: {  random: {  type: "text" } }, // Doesn't throw an error
        render: ({ children }) => {
          return <div>{ children }</div>
        },
      },
    };
    ```

3. Custom field types can't be used inside arrayFields or objectFields
```tsx
type UserConfig = Config<{
  components: {
    Example: {
      title: string;
      array: { flag?: boolean }[];
      object: { string: string };
    };
  };
  fields: {
    userField: {
      type: "userField";
    };
  };
}>;

const configExample: UserConfig = {
  components: {
    Example: {
      fields: {
        title: {
          type: "userField", // Doesn't throw an error
        },
        array: {
          type: "array",
          arrayFields: { flag: { type: "userField" } }, // Throws an error
        },
        object: {
          type: "object",
          objectFields: { string: { type: "userField" } } // Throws an error
        }
      },
      defaultProps: {
        title: "",
        array: [],
        object: {string: "something"}
      },
      render: ({ title }) => {
        return <>{ title }</>;
      },
    },
  },
};
```

5. Array fields are accepted as strings
This one is not as important as the rest since this level of type checking never was fully achieved, but you can define an array field for a string value. This wasn't the case before.
```tsx
type UserConfig = Config<{
  components: { Example: { title: string } };
}>;

const configExample1: UserConfig = {
  components: {
    Example: {
      fields: {
        title: {
          type: "array", // Doesn't throw an error
          arrayFields: {},
        },
      },
      defaultProps: {
        title: "",
      },
      render: ({ title }) => {
        return <>{ title }</>;
      },
    },
  };
};
```

FedericoBonel:
I found the following bugs or differences from the previous implementation of the `Config` type:

### 1. Categories don't get type checked

I tried using both arrays and unions (as it used to work), but neither approach worked. Categories don’t get autocompleted, nor are they required. You can also add any category name without TypeScript throwing an error.

```tsx
type UserConfig = Config<{
  components: { Example: { title: string } };
  categories: ["typography"]; // Also tried just "typography"
}>;

const configExample1: UserConfig = {
  components: { /* ... */ },
  categories: {}, // Doesn't ask for "typography"
};

const configExample2: UserConfig = {
  components: { /* ... */ },
  categories: {
    typography: {}, // "typography" doesn't get autocompleted
    random: {}, // No error thrown even though it's not defined in the type
  },
};
```

### 2. Root children are typed as any

```tsx
type UserConfig = Config<{
  components: { Example: { title: string } };
  root: { title?: string };
}>;

const configExample: UserConfig = {
  components: { /* ... */ },
  root: {
    render: ({ children }) => <div>{children}</div>, // 'children' implicitly has an 'any' type
  },
};
```

### 3. Root fields are not type checked

When providing the following config type with a root definition:

```tsx
type UserConfig = Config<{
  components: { Example: { title: string } };
  root: { title: string, object: { text: string }, array: { text: string }[] };
}>;
```

#### 3.1 Missing fields don't throw errors

```tsx
const configExample: UserConfig = {
  components: { /* ... */ },
  root: {
    fields: {}, // No error for missing expected fields
  },
};
```

#### 3.2 Adding extra fields don't throw errors

I can provide random fields that are not defined in the root type and it won't throw an error.

```tsx
const configExample: UserConfig = {
  components: { /* ... */ },
  root: {
    fields: {
      random: { type: "text" }, // "random" not in root type, but no error
    },
  },
};
```

#### 3.3 Field configs are not type checked

I can provide random fields, or an "array" field with no "arrayItems" and it won't throw an error.

```tsx
const configExample: UserConfig = {
  components: { /* ... */ },
  root: {
    fields: {
      array: { type: "array" }, // Should require arrayFields
      object: { type: "asdasd" }, // Invalid field type, but no error
    },
  },
};
```

### 4. Custom field types can't be used inside `arrayFields` or `objectFields`

```tsx
type UserConfig = Config<{
  components: {
    Example: {
      title: string;
      array: { flag?: boolean }[];
      object: { string: string };
    };
  };
  fields: {
    userField: {
      type: "userField";
    };
  };
}>;

const configExample: UserConfig = {
  components: {
    Example: {
      fields: {
        title: {
          type: "userField", // Works
        },
        array: {
          type: "array",
          arrayFields: { flag: { type: "userField" } }, // Throws an error
        },
        object: {
          type: "object",
          objectFields: { string: { type: "userField" } }, // Throws an error
        },
      },
      defaultProps: {
        title: "",
        array: [],
        object: { string: "something" },
      },
      render: ({ title }) => {
        return <>{ title }</>;
      },
    },
  },
};
```

### 5. Array fields are accepted for strings or numbers

You can define an array field for a string or number value. This wasn't the case before.

```tsx
type UserConfig = Config<{
  components: { Example: { title: string } };
}>;

const configExample: UserConfig = {
  components: {
    Example: {
      fields: {
        title: {
          type: "array", // No error even though "title" is a string
          arrayFields: {},
        },
      },
      defaultProps: {
        title: "",
      },
      render: ({ title }) => <>{title}</>,
    },
  },
};
```

FedericoBonel (changes requested):
See above issues

chrisvxd:
@FedericoBonel did another round of reworking - hoping that helps. Also includes modifications to `RootProps` type.

FedericoBonel:
This is great! Almost everything is solved:

1. Categories do not get type checked => **Fixed**  ✔️
2. Root children are typed as `any` => **Fixed**  ✔️
3. Root
3.1 Missing fields do not throw errors => **Fixed**  ✔️
3.2 Adding extra fields does not throw errors => **Fixed**  ✔️
3.3 Field configs are not type checked => **Fixed**  ✔️
5. Custom field types cannot be used inside `arrayFields` or `objectFields` => **Fixed**  ✔️


6. Array fields are accepted for strings or numbers => **Not Fixed**  ❌

---

**I did find a couple new issues:**  

1. The `fields` key is not fully optional. This happens with both `ComponentConfig` and `Config`:

```tsx
type UserConfig = Config<{
  components: { Example: { title: string } };
  // fields: {} — Not adding this
}>;

const config: UserConfig = {
  components: {
    Example: {
      fields: {
        title: { type: "asdasd" }, // type becomes "string", not type-safe
      },
      render: () => {
        return <div></div>;
      },
    },
  },
};
```

2. Object and array fields are not enforced:

```tsx
type UserConfig = Config<{
  root: { title: string; object: { text: string }; array: { text: string }[] };
}>;

const configExample: UserConfig = {
  components: { Example },
  root: {
    fields: {
      title: { type: "text" },
      object: { type: "object", objectFields: {} }, // Should require "text"
      array: { type: "array", arrayFields: {} },    // Should require "text"
    },
  },
};
```

chrisvxd:
Believe I've fixed all of these except "Array fields are accepted for strings or numbers => Not Fixed", which is an existing bug. Equally, you can use an object field for these.

PR #1148, opened 2025-07-10
fix: button by default submits form.

aetheldev:
If anyone using Puck under form (like me) this cause trigger validations while editing or changing viewport etc.

### DIFF
--- a/packages/core/components/MenuBar/index.tsx
+++ b/packages/core/components/MenuBar/index.tsx
@@ -46,10 +46,20 @@ export function MenuBar<UserData extends Data>({
     >
       <div className={getClassName("inner")}>
         <div className={getClassName("history")}>
-          <IconButton type="button" title="undo" disabled={!hasPast} onClick={back}>
+          <IconButton
+            type="button"
+            title="undo"
+            disabled={!hasPast}
+            onClick={back}
+          >
             <Undo2Icon size={21} />
           </IconButton>
-          <IconButton type="button" title="redo" disabled={!hasFuture} onClick={forward}>
+          <IconButton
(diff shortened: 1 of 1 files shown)


packages/core/components/ExternalInput/index.tsx
```diff
@@ -196,6 +196,7 @@ export const ExternalInput = ({
                   {hasFilterFields && (
                     <div className={getClassNameModal("searchActionIcon")}>
                       <IconButton
+                        type="button"
```
matthewlynch, line 199: I'd update the `<IconButton />` component to default the `type` prop to `"button"`.

wchttzcn: That a lot make sense, I just didin't know that `<IconButton/>` actual behaviour and that's why I just change just where did they used for Puck view etc.

matthewlynch: On second thought, this component is exported so updating the type default would be a breaking change. Your PR is fine.

chrisvxd: IconButton is publicly exposed in the package, so changing the behaviour to default to here would be breaking.

Additionally, default behaviour for `button` is `submit` in form, and `IconButton` just wraps button and doesn't apply defaults.

I think the provided fix is correct as is.

chrisvxd (approved):
I think it would make more sense to change the default of

matthewlynch (approved):

matthewlynch:
@FocusThen can you run `yarn format` in the root of the repo to fix the formatting lint issue.

PR #1163, opened 2025-07-15
Add star reminder to GitHub README  

FedericoBonel:


README.md
```diff
@@ -17,6 +17,8 @@ The visual editor for React.
   </a>
 </p>
 
+> ⭐️ Enjoying Puck? Star the repo on [Github](https://github.com/puckeditor/puck)!
```
chrisvxd: Hm I might be wrong, but I think 90% of the time, the user is probably reading this on GitHub rather than npm etc, so this CTA would be redundant.

Might be better to make more generic.

```suggestion
> ⭐️ Enjoying Puck? Please leave a star!
```

FedericoBonel: I originally wrote it as you’re suggesting, but then I thought it might be worth linking people, just in case they’re seeing it on npm, even if it’s only 10% of the time.  

What about doing something like this instead?  

```suggestion
> ⭐️ Enjoying Puck? Please [leave a star](https://github.com/puckeditor/puck)!
```

chrisvxd:
Suggest you fold this into your video branch

FedericoBonel:
Sure, will close this and add it to the other one

PR #1168, opened 2025-07-15
Add prompt to star Puck on GitHub after running recipe script 

FedericoBonel:


packages/create-puck-app/index.js
```diff
@@ -192,5 +197,12 @@ program
         console.log("Failed to commit git changes");
       }
     }
+
+    console.log("\nDone! Now run:\n");
+    console.log(`  cd ${appName}`);
+    console.log(`  ${packageManager} run dev\n`);
```
FedericoBonel, line 203: I was tempted to extract this into a lookup object where the keys are each package manager and the values are the commands. That way, we could easily support other tools in the future. 

But it might be overengineering for now, since the rest of the script currently just uses simple if statements.

PR #1171, opened 2025-07-16
Add video GIF to GitHub README

FedericoBonel:


README.md
```diff
@@ -21,6 +21,8 @@ The visual editor for React.
 
 Visit https://demo.puckeditor.com/edit to try the demo.
 
+[![GIF showing a page being created in the Puck Editor, with components being added, arranged, and customized in real time](https://github.com/user-attachments/assets/916a29f2-8c8d-4c15-a7c7-6346efea26cb)](https://demo.puckeditor.com/edit)
```
FedericoBonel: I uploaded the image directly to GitHub because the Cloudinary URL  
https://res.cloudinary.com/die3nptcg/image/upload/f_gif,fl_lossy,e_sharpen:2000/demo_oylxcv.gif  
didn’t load when previewing the README on GitHub.  

I think this is due to their proxy, so I just downloaded the optimized version from Cloudinary and reuploaded it to GitHub.

chrisvxd (commented):
I'm still a bit unsure about this one. I'm concerned that

1. The image is large
2. It pushes important content (documentation CTA) below the fold

I wonder if we need to do a bigger overhaul as part of this, moving to something closer to [pragmatic-drag-and-drop](https://github.com/atlassian/pragmatic-drag-and-drop) which groups all CTAs above the gif and (interestingly) doesn't include any badges.

Image

Want to try something like this?

FedericoBonel:
Oh yeah! I’ve seen that approach multiple times, adding CTAs at the top. I actually thought about doing it with these updates but wasn’t sure if you felt strongly about keeping the current setup.  

If you don’t, then I’ll try something like that.

README.md
```diff
@@ -1,29 +1,28 @@
-# Puck
+<div align="center">
 
-The visual editor for React.
+<a href="https://puckeditor.com">
```
chrisvxd, line 5: Can you add `64px` padding to the top of this?

chrisvxd:
Also, you have conflicts so need to rebase

README.md
FedericoBonel: GitHub heavily sanitizes any styles you append to HTML elements (aside from height and a few basic ones for images). So, I added this padding using 3 line breaks. It’s slightly more than `64px` and will depend on the browser’s font size and line height, but it should do the trick.  

Another option would be using a 1x1 transparent PNG image with a data URI, but that’s probably worse for screen readers.  

Let me know if you’d prefer 2 line breaks instead of 3.

README.md
```diff
@@ -1,32 +1,33 @@
...
-
-## Documentation
-
-Visit https://puckeditor.com to view the full documentation.
+<br /><br /><br />
+
+<div align="center">
+
+<a href="https://puckeditor.com">
+  <picture>
```
chrisvxd, line 6: Final comment - can you make this max 768px and centre align? Less overwhelming and most projects do this

Preview -

Image

FedericoBonel: Done, I had to resize the image because of what we discussed before

PR #1176, opened 2025-07-18
fix: retain DropZone content in interactive mode

### LINKED ISSUE #1174: DropZone content not displayed when previewMode is "interactive"
## Description

When I switch to previewMode to "interactive", the content of my DropZones is not displayed. They display fine when the previewMode is "edit". I am not using the new slots feature, but still using DropZones.

I have identified in the code where the issue occurs. It's in puck/packages/core/components/DropZone/index.tsx in function DropZoneRender at line 543: "if (!content)". That line prevents the registerZone method from being called, so the zone content is not rendered. If I comment out that line (and line 553), it works fine.

## Environment

- Puck version: 0.19.3
- Browser version: Chrome 138 (desktop), Safari 18.5
- Additional environment info: MacOS

## Steps to reproduce

Render Puck with DropZones, then switch to "interactive" mode

chrisvxd:
Closes #1174

### DIFF
--- a/packages/core/components/DropZone/context.tsx
+++ b/packages/core/components/DropZone/context.tsx
@@ -89,21 +89,10 @@ export const DropZoneProvider = ({
     [dispatch]
   );
 
-  const unregisterZone = useCallback(
-    (zoneCompound: string) => {
-      dispatch({
-        type: "unregisterZone",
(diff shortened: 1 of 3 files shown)


FedericoBonel (approved):
Your changed fixed it!

PR #1178, opened 2025-07-18
feat: add no-external.css bundle without external fonts

### LINKED ISSUE #1157: Puck breaks with strict CSP
## Description

Hey, Our current environment has a strict CSP (Content Security Policy) in place. We noticed that the stylesheet is importing an external font:
https://github.com/puckeditor/puck/blob/7bb4823fafd73c9c3eb507f1999b153a373c5c2d/packages/core/styles.css#L1

Unfortunately, we cannot simply update our CSP policy to allow this, and it is also preventing Puck from loading.

Would you be open to including the font within the library to avoid the external import? If so, we would be happy to work on the PR.

## Environment

- Puck version: 0.19.1

## Steps to reproduce

This is happening  when `https://rsms.me/inter/inter.css"` is not part of the `Content-Security-Policy header.

Image

## What I expect to happen

Ideally, all resources should be imported from the library itself.

chrisvxd:
* Adds a new CSS module `no-external.css`, which eliminates Inter in case the user is using their own font.
* Document this under **Extending Puck / Theming** and **API Reference / Theming**
* Mark font overrides as a "feat", since these are now documented

There is a no-external JS bundle too, but this is not exported by the package module resolution.

Closes #1157 

## Tasks

- [x] Implement separate bundle
- [x] Test bundle via canary
- [x] Add docs


chrisvxd:
@fkhadra I believe this PR should address your issues. Please install `0.20.0-canary.3badc0ce` and then do the following to verify?

```css
@import "@measured/puck/no-external.css";
 
:root {
  --puck-font-family: "Inter Variable", var(--fallback-font-stack);
}
```

fkhadra:
Thanks @chrisvxd, I'm back to work tomorrow, I'll try it first thing.

fkhadra:
@chrisvxd @FedericoBonel, I tested the change in one of our live instance and I can confirm it works well. Thank you both

Issue #895, opened 2025-02-13
Respect optional fields in select options


chrisvxd:
Puck does not allow optional values in select fields.

## Steps to reproduce

Add an optional value with a select field

## Code example
```
type Props = {
  MyComponent: {
    fruit?: 'apple' | 'banana' // Make optional with `?`
  }
}

const conf: Config<Props> = {
  components: {
    MyComponent: {
      fields: {
        fruit: {
          type: "select",
          options: [
            { label: '', value: undefined }, // Should work because `fruit` is optional
            { label: 'Apple', value: 'apple' },
            { label: 'Banana', value: 'banana' },
          ]
        }
      }
    }
  }
}
```

## What happened

TS complains that `undefined` is not a valid value for `fruit`

## What I expected to happen

TS would not complain because `fruit` is optional in the Props type.

PR #1188, opened 2025-07-18
feat: support object, null and undefined in select/radio options

### LINKED ISSUE #1124: Allow undefined/null values in select options
## Description

When defining options for a select field `undefined` is not accepted, both rejected by Typescript and also the pipeline converts it into an empty string. Which is unexpected and unreliable to send a value and receive something different.

## Environment

- Puck version: 0.19.0
- Browser version: Version 137.0.7151.120 (Official Build) (64-bit)
- Additional environment info: Windows

## Steps to reproduce

1. Set select options as such

```tsx
        emphasis: {
          type: 'select',
          label: 'Emphasis',
          options: [
            { label: 'None', value: undefined }, // this is what the underlying component expects
            { label: 'Bold', value: 'bold' },
            { label: 'Italic', value: 'italic' },
          ] ,
        },
```

2. And your render logic as such

```tsx
      render: ({ editMode, puck, ...props }) => {
        console.log(`TextBlock props:`, props); // This will log {...rest,  emphasis: "" } when 'None' is selected. 

        return <TextBlock {...props} />;
      },
```

## What happens

Nullish values are converted into a empty strings

## What I expect to happen

Values to remain the same, kinda like with Storybook

chrisvxd:
Closes #1124

### DIFF
--- a/packages/core/components/AutoField/fields/RadioField/index.tsx
+++ b/packages/core/components/AutoField/fields/RadioField/index.tsx
@@ -1,9 +1,7 @@
-import { useMemo } from "react";
 import getClassNameFactory from "../../../../lib/get-class-name-factory";
 import styles from "../../styles.module.css";
 import { CheckCircle } from "lucide-react";
 import { FieldPropsInternal } from "../..";
-import { safeJsonParse } from "../../../../lib/safe-json-parse";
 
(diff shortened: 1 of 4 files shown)


FedericoBonel:
This also should close #895

PR #1189, opened 2025-07-18
feat: add overlay override

### LINKED ISSUE #934: dragOverlay override
## Description  
Currently, there is no built-in way to override the **blue dragOverlay** that Puck renders around hovered elements in the preview.  

However, when customizing the editor to better match the look and feel of your application, having control over this overlay would be helpful. A feature allowing customization of the **drag overlay** would make it easier to embed the editor and create a more seamless experience.  

## Considerations  
- This might be related to **editor theming**, which is currently being tracked in: [#139](https://github.com/measuredco/puck/issues/139).  
- This may also tie into the **plugin-based UI customization** for Puck, being tracked in: [#731](https://github.com/measuredco/puck/issues/731).  
- The class that adds the blue overlay styles seems to be implemented [here](https://github.com/measuredco/puck/blob/0ea6ce41281710ceecdc5dee8a632c06004244f8/packages/core/components/DraggableComponent/styles.module.css#L45).

## Image of the current blue overlay

Image

chrisvxd:
Closes #934


apps/docs/pages/docs/api-reference/overrides.mdx
```diff
@@ -27,5 +27,6 @@ const overrides = {
 - [`headerActions`](overrides/header-actions): Override the header actions. Return a fragment so your items appear inline.
 - [`iframe`](overrides/iframe): Override the root of the iframe. Useful for injecting styles.
 - [`outline`](overrides/outline): Override the outline.
+- [`overlay`](overrides/overlay): Override the overlay shown on hover or selection of a component.
```
chrisvxd: Wondering if this would be better called `dragOverlay`

FedericoBonel: I think `dragOverlay` would be a better choice because it would allow us to support other overlays in the future without breaking changes

FedericoBonel:
Looks good! There are a couple of `componentItem` mentions that I want to highlight, just in case they shouldn’t be there:  

- [In an example of the API reference for Plugins](https://github.com/puckeditor/puck/blob/755ee18b5d9152bad55d10ed5c1181c74c31a86a/apps/docs/pages/docs/api-reference/plugin.mdx?plain=1#L17)  
- [In the `overrideKeys` const](https://github.com/puckeditor/puck/blob/755ee18b5d9152bad55d10ed5c1181c74c31a86a/packages/core/types/API/Overrides.ts#L17) — I’m guessing this is needed until we remove these names, but shouldn’t that array also include `drawer` and `drawerItem`?  
- [In the extending Puck override example](https://github.com/puckeditor/puck/blob/755ee18b5d9152bad55d10ed5c1181c74c31a86a/apps/docs/pages/docs/extending-puck/ui-overrides.mdx?plain=1#L27)  
- [In the `DropZoneChild` component](https://github.com/puckeditor/puck/blob/755ee18b5d9152bad55d10ed5c1181c74c31a86a/packages/core/components/DropZone/index.tsx#L165) — though this might be how it’s supposed to be called internally.  
- [In the extending Puck plugins example](https://github.com/puckeditor/puck/blob/755ee18b5d9152bad55d10ed5c1181c74c31a86a/apps/docs/pages/docs/extending-puck/plugins.mdx?plain=1#L28)

FedericoBonel (approved):
I think this looks good! I searched again and couldn’t find any other mentions of `componentItem`.  

As for `components`, there don’t seem to be any, but we’ll probably find some in the future that slipped through since there are so many references to the word itself.

PR #1190, opened 2025-07-18
feat: add overlay portals for UI beneath overlay

chrisvxd:
Add overlay portals via a new `registerOverlayPortal` API. Overlay portals enable parts of the component to be punch-through the overlay, enabling targeted interactivity.

This will enable inline content editing (often conflated with rich text), amongst other functionality.

https://github.com/user-attachments/assets/020a5ec9-881e-4b8d-9a72-f49122829b76

## Usage

```tsx
import { registerOverlayPortal } from "@measured/puck";

const MyComponent = () => {
  const ref = useRef<HTMLDivElement>(null);

  useEffect(() => registerOverlayPortal(ref.current), [ref.current]);

  return <button ref={ref}>Clickable</button>
}
```

## Tasks

- [x] Don't block `click` event (buttons not clickable)
- [x] Optionally prevent dragging
- [x] Document API

### DIFF
--- a/packages/core/lib/overlay-portal/index.tsx
+++ b/packages/core/lib/overlay-portal/index.tsx
@@ -1,6 +1,5 @@
 import "./styles.css";
 
-// TODO prevent drag
 export const registerOverlayPortal = (
   el: HTMLElement | null | undefined,
   opts: { disableDrag?: boolean } = {}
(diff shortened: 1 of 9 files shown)


packages/core/lib/overlay-portal/index.tsx
```diff
@@ -0,0 +1,42 @@
+import "./styles.css";
+
+// TODO prevent drag
```
chrisvxd: Remove this comment

FedericoBonel (approved):

PR #1191, opened 2025-07-20
Inline text editing

chrisvxd:
Adds support for inline text editing 🎉 

This is achieved via two new APIs:

1. The `contentEditable` param on `text` and `textarea` fields, which will tell Puck to render a ReactNode (`span`) that enables inline editing
2. The `transforms` API, which enables Puck to modify the value of a prop before rendering in the editor, similar to how slots work (which the contentEditable mode uses under the hood). This is also available to users via a prop, and to plugins.

It relies on overlay portals (#1190)

## Usage

TBD

## TODOs

- [ ] Document transform API
- [ ] Document `contentEditable` API
- [ ] Prevent formatting in both fields (this will require separate rich text field, or rich text mode in textarea)
- [ ] Prevent new lines in `text` field


FedericoBonel:
This is awesome! We might need to change the `cmd + i` / `ctrl + i` shortcut for swapping between edit and interactive modes since it conflicts with the italics shortcut.

chrisvxd:
@FedericoBonel we don't need to change it - just disable it when focused on an inline field.

Additionally, italics and bold actually need to be disabled completely until we add rich text behaviour (separate from inline).

packages/core/lib/transforms/default-transforms.tsx
```diff
@@ -0,0 +1,45 @@
...
+        zone: propName,
+        content,
+      });
+
+    return Slot;
+  },
+});
+
+export const getInlineTextTransform = (): Transforms => ({
+  text: ({ value, parentId, field: _field, propPath, isReadOnly }) => {
```
chrisvxd: Add this to textarea

packages/core/components/InlineTextField/index.tsx
```diff
@@ -0,0 +1,129 @@
...
+      return () => {
+        ref.current?.removeEventListener("input", handleInput);
+
+        cleanupPortal?.();
+      };
+    }
+  }, [appStoreApi, ref.current, value]);
+
+  return (
+    // TODO inline elements can cause strange outline
```
chrisvxd: `inline-flex` can address this

FedericoBonel:
We can use [`contentEditable="plaintext-only"`](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Global_attributes/contenteditable#value) to disable rich text editing

ojj1123:
@chrisvxd I was wondering if you have the plan to support inline text editing on custom field.
As i've explored your demo, there is no text field on right side if `contentEditable` is `true`. We may want to use both of inline text editing and text field on right side.

```tsx
const config = {
  components: {
    Example: {
      fields: {
        title: {  // ⬅️ custom field
          type: "custom",
          render: (props) => <CustomInput {...props} />
          contentEditable: true // ⬅️ support it?
        },
      },
      render: ({ title }) => {
        return <p>{title}</p>; // Will be editable inline
      },
    },
  },
};
```

chrisvxd:
@ojj1123 good suggestion, that's been added.

PR #1216, opened 2025-07-31
chore: prevent FOUC with sidebars

chrisvxd:


packages/core/components/Puck/index.tsx
```diff
@@ -509,32 +519,13 @@ function PuckLayout<
...
-                    var(--puck-frame-width) 
-                    ${
-                      rightSideBarVisible
-                        ? rightWidth
-                          ? `${rightWidth}px`
-                          : "var(--puck-side-bar-width)"
-                        : "0"
-                    }
-                  `,
-                  }}
```
chrisvxd, line 530: This was duplicative of the CSS and didn't react to breakpoints or default vars. The `0` was what caused the FOUC.

packages/core/components/Puck/styles.module.css
```diff
@@ -38,6 +38,15 @@
...
   --puck-frame-width: auto;
   --puck-side-bar-width: 0px;
+  --puck-left-side-bar-width: var(
+    --puck-user-left-side-bar-width,
+    var(--puck-side-bar-width)
+  );
+  --puck-right-side-bar-width: var(
+    --puck-user-right-side-bar-width,
+    var(--puck-side-bar-width)
+  );
```
chrisvxd, line 48: This approach reuses existing CSS, just overriding the variables

FedericoBonel:
I love the approach, but one issue I noticed while trying out the demo, opening and closing the sidebars from the header results in two empty white columns.

https://github.com/user-attachments/assets/4cc8233f-5b0d-466b-a19e-d31370006e55

PR #1230, opened 2025-08-06
chore: prevent custom fields from losing focus on initial keydown

### LINKED ISSUE #1204: Custom fields losing focus
## Description

Custom fields are losing focus on first keystroke in canary

## Environment

- Puck version: 0.20.0 canary

## Steps to reproduce

1. Visit the canary [custom field](https://puckeditor.com/v/canary/docs/api-reference/fields/custom) docs
2. Enter text in the input

## What happens

The input loses focus on first keypress

## What I expect to happen
 
The input to retain focus

chrisvxd:
Closes #1204

### DIFF
--- a/packages/core/components/AutoField/index.tsx
+++ b/packages/core/components/AutoField/index.tsx
@@ -122,6 +122,17 @@ export type FieldPropsInternal<ValueType = any, F = Field<any>> = FieldProps<
   name?: string;
 };
 
+const defaultFields = {
+  array: ArrayField,
+  external: ExternalField,
+  object: ObjectField,
+  select: SelectField,
+  textarea: TextareaField,
+  radio: RadioField,
+  text: DefaultField,
+  number: DefaultField,
+};
+
 function AutoFieldInternal<
   ValueType = any,
   FieldType extends FieldNoLabel<ValueType> = FieldNoLabel<ValueType>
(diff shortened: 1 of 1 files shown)


FedericoBonel (changes requested):
There seems to be something wrong with the demo on this branch — when I navigate to it and click on the Hero component, it throws an error.

FedericoBonel:
These are the errors I got:

```
..\..\packages\core\components\AutoField\index.tsx (235:12) @ AutoFieldInternal
Error: React has detected a change in the order of Hooks called by AutoFieldInternal. This will lead to bugs and errors if not fixed. For more information, read the Rules of Hooks: https://react.dev/link/rules-of-hooks
```

```
..\..\packages\core\components\AutoField\fields\ExternalField\index.tsx (29:9) @ Object.ExternalField [as external]
Do not call Hooks inside useEffect(...), useMemo(...), or other built-in Hooks. You can only call Hooks at the top level of your React function. For more information, see https://react.dev/link/rules-of-hooks
```

```
..\..\packages\core\components\AutoField\index.tsx (235:12) @ AutoFieldInternal
Error: Cannot read properties of undefined (reading 'length')
```

I think is because of the `useMemo`, if the field uses hooks internally then it can crash, but I might be wrong.

chrisvxd:
@FedericoBonel good catch, only tested in docs. Pushed a fix.

Issue #1254, opened 2025-08-18
valueOf is not a function due to fast-deep-equal


themre:
## Description

When I want to use TipTap editor inside of puck editor, i get errors from fast-deep-equal library. At first I thought it's from TipTap (this library uses same library, but it has an option to replace it) and it was still causing the issue. 

## Steps to reproduce
- Add TipTap to a field
- Add TipTap extension @tiptap/extension-text-style for Color change
- Select text and change first style and then some color (no deselection of text)

I get then the following stack error:

<img width="293" height="223" alt="Image" src="https://github.com/user-attachments/assets/1e3755c5-ba7a-4b25-81fa-c068b4464fa2" />

Also path to the affected line:

<img width="269" height="114" alt="Image" src="https://github.com/user-attachments/assets/d1232361-7c58-4d6d-91a2-83cf7a33dfbc" />

## Environment

- Puck version: 0.20.0 (tried also 0.19.1)
- Browser version: Chrome 138
- Additional environment info: Vite bundler

themre:
I created a PR that replaces current comparing library. For now my fix solves the issue, that library seems to work nicely. Also i see it's more maintained compared to the current one.

themre:
Hi again, i saw that new version 0.21.0 was released. I tried it, but i still get the same error. Was the PR that was merged also part of this release? I checked also dist folder in node_modules and i see this:
```
// ../../node_modules/fast-deep-equal/index.js
var require_fast_deep_equal = __commonJS({
  "../../node_modules/fast-deep-equal/index.js"(exports, module) {
    "use strict";
    init_react_import();
    module.exports = function equal(a, b) {
      if (a === b) return true;
      if (a && b && typeof a == "object" && typeof b == "object") {
        if (a.constructor !== b.constructor) return false; ....
```

themre:
bah, noticed i missed one usage of fast-deep-equal. will create a separate PR to fix this. sorry for this!

PR #1266, opened 2025-08-21
replace fast-deep-equal with fast-equals

themre:
Hi,
I quickly tested replacing fast-deep-equal with fast-equals and it proved to be a nice replacement. It also doesn't cause the error that i posted under https://github.com/puckeditor/puck/issues/1254. Hope we can continue with this to not cause others same issue.
Also great tool that you created,really nice API!

Closes: #1254


chrisvxd (changes requested):
Thanks for the contribution! Happy to evaluate this but your PR contains a lot of unrelated changes - formatting, import changes, etc.

Would you be able to strip it back just to the changes you're introducing, so I can test it properly?

themre:
hi @chrisvxd , thx for the reply. i reverted formatting changes (Biome kicked in, sorry). for the other changes (included CSS in demo, replaced `assert` with `with`), i had to do those since i was without styling when testing locally and locally build didn't work if i used `assert`. Hope this is ok.

apps/docs/next.config.mjs
```diff
@@ -1,4 +1,4 @@
-import packageJson from "./package.json" assert { type: "json" };
+import packageJson from "./package.json" with { type: "json" };
```
chrisvxd: Please revert this, you're running an incorrect version of node

apps/demo/app/styles.css
```diff
@@ -1,4 +1,5 @@
 @import url("https://rsms.me/inter/inter.css");
+@import "@/core/dist/index.css";
```
chrisvxd: Please remove this

packages/core/components/Puck/index.tsx
```diff
@@ -1,41 +1,41 @@
...
+	createContext,
+	PropsWithChildren,
+	ReactElement,
+	ReactNode,
+	useCallback,
+	useContext,
+	useEffect,
+	useMemo,
+	useRef,
+	useState,
```
chrisvxd: This file still contains extensive formatting changes

apps/docs/next.config.mjs
themre: done

apps/demo/app/styles.css
themre: done

packages/core/components/Puck/index.tsx
themre: done

FedericoBonel:
@chrisvxd

I ran benchmarks on the libraries themselves, comparing two Puck data objects taken straight out of the demo. Both objects had differences in deeply nested cards.  

All values are fixed-point, so results are approximate:  

```
fast-deep-equal: 0.0039 ms (avg over 1,000,000 comparisons)
fast-deep-equal ops/sec: 258,874.6300

fast-equals: 0.0037 ms (avg over 1,000,000 comparisons)
fast-equals ops/sec: 271,075.6580

fast-equals took an average of 4.50% less time over 1,000,000 comparisons.
```

I also ran some benchmarks inside the editor itself. Results varied widely between tests (mainly due to environment factors), but overall performance seems to break about even with the current library (I tested this with a production build of the demo app):  

```
insertion in fast-deep-equal: 49.6000 ms, renders: 48
insertion in fast-equals: 50.4333 ms, renders: 39
→ insertion took 1.68% more time, and 18.75% fewer re-renders

replacement in fast-deep-equal: 9.6000 ms, renders: 8
replacement in fast-equals: 9.5000 ms, renders: 8
→ replacement took 1.04% less time, with the same number of re-renders

reorder in fast-deep-equal: 43.0100 ms, renders: 43
reorder in fast-equals: 38.2000 ms, renders: 28
→ reorder took 11.37% less time, and 34.88% fewer re-renders

deletion in fast-deep-equal: 12.1000 ms, renders: 8
deletion in fast-equals: 13.6000 ms, renders: 7
→ deletion took 12.40% more time, and 12.50% fewer re-renders
```

chrisvxd (approved):

Issue #1132, opened 2025-07-09
Delete hotkey


chrisvxd:
## Description

Add a delete hotkey.

## Considerations

- Should we support backspace or just del?
- Or CMD + backspace?

### Proposal 1

Delete the currently selected component when the backspace or delete key is pressed.

### Proposal 2

Delete the currently selected component when the cmd+backspace or delete key is pressed.

coftad:
@chrisvxd Created pull request

chrisvxd:
Closed in #1275

PR #1275, opened 2025-08-25
feat: add delete hotkey for component deletion (backspace/delete)

SilasRopp1012:
## Description  
Addresses #1132
This PR adds delete hotkeys for removing the currently selected component. 

### Changes  
- Implemented **Proposal 1**: pressing **Backspace** or **Delete** will remove the selected component.  
- Made a small adjustment to the global hotkey functionality to ensure the delete/backspace combo works smoothly without interfering with other bindings and the native deletion in sidebar text editing.  

### Rationale  
Most design tools (e.g., Figma, Canva, etc.) allow deleting selected components with backspace or delete, so this behavior aligns with user expectations. Proposal 1 was chosen for consistency with those tools.

### DIFF
--- a/packages/core/lib/use-delete-hotkeys.ts
+++ b/packages/core/lib/use-delete-hotkeys.ts
@@ -1,0 +1,41 @@
+import { useCallback } from "react";
+import { useHotkey } from "./use-hotkey";
+import { useAppStoreApi } from "../store";
+
+const isEditableElement = (target: EventTarget | null): boolean => {
+  if (!target || !(target instanceof HTMLElement)) return false;
+
+  const tagName = target.tagName.toLowerCase();
+  if (tagName === "input" || tagName === "textarea") return true;
+  if (target.isContentEditable) return true;
+
+  return false;
+};
+
+export const useDeleteHotkeys = () => {
+  const appStore = useAppStoreApi();
+
+  const deleteSelectedComponent = useCallback((e?: KeyboardEvent) => {
+    if (isEditableElement(document.activeElement)) {
+      return false;
+    }
+
+    const { state, dispatch } = appStore.getState();
+    const { itemSelector } = state.ui;
+
+    if (!itemSelector || !itemSelector.zone) {
+      return false;
+    }
+
+    dispatch({
+      type: "remove",
+      index: itemSelector.index,
+      zone: itemSelector.zone,
+    });
+
+    return true;
+  }, [appStore]);
+
+  useHotkey({ delete: true }, deleteSelectedComponent);
+  useHotkey({ backspace: true }, deleteSelectedComponent);
+};
--- a/packages/core/components/Puck/index.tsx
+++ b/packages/core/components/Puck/index.tsx
@@ -50,6 +50,7 @@ import { useLoadedOverrides } from "../../lib/use-loaded-overrides";
 import { DefaultOverride } from "../DefaultOverride";
 import { useInjectGlobalCss } from "../../lib/use-inject-css";
 import { usePreviewModeHotkeys } from "../../lib/use-preview-mode-hotkeys";
+import { useDeleteHotkeys } from "../../lib/use-delete-hotkeys";
 import { useRegisterHistorySlice } from "../../store/slices/history";
 import { useRegisterPermissionsSlice } from "../../store/slices/permissions";
 import { monitorHotkeys, useMonitorHotkeys } from "../../lib/use-hotkey";
@@ -481,6 +482,7 @@ function PuckLayout<
   const ready = useAppStore((s) => s.status === "READY");
 
   useMonitorHotkeys();
+  useDeleteHotkeys();
 
   useEffect(() => {
     if (ready && iframe.enabled) {
--- a/packages/core/lib/use-hotkey.ts
+++ b/packages/core/lib/use-hotkey.ts
@@ -32,6 +32,8 @@ const keys = [
   "x",
   "y",
   "z",
+  "delete",
+  "backspace",
 ] as const;
 
 type KeyStrict = (typeof keys)[number];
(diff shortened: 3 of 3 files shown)


packages/core/lib/use-delete-hotkeys.ts
```diff
@@ -0,0 +1,41 @@
...
+      type: "remove",
+      index: itemSelector.index,
+      zone: itemSelector.zone,
+    });
+
+    return true;
+  }, [appStore]);
+
+  useHotkey({ delete: true }, deleteSelectedComponent);
+  useHotkey({ backspace: true }, deleteSelectedComponent);
```
chrisvxd, line 94: These need to respect the `delete` permissions

chrisvxd (changes requested):
Excellent work! Love the change to the hotkey API, too.

Your linting is failing (need to run prettier and commit via `npm run format`), and I left a comment.

packages/core/lib/use-delete-hotkeys.ts
```diff
@@ -0,0 +1,41 @@
...
+  return false;
+};
+
+export const useDeleteHotkeys = () => {
+  const appStore = useAppStoreApi();
+
+  const deleteSelectedComponent = useCallback((e?: KeyboardEvent) => {
+    if (isEditableElement(document.activeElement)) {
+      return false;
+    }
```
chrisvxd: Hm, on second thoughts, this check might not quite robust enough. For example, `select` has default behaviour on "Delete" in some browsers, too. Additionally, the user might have a custom field that listens for `delete` hotkeys for whatever reason (RTE, etc), or have a modal open and not realise that hitting backspace will delete something they can't see.

Is there a way you could refine this?

packages/core/lib/use-delete-hotkeys.ts
SilasRopp1012: I’ve updated the handler to check delete permissions before dispatching remove. The hotkeys are still registered unconditionally (Rules of Hooks), but the callback now checks delete permissions. If the item lacks permission, we swallow the key to avoid browser back-nav but don’t delete.

packages/core/lib/use-delete-hotkeys.ts
SilasRopp1012: Thanks for pointing that out. I replaced the simple isEditableElement with a more robust shouldBlockDeleteHotkey that: 
 - defers to native inputs (including select), contentEditable, and common ARIA roles that use delete
 - honors e.defaultPrevented and uses e.composedPath()[0] for shadow-DOM origins
 - blocks when a modal/dialog is open.

In the canvas, if there’s no selection or no permission, we still swallow Delete/Backspace to prevent browser back-nav. I've tested manually and the functionality remains the same for simple delete actions on the demo page.

FedericoBonel:
Hey @SilasRopp1012! This is amazing, thank you for all the contributions you have been doing lately, they are great 🙏.

I gave this a read and also tried it out locally. Here are my thoughts:

### On modals
- The modal and dialog selector you’re using doesn’t work with Puck’s [external field](https://puckeditor.com/docs/integrating-puck/external-data-sources). **That’s on us**, because we should be adding `role="dialog"` (or using the native `dialog` element) for better accessibility and document structure.  
- The selector can also collide with dialogs that are not visible but still remain in the DOM, which is a pretty common pattern. Right now you’re matching any element with `aria-modal="true"` or with `role="dialog"` where `aria-hidden="true"` isn’t defined. For example, if I only set `role="dialog"` and don’t set `aria-hidden`, or if I set `aria-modal="true"` and don’t remove it from the DOM, the selector still matches it and it would block the delete. 
If you go to the [React MUI dialog docs](https://mui.com/material-ui/react-dialog/), open the console, and run your selector, you’ll see it returns results even when no dialog is actually open.
  - It’s really hard to cover every single case. But we could also add a check of any dialog's (`dialog, [role="dialog"], [role="alertdialog"], [aria-modal="true"]`) and their ancestor's styles and see if they're visible or not to decide if we should block the delete (i.g. `style.display === "none"`, `style.visibility === "hidden"`, etc.).

### On RTE libraries
- I tried a couple and your handling works correctly. Most reliable ones either set `contenteditable="true"` for accessibility (for example, Tiptap), or use a `<textarea>` element (for example, Monaco). We can update the heuristics based on community feedback once we merge this too, and we can also ask users to add an `onKeyDown` listener on the parent that calls `preventDefault` when needed.

SilasRopp1012:
Thanks for the feedback 🙏  

I updated the modal check so that the delete hotkey is only blocked when a dialog is actually **visible**.

### Changes
- Added an `isElementVisible` helper that checks:
  - `display: none`
  - `visibility: hidden`
  - `opacity: 0`
  - `aria-hidden="true"`
  - `hidden` attribute
  - (walks up ancestor chain to catch inherited styles)
- Updated the modal selector to cover `.Modal--isOpen`, more consistent with Puck's modal component implementation
- Only visible modals will block delete, so hidden/inactive dialogs (like in MUI docs) no longer interfere.
- RTE handling remains unchanged and continues to work as before.

packages/core/lib/use-delete-hotkeys.ts
```diff
@@ -0,0 +1,95 @@
...
+      role === "searchbox" ||
+      role === "listbox" ||
+      role === "grid"
+    ) {
+      return true;
+    }
+  }
+
+  const modal = document.querySelector(
+    'dialog[open], [aria-modal="true"], [role="dialog"], [role="alertdialog"], .Modal--isOpen'
```
FedericoBonel:
```suggestion
    'dialog[open], [aria-modal="true"], [role="dialog"], [role="alertdialog"]'
```

Great improvements, this is exactly what I had in mind! Thank you very much.

The only thing I’d point out is that the modal selector for the Puck modal isn’t working because we’re using CSS modules. When you deploy the core package, the class you’re using will be appended with a hash and other modifiers, so it won’t match "Modal--isOpen".

I think you can drop it, the issue is on our side. We should add a `role` attribute to the external field to follow best HTML practices, rather than tying this implementation to our specific modal, which could change in the future.

FedericoBonel (approved):

PR #1289, opened 2025-08-29
fix(core): update FieldComponent on `custom` field update using `resolveFields`

### LINKED ISSUE #1288: Custom field resolution in `resolveFields`
## Description

The fields list does not re-render after `resolveFields` is resolved when two `custom` fields are interchanged.

```jsx
const CustomString = {
  type: "custom",
    return (
      <div>
        <span>Text input</span>
        <input value={value} onChange={(e) => onChange(e.target.value)} />
      </div>
    );
  },
};

const CustomNumber = {
  type: "custom",
    return (
      <div>
        <span>Number input</span>
        <input
          type="number"
          value={value}
        />
      </div>
    );
  },
};

  return (
    <Puck
      config={{
        components: {
          CustomHeading: {
            fields: {
              value: CustomString,
            },
            resolveFields: (data, { fields }) => {
              // After the condition below is met, the `CustomNumber` field is not rendered. 
              // `CustomString` is retained.

              if (data.props.value === "123") {
                return {
                  ...fields,
                  value: CustomNumber,
                };
              }

              return fields;
            },
            render: ({ value }) => <h1>{value}</h1>,
          },
        },
        root: {
        },
      }}
      data={{}}
    />
  );
}
```

## Environment

- Puck version: 0.20.1

## Steps to reproduce

Sandbox: https://codesandbox.io/p/sandbox/puck-bug-forked-x4ts9n

## What happens

In the sandbox, you only need to type **123** in the field `value`:
- `Heading`: This block switches between two native Puck fields. **It works fine**.
- `CustomHeading`: This block switches between two custom fields. **Does not work**.
- `MixedHeading`: This block switches between a puck native field and a custom field. **It works**.

## What I expect to happen

We should be able to interchange `custom` fields on `resolveFields.`

## Additional Media

Discord: https://discord.com/channels/1153376562259951687/1153376562830397563/1410649324441964574

leweyse:
Closes #1288.

Since [`mergedProps`](https://github.com/puckeditor/puck/blob/main/packages/core/components/AutoField/index.tsx#L173) also depends on `field`, this change does not create extra re-renders.

I used [`react-scan`](https://react-scan.com/) to track the perfomance.

Before:

https://github.com/user-attachments/assets/537dbba3-5cd8-4b46-a104-ff90e6055f68

After:

https://github.com/user-attachments/assets/84f715a0-5b37-4f18-80d5-1acd4aaabe7c

### DIFF
--- a/packages/core/components/AutoField/index.tsx
+++ b/packages/core/components/AutoField/index.tsx
@@ -229,7 +229,7 @@ function AutoFieldInternal<
     } else if (field.type !== "slot") {
       return render[field.type] as (props: FieldProps) => ReactElement;
     }
-  }, [field.type, render]);
+  }, [field, render]);
 
   const { visible = true } = props.field;
 


FedericoBonel:
Hey @leweyse! Thanks for the contribution.  

I think the issue with this is that custom fields would now be unmounted and remounted on every prop change when using `resolveFields`. This is actually a bug that [was fixed in 0.20.0 compared with 0.19.3](https://github.com/puckeditor/puck/issues/1291).  

For example, if you use the following component, you’ll see that the `useEffect` runs on any prop change. This also causes the custom field to lose focus whenever you type in it. Another problem is that custom fields can contain any component with internal state, and remounting these components resets their state, which could break functionality:  

```tsx
Custom: {
  resolveFields: () => {
    return {
      anotherField: {
        type: "text",
      },
      title: {
        type: "custom",
          const [state, setState] = useState("");

          useEffect(() => {
            console.log("Mounting field"); // Logs on any prop change

            return () => console.log("Unmounting field"); // Logs on any prop change
          }, []);

          console.log(state); // State resets to "" whenever anotherField changes

          return (
            <input
              value={value}
              onChange={(e) => {
                onChange(e.target.value);
                setState(e.target.value);
              }}
            />
          );
        },
      },
    };
  },
  defaultProps: { anotherField: "Another", title: "Title" },
  render: ({ anotherField, title }) => {
    return <div>Test</div>;
  },
},
```

Do you think there’s a way we could add some kind of dependency check, so the component only remounts when it actually needs to?

leweyse:
Hey @FedericoBonel! Thank you for taking a look.

I understand the issue, and I have a couple of suggestions. 

The problem in the example above is that the render function does not hold a reference between `prop` changes. To solve this, the user could define the custom field outside the `resolveFields` scope:

```tsx
const CustomTextarea: CustomField<string | undefined> = {
  type: "custom",
    return (
      <textarea
        value={value}
      />
    );
  }
};

const YourBlock = {
  resolveFields: (data) => {
     return {
       your_field: CustomTextarea
     }
  },
};
```

This way, the `custom` field component is not remounted on `prop` changes. Maybe Puck can encourage this pattern?

But, if the user really needs to define the `custom` field inside the `resolveFields` scope, my only suggestion would be to add a new property `key` (optional) to the field definition:
 
```tsx
const CustomField = {
  key: "custom_text_field",
  type: "custom",
  render: ({ name, onChange, value }) => (
    <input
      defaultValue={value}
      name={name}
    />
  ),
}
```

This key could be added to the `FieldComponent` dependencies to update the component rendered.

FedericoBonel:
Hey @leweyse!

Thank you so much for your patience and for the work you put into this. We've decided to add a `key` prop to the custom field as per your suggestion, and we've already opened a PR for it over [here](https://github.com/puckeditor/puck/pull/1378) 🙏. I'm so sorry for taking over this one.

Feel free to give it a try or leave any feedback you might have!

Issue #1273, opened 2025-08-25
`resolveComponentData` cache ignores "force" trigger parameter, preventing forced re-execution


b-barry:
## Description

The `resolveAllData` function's "force" trigger parameter is ignored by the caching mechanism in `resolveComponentData`, preventing forced re-execution of `resolveData` functions even when external data has changed but component props remain the same.

## Environment

- Puck version: 0.19.3
- Browser version: Nothing
- Additional environment info: 22

## Steps to reproduce

1. Create a component with a `resolveData` function that fetches external data:

```tsx
const config = {
  components: {
    MyComponent: {
      fields: { title: { type: "text" } },
      resolveData: async ({ props }, {metadata}) => {
        // Simulate external data fetch
        const externalData = await fetchExternalData(props, metadata);
        return {
          props: { ...props, resolvedTitle: externalData.title }
        };
      },
      render: ({ resolvedTitle }) => <div>{resolvedTitle}</div>
    }
  }
};
```

2. Call `resolveAllData` multiple times with the same data:

```tsx
const data = {
  content: [{ type: "MyComponent", props: { id: "test", title: "Original" } }],
  root: { props: {} }
};
const metadata={ id:1}

// First call - executes resolveData
const result1 = await resolveAllData(data, config, metadata);

// Second call - should force re-execution but doesn't due to cache
const result2 = await resolveAllData(data, config, metadata);
```

## What happens

The second call to `resolveAllData` returns cached results from the first call, even though the function passes `"force"` as the trigger parameter. [1](#6-0)  The cache check in `resolveComponentData` [2](#6-1)  only considers deep equality of the item data and ignores the trigger parameter entirely.

## What I expect to happen

When `resolveAllData` calls `resolveComponentData` with `trigger: "force"`, the cache should be bypassed and `resolveData` functions should be re-executed to fetch fresh data, even if the component props haven't changed.

FedericoBonel:
Hey @b-barry! Thanks for reporting this.

I was able to reproduce the issue with your example. For reference, [here is the check in `resolveComponentData`](https://github.com/puckeditor/puck/blob/9172a57ff36343ed661d44c5de9f71695f0e5f60/packages/core/lib/resolve-component-data.ts#L43).

chrisvxd:
Closed by #1295

PR #1295, opened 2025-09-03
fix(resolveComponentData): bypass cache when trigger is force

SilasRopp1012:
Adresses #1273 Fix: honor trigger "force" in resolveComponentData

This should fix the issue where data is pulled from the cache even when force trigger is present.

I manually tested and it still passes existing tests, just a small change.


henning-f:
I just came to the office and updated to the fresh version of puck on my feature branch to figure out that you solved the issue that was causing my worst headaches in the last days implementing SSR with resolveAllData. Now it works like a charm. Mille Grazie 😁

Issue #1226, opened 2025-08-05
Support CSS units for `minEmptyHeight` (`px`, `%`, `vh`, …)


rinormhi:
## Description

Currently the `minEmptyHeight` option only accepts a numeric value (always interpreted as pixels). In many use cases—especially flexible or responsive layouts—it would be extremely helpful to specify other CSS units (e.g. `vh`, `%`, `rem`) so that the empty area can adapt to its container or the viewport.

FedericoBonel:
Hey @rinormhi!

Thanks for requesting this. I think having more possibilities is always welcome, though there might be some limitations with how it’s currently implemented. Regardless, this is definitely something people could need.

PR #1298, opened 2025-09-04
Support css units for `minEmptyHeight`

DamianKocjan:
Closes: #1226


apps/docs/pages/docs/api-reference/fields/slot.mdx
```diff
@@ -182,15 +182,15 @@ const config = {
...
-| [`minEmptyHeight`](#minemptyheight) | `minEmptyHeight: 256`        | Number        | -      |
-| [`ref`](#ref)                       | `ref: ref`                   | Ref           | -      |
-| [`style`](#style)                   | `style: {display: "flex"}`   | CSSProperties | -      |
+| Param                               | Example                      | Type                       | Status |
+| ----------------------------------- | ---------------------------- | -------------------------- | ------ |
+| [`allow`](#allow-1)                 | `allow: ["HeadingBlock"]`    | Array                      | -      |
+| [`className`](#classname)           | `className: "MyClass"`       | String                     | -      |
+| [`collisionAxis`](#collisionaxis)   | `collisionAxis: "x"`         | String                     | -      |
+| [`disallow`](#disallow-1)           | `disallow: ["HeadingBlock"]` | Array                      | -      |
+| [`minEmptyHeight`](#minemptyheight) | `minEmptyHeight: "256px"`    | CSSProperties["minHeight"] | -      |
```
chrisvxd: This is a breaking change that we can't introduce, as many users already use this with numbers

rinormhi: Hey @chrisvxd, would it be possible to extend the type to
`userMinEmptyHeight: CSSProperties["minHeight"] | number;` ?
That way, existing users who pass a number would continue to work as before (falling back to px), so it wouldn’t introduce a breaking change.

DamianKocjan:
@chrisvxd I've tested this on Firefox 144.0 (64-bit) and Google Chrome 141.0.7390.108 (64-bit). 

The animations work correctly with both pixel and non-pixel values (`rem`, `em`, `vw`). Percentage values (`%`) don't work because height resolves to `0px`.

Issue #1221, opened 2025-08-05
Using `defaultProps` for slots at the root level results in error


FedericoBonel:
## Description

Slots defined in the [root config](https://puckeditor.com/docs/api-reference/configuration/config#root) break when `defaultProps` is used to provide nested components.

This issue occurs because `defaultProps` in the root config does not automatically generate IDs for slot components, unlike when used in component configs.

## Environment

- Puck version: 0.19.3

## Steps to reproduce

1. Render Puck with the following config that uses slots with default props at the root level, without setting component IDs for nested content:

```tsx
export const config: Config<Props> = {
  components: {
    Heading: {
      fields: {
        title: { type: "text" },
      },
      render: ({ title }) => <Heading title={title} />,
    },
  },
  root: {
    fields: {
      content: { type: "slot" },
      object: { type: "object", objectFields: { content: { type: "slot" } } },
      array: {
        type: "array",
        arrayFields: { content: { type: "slot" } },
      },
    },
    defaultProps: {
      content: [
        {
          type: "Heading",
          // Does not set an ID
          props: { title: "Hello from the root" },
        },
      ],
      object: {
        content: [
          {
            type: "Heading",
            // Does not set an ID
            props: { title: "Hello from the root" },
          },
        ],
      },
      array: [
        {
          content: [
            {
              type: "Heading",
              // Does not set an ID
              props: { title: "Hello from the root" },
            },
          ],
        },
      ],
    },
    render: ({ children, slot: Content, object: { content: ObjectContent }, array }) => {
      return (
        <div>
          <h2>Default content:</h2>
          {children}
          <hr />
          <h2>Non nested content:</h2>
          <Content />
          <hr />
          <h2>Object nested content: </h2>
          <ObjectContent />
          <hr />
          <h2>Array nested content: </h2>
          {array.map(({ content: ArrayContent }, index) => (
            <ArrayContent key={index} />
          ))}
        </div>
      );
    },
  },
};
```

2. Navigate to the editor

## What happens

The editor throws an error:
```bash
Each child in a list should have a unique "key" prop.

Check the render method of `ul`. It was passed a child from LayerTree. See https://react.dev/link/warning-keys for more information.
```

*This does not occur when IDs are explicitly provided for the nested components.*

## What I expect to happen

Slot fields in the root config should behave the same as in component configs. If default slot content is provided, it should be rendered and used without having to manually generate IDs.

## Additional Media

[...]

PR #1303, opened 2025-09-08
fix: Generating new item id in walk-app-state if one doesn't exist

xaviemirmon:
Linked to #1221 

If only default props were defined for slots on root fields no ids were being created.   Without the id the keys weren't being added and all were `undefined` leading to the duplicate ids error. I've added a check to create an id if one doesn't exist when walking the app state.

### DIFF
--- a/packages/core/lib/data/walk-app-state.ts
+++ b/packages/core/lib/data/walk-app-state.ts
@@ -15,6 +15,7 @@ import {
 } from "../../types/Internal";
 import { mapFields } from "./map-fields";
 import { flattenNode } from "./flatten-node";
+import { generateId } from "../generate-id";
 
 /**
  * Walk the Puck state, generate indexes and make modifications to nodes.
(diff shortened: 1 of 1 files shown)


packages/core/lib/data/walk-app-state.ts
```diff
@@ -102,7 +103,8 @@ export function walkAppState<UserData extends Data = Data>(
     // Only modify the item if the user has returned it, enabling us to prevent unnecessary mapping and creating new references, which results in re-renders
     if (!mappedItem) return item;
 
-    const id = mappedItem.props.id;
+    // Generate a unique ID for the item if it doesn't have one
+    const id = mappedItem.props.id || generateId(mappedItem.type);
```
chrisvxd, line 107: Hm, this will work, but what do we do for defaultProps outside of root? Not saying we should necessarily use that approach, but it would be good to be consistent.

Additionally, if we do go with this approach, we should definitely add some regression tests. walk-app-state is super complex.

chrisvxd:
Closing in favour of #1347

Issue #1161, opened 2025-07-14
Array items with slots break when adding or removing items too quickly


FedericoBonel:
## Description

If you nest slots and other fields inside array fields and add or remove items too quickly, the number of slots rendered in the canvas becomes inconsistent with the number of items in the array field.

## Environment

- Puck version: 0.19.2

## Steps to Reproduce

1. Render the editor with the following config that nests Slots and a text field inside an array field:

```tsx
const config = {
  components: {
    Example: {
      fields: {
        someArray: {
          type: "array",
          arrayFields: {
            title: { type: "text" },
            someContent: { type: "slot" },
          },
        },
      },
      defaultProps: {
        someArray: [],
      },
      render: ({ someArray }) => {
        return (
          <>
            {someArray.map(({ someContent: SomeContent }, idx) => (
              <SomeContent key={idx} />
            ))}
          </>
        );
      },
    },
  },
};
```

2. Navigate to the editor.
3. Drag and drop an "Example" component.
4. Add more than 20 items.
5. Remove 10 items as quickly as possible.

## What Happens

The number of Slots rendered in the canvas becomes inconsistent with the number of items in the array field.

## What I Expect to Happen

The number of Slots rendered in the canvas should match the number of items currently present in the array field.

## Additional Media

https://github.com/user-attachments/assets/22b9089e-d2de-4eea-91e4-8270741faca8

xaviemirmon:
I've just been playing around with this to reproduce and it's not just with slots it's generic to an array field.  I suspect the state is getting debounced causing not all items to be removed.

xaviemirmon:
From what I can work out the problem boils down to us hitting a race condition where data isn't updated quickly enough after the onclick event before the next event is fired.  I've added a 5ms debounce to allow the a bit of time for the process to complete before the next event happens.

FedericoBonel:
From what I can tell this was fixed with #1213, I can't reproduce this with and without throttling my CPU

PR #1304, opened 2025-09-08
fix: debouncing arrayField delete event by 5ms so that the state doesn't get out of sync

xaviemirmon:
Related issue: #1161 

The arrayField data was getting out of sync if the user clicked delete too quickly.  I've added a debounce to avoid hitting the race condition.


xaviemirmon:
Closing down in favour of upcoming arrayField refactor.

PR #1306, opened 2025-09-09
fix: resolve permissions in drop zone so that slots maintain permissions

### LINKED ISSUE #1198: `resolvePermissions` does not work for components nested inside slots
## Description  

Permissions set using `resolvePermissions` do not work when dropping components inside slot fields.  
They do work at the `content` level (outside nested slots in the `default-zone`).  

## Environment  

- **Puck version:** 0.19.3  

## Steps to reproduce  

1. Render the `<Puck />` component with the following config, which blocks all permissions for `Heading` component instances using `resolvePermissions`:  

   ```tsx
   const config = {
     components: {
       Heading: {
         fields: {
           title: { type: "text" },
         },
         defaultProps: { title: "Title" },
         resolvePermissions: () => {
           return {
             drag: false,
             duplicate: false,
             delete: false,
             edit: false,
           };
         },
         render: ({ title }) => {
           return <h1>{title}</h1>;
         },
       },
       Container: {
         fields: {
           Content: { type: "slot" },
         },
         render: ({ Content }) => {
           return (
             <div>
               <Content />
             </div>
           );
         },
       },
     },
   };
   ```

2. Navigate to the editor.  
3. Drag and drop a `Container` into the canvas.  
4. Drag and drop a `Heading` inside the `Container` in the canvas.  

## What happens  

The resolved permissions are **not** applied to components nested inside slots. The `Heading` can still be dragged, deleted, and edited.  

## What I expect to happen  

The resolved permissions should apply to **all components**, regardless of where they are dropped.  

## Additional media  

https://github.com/user-attachments/assets/78ab3995-c58c-4359-a45c-1f549fe3738a

xaviemirmon:
Resolves #1198 

@chrisvxd not sure if this is the correct place to have this but I've added resolve permissions to the dropzone so that component permissions are resolved even if they are in a slot.

### DIFF
--- a/packages/core/components/DropZone/index.tsx
+++ b/packages/core/components/DropZone/index.tsx
@@ -149,6 +149,15 @@ const DropZoneChild = ({
   const isLoading = useAppStore(
     (s) => s.componentState[componentId]?.loadingCount > 0
   );
+
+  // refreshPermissions in DropZone so that slots maintain component permissions
+  const refreshPermissions = useAppStore(
+    (s) => s.permissions.refreshPermissions
+  );
+  useEffect(() => {
+    refreshPermissions({ item });
+  }, [item, refreshPermissions]);
+
   const isSelected = useAppStore(
     (s) => s.selectedItem?.props.id === componentId || false
   );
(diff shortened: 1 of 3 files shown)


xaviemirmon:
@chrisvxd this should be good to go now

PR #1307, opened 2025-09-09
WIP fix: moving array logic above slots in field map so that slots inside array fields with the same name can are processed in the correct order

### LINKED ISSUE #1126: Error when nesting slots inside array fields with the same name
## Description

Slots nested inside array fields that share the same field name throw an error when adding new array items.

## Environment

- Puck version: 0.19.1

## Steps to reproduce

1. Nest a `Slot` field inside an array field with the same name.

```tsx
const config = {
  components: {
    Example: {
      fields: {
        // Array field with name: "list"
        list: {
          type: "array",
          arrayFields: {
            // Slot field with the same name as the array field: "list"
            list: { type: "slot" },
          },
        },
      },
      defaultProps: {
        list: [],
      },
      render: ({ list }) => <div>{list.map((item: Item, idx) => <Item key={idx} />)}</div>,
    },
  },
};
```

2. Navigate to the editor.
3. Drag and drop the component that nests the slot inside the array field with the same name.
4. Add an item to the array field.

## What happens

An error is thrown:

```bash
TypeError: arr.some is not a function
```

## What I expect to happen

A new item should be added to the array field, and a new dropzone should render in the canvas.

## Additional media

https://github.com/user-attachments/assets/d457ecab-4874-486e-9cc1-e6912fa97c25

xaviemirmon:
fixes #1126

I've reversed where the value mapping for objects/arrays and slots go.  The name clash was caused by slots getting mapped first when the propKey was the same. 

Worth noting that having the same names in a nested config required some additional mapping in the config file:

```tsx
const config = {
  components: {
    Example: {
      fields: {
        // Array field with name: "list"
        list: {
          type: "array",
          arrayFields: {
            // Slot field with the same name as the array field: "list"
            list: { type: "slot" },
          },
        },
      },
      defaultProps: {
        list: [{ list: [] }],
      },
      render: ({ list }: { list: Array<{ list: Item[] }> }) => {
        return (
          <div>
            {list.map((slot, index) => {
              return slot.list.map((item: Item, idx) => <Item key={idx} />);
            })}
          </div>
        );
      },
    },
  },
};
export default config;
```

### DIFF
--- a/packages/core/lib/data/map-fields.ts
+++ b/packages/core/lib/data/map-fields.ts
@@ -65,52 +65,6 @@ export const walkField = ({
 }: WalkFieldOpts): any | Promise<any> => {
   const fieldType = fields[propKey]?.type;
   const map = mappers[fieldType];
-
-  if (map && fieldType === "slot") {
-    const content = (value as Content) || [];
-
-    const mappedContent = recurseSlots
-      ? content.map((el) => {
-          const componentConfig = config.components[el.type];
-
-          if (!componentConfig) {
-            throw new Error(`Could not find component config for ${el.type}`);
-          }
-
-          const fields = componentConfig.fields ?? {};
-
-          return walkField({
-            value: { ...el, props: defaultSlots(el.props, fields) },
-            fields,
-            mappers,
-            id: el.props.id,
-            config,
-            recurseSlots,
-          });
-        })
-      : content;
-
-    if (containsPromise(mappedContent)) {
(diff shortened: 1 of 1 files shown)


xaviemirmon:
Test have failed because the data structure of the test snapshots changed would be could to get your thoughts @chrisvxd.  It's not `slot: []` not `slot: null`.

xaviemirmon:
Going to close for now as the test failures are problematic

Issue #730, opened 2024-12-14
Improve Outline UI


chrisvxd:
## Description

The current outline UI is somewhat clunky, and should be streamlined to better match a more typical "layer" outline that is common in design tools like Figma.

## Considerations

* The paradigm is slightly different as DropZones do not behave like layers -- a single component can have multiple DropZones.

## Requirements

* Drag and drop
* Ability to delete and duplicate (perhaps via right click menu)

lukasfischer:
We just had a discussion at our company, and we also wanted to open this issue. I think puck editor should provide this feature.

There is one additional small improvement. We have the issue, when offering 10+ components in the left sidebar, the outline out of sight. I like how cursor ai editor solves this with little tabs at the top of the left sidebar. This could be a solution for that. So we would need 2 tabs - one for components, one for outline. Probably this "tabs" could be pluggable - for custom functions in puck editor.

<img width="272" alt="Image" src="https://github.com/user-attachments/assets/7e8778ab-5179-4c38-9664-5d16fd5bb49a" />

FedericoBonel:
#634 included comments requesting a feature to allow the user to rename each outline item for large payloads, where the component name itself could be repeated multiple times and become more confusing than helpful as an identifier in the UI. 

It also suggested allowing the developer to override how this item is presented to the user with an override like this:  

```tsx
const overrides = {
  // These "props" would hold the props given to the component the item represents. 
  // Here I'm assuming there is a title.
  outlineItem: ({ props }) => (
    <Puck.OutlineItem>{props.title}</Puck.OutlineItem>
  ),
};
```

muchisx:
Here to chime in a great UX example of how dnd could be implemented in the outline, Shopify's Theme editor takes the crown for the ones I've used in the past:

## Key highlights: 

### Zoom out view when moving page sections (maybe could be enabled via an API in Puck, something that signals that dragging this item zooms out the viewport)

https://github.com/user-attachments/assets/098ec99d-16e9-428c-abb2-8b6bdc2b6348

### Reordering slots of a parent component, and also changing their parent

https://github.com/user-attachments/assets/ff4a8838-b802-47d8-8b3f-74fa180c7f2c

### Folding composable components

https://github.com/user-attachments/assets/5dfcba49-fd7b-4283-bfaa-1020fe467a61

### Add a component via the outline through a popover. Also smart "Add component" buttons that show up for the root outline but also for empty slots

[...]

PR #1317, opened 2025-09-15
#730 outline menu

conradkirschner:
At current point this is just a design preview of possible future changes on the outline menu - very early preview. 

It tries to implement the solution that was proposed from @FedericoBonel in the issue. 
https://github.com/puckeditor/puck/issues/730#issuecomment-3222758191


FedericoBonel:
Closing this as stale! If you want you can continue to work on this and re-open it.

PR #1331, opened 2025-09-24
Support defaultItemProps as a function

### LINKED ISSUE #1366: Support Function-based defaultItemProps dynamic  for ArrayField
## Description

Currently, the `array` field only supports `defaultItemProps` as a static object to provide default values for new items added to the array. In many scenarios, we need to generate dynamic default values based on the current state of the array, particularly in the following situations:

- **React List Rendering Requirements**: Generate unique `key` attributes for each array item to avoid React rendering warnings and performance issues
- **Order Management**: Automatically set `order` or `index` properties based on current array length to ensure new items are arranged in correct order
- **Unique Identifier Generation**: Create unique `id` based on timestamp, UUID, or index to avoid data conflicts
- **Default Naming**: Generate sequential names like "Item 1", "Item 2" to improve user experience
- **Avoiding Extra Renders**: Without this feature, default initialization data would need to be populated during component creation and rendering, leading to unnecessary React state updates and renders, affecting performance

## Considerations

- This feature affects the ArrayField component located at `packages/core/components/AutoField/fields/ArrayField/index.tsx`
- ArrayField type definitions are located at `packages/core/types/Fields.ts`
- This enhancement should maintain backward compatibility with existing static `defaultItemProps` configurations

### Proposal 1

Extend the `defaultItemProps` property to accept a function that receives the current array length as a parameter and returns the default properties object.

```tsx
const config = {
  components: {
    NavigationItems: {
      fields: {
        items: {
          type: "array",
          arrayFields: {
            title: { type: "text" },
            order: { type: "number" },
            id: { type: "text" },
          },
          defaultItemProps: (currentLength) => ({
            title: `Item ${currentLength + 1}`,
            order: currentLength,
            id: `item-${currentLength}`,
          }),
        },
      },
    },
  },
};
```

Resolves  #1331

LengYXin:
Adds support of defaultItemProps as a function so we can generate dynamic item defaults based on the current array index

Closes #1366

### DIFF
--- a/apps/docs/pages/docs/api-reference/fields/array.mdx
+++ b/apps/docs/pages/docs/api-reference/fields/array.mdx
@@ -213,7 +213,7 @@ const config = {
       return (
         <ul>
           {items.map((item, i) => (
-            <li key={item.key ?? i}>
+            <li key={i}>
               {item.title} (Order: {item.order})
             </li>
           ))}


apps/docs/pages/docs/api-reference/fields/array.mdx
```diff
@@ -140,6 +140,31 @@ const config = {
 };
 ```
 
+You can also use a function to generate dynamic defaults based on the array index:
```
FedericoBonel: Can you move this example underneath `You can also use a function to generate dynamic defaults:` on line 195, right before the config preview, so that the code and the example are together?

packages/core/bundle/core.ts
```diff
@@ -14,7 +14,7 @@ export { Drawer } from "../components/Drawer";
 
 export { DropZone } from "../components/DropZone";
 export * from "../components/IconButton";
-export { Puck } from "../components/Puck";
+export { Puck,usePropsContext } from "../components/Puck";
```
FedericoBonel:
```suggestion
export { Puck } from "../components/Puck";
```
We don't want to expose `usePropsContext` at this time. Open to discussing this in a follow-up feature request ticket 👍, but it should not be included in this PR.

packages/core/components/AutoField/fields/ArrayField/index.tsx
```diff
@@ -405,8 +405,13 @@ export const ArrayField = ({
 
                 const existingValue = value || [];
 
+                // Support defaultItemProps as a function so we can generate dynamic defaults based on the current length of the array
+                const defaultProps = typeof field.defaultItemProps === 'function'
```
FedericoBonel: Could you run `yarn format`? 

There are single quotes and other code styles that don't match the rest of the codebase and will prevent it from being built and merged 😊.

packages/core/components/Puck/index.tsx
```diff
@@ -576,3 +576,4 @@ Puck.Components = Components;
 Puck.Fields = Fields;
 Puck.Outline = Outline;
 Puck.Preview = Preview;
+Puck.Canvas = Canvas;
```
FedericoBonel:
```suggestion

```
We also don't want to expose `Puck.Canvas` at this time.

FedericoBonel (changes requested):
Hey @LengYXin!


For the record, we normally like to have some kind of feature request or bug report for opening PRs so that we might understand what's the need and discuss the best approach

This could a nice feature to have

packages/core/bundle/core.ts
LengYXin: Can you explain why? The reason for making it public is that in practical applications, people like to customize the interface while still wanting to reuse some functionalities. By reading the source code, these happen to match the customization needs; otherwise, you would have to copy it yourself.

packages/core/components/Puck/index.tsx
LengYXin: Same as using usePropsContext

apps/docs/pages/docs/api-reference/fields/array.mdx
LengYXin: Sure, I can update it uniformly. Before that, could you please answer the following questions? Thank you very much.

packages/core/components/Puck/index.tsx
LengYXin: image
This is our design draft. You can see that many parts are actually common, but Puck only exported some of the APIs.
The need for custom-built interfaces is probably unavoidable for most application development.

packages/core/bundle/core.ts
FedericoBonel: Absolutely! First off, let me suggest taking a look at `createUsePuck` and the `useGetPuck` hook. These expose the main APIs you'd need to build a custom interface that can interact with the Puck context.

That being said, there are important considerations when exposing any internal API, this ties directly into API design.

Whenever you expose something, you need to take into account all the people using Puck and their various use cases (some of them do pretty unique stuff). Among other things, we have to consider:

- **Value and usability:** What is the purpose of the API internally? How would it work externally, and for what use cases? Can we simplify or improve it to better serve that need?
- **Encapsulation:** Could this API expose or interfere with internal data or flows in a way that breaks assumptions elsewhere in the codebase? Internal APIs are often designed with the expectation that they won't be used outside our controlled environment
- **Communication:** Every exposed API is a form of communication. Just because something is structured a certain way internally doesn't mean it should be exposed that way externally. We might need a different name, structure, or behavior since it all shapes how people use it, and might unintentionally encourage patterns we want to avoid. It could also be a lot of work to properly document.

In short, we need to consider all of this before exposing new APIs. It might be that your need could be met with a simpler or more targeted API instead.

Opening a feature request that describes your specific need would really help. That gives us the context to evaluate the use case, see if others have similar needs, and decide the best path forward.

packages/core/components/Puck/index.tsx
FedericoBonel: The main thing that `Canvas` has that `Preview` doesn't is the viewports. We have an issue tracking that [here](https://github.com/puckeditor/puck/issues/1227). Feel free to jump in there with ideas, but again, this isn't something for this PR.

packages/core/bundle/core.ts
LengYXin: Thank you for your patient answer. It's my fault for not considering everything comprehensively enough. Learn from you

packages/core/components/Puck/index.tsx
LengYXin: Thank you. I've learned about this. I'll update these submissions next

packages/core/bundle/core.ts
FedericoBonel: That's okay, It's not anyone's fault 😊! It's fine to ask questions.

Thanks for pushing the required changes

apps/docs/pages/docs/api-reference/fields/array.mdx
```diff
@@ -167,6 +167,63 @@ const config = {
...
+    Example: {
+      fields: {
+        items: {
+          type: "array",
+          arrayFields: {
+            title: { type: "text" },
+            order: { type: "number" },
+          },
+          defaultItemProps: (index) => ({
+            key: `item-${index + 1}`,
```
FedericoBonel:
```suggestion
```
I don't think we need this in the example, it could confuse people.

apps/docs/pages/docs/api-reference/fields/array.mdx
```diff
@@ -167,6 +167,63 @@ const config = {
...
+          order: index + 1,
+        }),
+      },
+    },
+    defaultProps: { items: [] },
+    render: ({ items }) => {
+      return (
+        <ul>
+          {items.map((item, i) => (
+            <li key={item.key ?? i}>
```
FedericoBonel:
```suggestion
            <li key={i}>
```
The key field is not defined, so this will never return `item.key`

Also, keys could collide if you define them using a field like this. For example: if you add two items, reorder the second item to the first position, delete the second item, and then add a new one; both the first and second items would end up with the same key because the array had the same length when it generated it.

PR #1341, opened 2025-09-30
feat: add "richtext" field

xaviemirmon:
Add rich text support via the `richtext` field, built with tiptap.

## Usage

```ts
const config = {
  components: {
    Example: {
      fields: {
        body: {
          type: "richtext",
        },
      },
      render: ({ body }) => body,
    },
  },
};
```

Full usage in docs: https://puck-docs-git-149-add-rich-text-support-wi-af6474-puck-9db9778b.vercel.app/docs/api-reference/fields/richtext

## Features

- New `richtext` field
  - Inline mode via `contentEditable`
  - Custom menus via `renderMenu`, `renderInlineMenu` and `<RichTextMenu>`
  - Custom rich text behavior via `options`
  - Extension of field via `tiptap.extensions` and `tiptap.selector`
  - Full documentation of all features, including Integration Guide and API references
- New ActionBar.Separator component
- New `disabled` option on ActionBar.Action
- Add `active` API to IconButton
- Switch style of focused portal outline to dashed
- Fix a bug where replace would not consider the `ui` prop

## Outstanding

- [x] Establish bundling and code splitting strategy
- [x] Finalise type strategy for custom extensions
- [x] Fix RSC

## Follow-ons

- Add link control


packages/core/types/Fields.ts
```diff
@@ -45,6 +51,14 @@ export type RadioField = BaseField & {
   options: FieldOptions;
 };
 
+export type RichtextField = BaseField & {
+  type: "richtext";
+  placeholder?: string;
+  contentEditable?: boolean;
+  config?: RichTextConfigType;
+  extensions?: Extensions[];
+};
```
chrisvxd: Note to remember to document these

packages/core/types/index.ts
```diff
@@ -6,3 +6,4 @@ export * from "./Data";
 export * from "./Fields";
 export * from "./Props";
 export * from "./Utils";
+export * from "./RichTextEditor";
```
chrisvxd: We should consider whether or not it is necessary to export all these types, as types significantly increase API surface area and risk of breaking changes

packages/core/types/RichTextEditor.tsx
```diff
@@ -0,0 +1,47 @@
...
+import { defaultExtensions } from "../components/RichTextEditor/extensions";
+import { useSyncedEditor } from "../components/RichTextEditor/lib/use-synced-editor";
+
+// Generic icon type (Lucide / React component)
+type IconType = ComponentType<any>;
+
+// Base menu item
+export type RichTextMenuItem = {
+  title: string;
+  icon: IconType;
```
chrisvxd: Elsewhere we support a ReactNode instead of component: https://puckeditor.com/docs/api-reference/components/field-label#icon

packages/core/tsconfig.json
```diff
@@ -3,6 +3,7 @@
   "include": ["."],
   "exclude": ["dist", "build", "node_modules"],
   "compilerOptions": {
-    "types": ["node", "jest", "@testing-library/jest-dom"]
+    "types": ["node", "jest", "@testing-library/jest-dom"],
+    "moduleResolution": "bundler"
```
chrisvxd: What was the reason for changing this?

packages/core/package.json
```diff
@@ -95,6 +95,10 @@
   "dependencies": {
     "@dnd-kit/helpers": "0.1.18",
     "@dnd-kit/react": "0.1.18",
+    "@tiptap/extension-text-align": "^3.6.2",
+    "@tiptap/extension-text-style": "^3.6.2",
+    "@tiptap/react": "^3.6.2",
+    "@tiptap/starter-kit": "^3.6.2",
```
chrisvxd: There's a question mark for me around how people will be expecting this to work with existing tiptap setups, and if we should consider peer dependencies 🤔

packages/core/components/DropZone/index.tsx
```diff
@@ -528,10 +531,16 @@ const DropZoneRenderItem = ({
     [props]
   );
 
+  const richTextRenderer =
+    "richtext" in props
```
chrisvxd: Doesn't this rely on the field name being `"richtext"`?

I imagine we should instead be scanning the config for any fields of `type === "richtext"`

packages/core/components/RichTextEditor/Editor.tsx
```diff
@@ -0,0 +1,44 @@
...
+    });
+
+    if (!editor) return null;
+
+    return (
+      <>
+        {!readOnly && (
+          <MenuBar menuConfig={config.menu || {}} editor={editor} />
+        )}
+        <EditorContent editor={editor} className="rich-text-editor" />
```
chrisvxd: There's a mixture of CSS styles going on here. We tend to use CSS modules (where possible) to isolate styles and discourage theming abuse (which leads to breaking API changes, and we'd rather solve this properly via #139).

I would probably expect us to be using the class factory here, with a CSS module. If you need to define global styles within a CSS module (such as for ProseMirror), you can use the `:global()` selector

packages/core/components/RichTextEditor/InlineEditor.tsx
```diff
@@ -0,0 +1,37 @@
...
+    extensions,
+    editable: true,
+  });
+
+  if (!editor) return null;
+
+  return (
+    <>
+      <InlineMenu editor={editor} menuConfig={config.inlineMenu || {}} />
+      <EditorContent editor={editor} className="rich-text-editor-inline" />
```
chrisvxd: I like the reuse of Editor content here - smart!

packages/core/components/RichTextEditor/Render.tsx
```diff
@@ -0,0 +1,17 @@
...
+
+  if (typeof content === "string") {
+    // fallback: assume HTML
+    html = content;
+  } else {
+    // assume JSON
+    html = generateHTML(content, defaultExtensions);
+  }
+
+  return <div dangerouslySetInnerHTML={{ __html: html }} />;
```
chrisvxd: We should sanitise this

packages/core/components/RichTextEditor/Render.tsx
```diff
@@ -0,0 +1,17 @@
+import { generateHTML } from "@tiptap/react";
+import { defaultExtensions } from "./extensions";
+import "./editor-styles.css";
+
+export function Render({ content }: { content: any }) {
```
chrisvxd: Would be good to tighten this type

packages/core/components/RichTextEditor/components/BlockStyleSelect.tsx
```diff
@@ -0,0 +1,52 @@
+import { useEditorState, Editor as EditorType } from "@tiptap/react";
+import { HeadingLevel, RichTextConfigType } from "../../../types";
+import styles from "../styles.module.css";
+import getClassNameFactory from "../../../lib/get-class-name-factory";
+const getClassName = getClassNameFactory("Input", styles);
+export const BlockStyleSelect = ({
```
chrisvxd: Nit: can we break these out with new lines between imports, getClassName and the component for readability?

packages/core/components/RichTextEditor/components/InlineMenu.tsx
```diff
@@ -0,0 +1,72 @@
...
+import { renderButtons } from "../lib/render-buttons";
+import {
+  RichTextConfigType,
+  RichTextEditor,
+  RichTextMenuItem,
+} from "../../../types";
+import { Loader } from "../../Loader";
+import { buildEditorState } from "../lib/build-editor-state";
+import getClassNameFactory from "../../../lib/get-class-name-factory";
+const getClassName = getClassNameFactory("Editor", styles);
```
chrisvxd:
```suggestion
import getClassNameFactory from "../../../lib/get-class-name-factory";

const getClassName = getClassNameFactory("Editor", styles);
```

packages/core/components/RichTextEditor/components/InlineMenu.tsx
```diff
@@ -0,0 +1,72 @@
...
+      return buildEditorState(editor, menuConfig);
+    },
+  });
+  if (!editor || !editorState) {
+    return <Loader />;
+  }
+  if (Object.keys(menuConfig).length === 0) {
+    return null;
+  }
+  return (
```
chrisvxd: Could do with a few more line breaks between these for readability

packages/core/components/RichTextEditor/components/InlineMenu.tsx
```diff
@@ -0,0 +1,72 @@
...
+    <BubbleMenu editor={editor} className={getClassName("button-group")}>
+      {(
+        Object.keys(menuConfig) as Array<keyof RichTextConfigType["inlineMenu"]>
```
chrisvxd: It's probably worth placing the result of Object.keys() in a memo to prevent creating a new reference on each render

packages/core/components/RichTextEditor/components/MenuBar/MenuBar.tsx
```diff
@@ -0,0 +1,66 @@
...
+    selector: (ctx) => {
+      const { editor } = ctx;
+      if (!editor) return null;
+      return buildEditorState(editor, menuConfig);
+    },
+  });
+  if (!editor || !editorState) {
+    return <Loader />;
+  }
+  return (
```
chrisvxd:
```suggestion
  const editorState = useEditorState({
    editor,
    selector: (ctx) => {
      const { editor } = ctx;
      if (!editor) return null;
      return buildEditorState(editor, menuConfig);
    },
  });

  if (!editor || !editorState) {
    return <Loader />;
  }

  return (
```

packages/core/components/RichTextEditor/components/MenuBar.tsx
```diff
@@ -0,0 +1,66 @@
...
+      if (!editor) return null;
+      return buildEditorState(editor, menuConfig);
+    },
+  });
+  if (!editor || !editorState) {
+    return <Loader />;
+  }
+  return (
+    <div className={getClassName("button-group")}>
+      {(Object.keys(menuConfig) as Array<keyof RichTextConfigType["menu"]>).map(
```
chrisvxd: As above, let's use a memo for this to prevent the creation of a new reference on each render

packages/core/components/RichTextEditor/components/MenuBar/index.tsx
```diff
@@ -0,0 +1,66 @@
...
+                  editor
+                )}
+              </div>
+            );
+          }
+        }
+      )}
+    </div>
+  );
+};
```
chrisvxd: What's the difference between this and the InlineMenu? Just the use of BubbleMenu?

packages/core/components/RichTextEditor/config.tsx
```diff
@@ -0,0 +1,142 @@
...
+  Link,
+  Underline,
+  Minus,
+  Quote,
+  SquareCode,
+  Strikethrough,
+} from "lucide-react";
+import { RichTextConfigType, RichTextEditor as Editor } from "../../types";
+
+export const defaultConfig: RichTextConfigType = {
```
chrisvxd: I think this API is one of the most important parts of this implementation. I'm a little wary of adding an additional layer of API abstraction over the tiptap API.

This needs further discussion, but we could consider something closer to this, giving the user a bit more control without introducing a new API:

```tsx
type: "richtext",
renderMenu: ({ editor }: { editor: Editor }) => {
  return <>
    <IconButton onClick=(() => {editor.chain().setTextAlign("left").run()})>
      <AlignLeft />
    </IconButton
  </>
}
```

Or, if we use reuse ActionBar (but would need to add some state support)

```tsx
type: "richtext",
renderMenu: ({ children, editor }: { children: ReactNode, editor: Editor }) => {
  return <ActionBar>
    {children} {/* Render existing actions */}
    
    <ActionBar.Group>
      <ActionBar.Action onClick=(() => {editor.chain().setTextAlign("left").run()})>
        <AlignLeft />
      </ActionBar.Action>
    </ActionBar.Group>
  </ActionBar>
}
```

I'm not clear if this is the right approach, and likely need to experiment with the implementation myself to get my head around it, but I still wanted to use this as an opportunity to discuss the API.

packages/core/components/RichTextEditor/editor-styles.css
```diff
@@ -0,0 +1,54 @@
...
+  border-radius: 0.2em;
+}
+
+p:empty::before {
+  content: "\00a0"; /* non-breaking space */
+}
+
+* {
+  white-space: pre-wrap;
+}
```
chrisvxd: These are global styles and will affect everything in the page. At the very least, we need to scope these underneath `.rich-text-editor`. We should also probably drop these from rendering within `inline` entirely - the user shouldn't need to depend on Puck styles when rendering the final page.

packages/core/components/RichTextEditor/lib/render-buttons.tsx
```diff
@@ -0,0 +1,24 @@
...
+    <IconButton
+      key={title}
+      onClick={(e) => {
+        e.stopPropagation();
+        action(editor);
+      }}
+      disabled={can ? !editorState[title + "Can"] : false}
+      variant={
+        state ? (editorState[title] ? "secondary" : "primary") : "primary"
+      }
```
chrisvxd: We could consider adding an active state to this

packages/core/components/RichTextEditor/lib/build-editor-state.ts
```diff
@@ -0,0 +1,18 @@
...
+  editor: Editor,
+  menuConfig: RichTextConfigType["menu"]
+) {
+  const state: Record<string, boolean> = {};
+
+  Object.values(menuConfig)
+    .flat()
+    .forEach((item: any) => {
+      if (item.state) state[item.title] = item.state(editor) ?? false;
+      if (item.can) state[item.title + "Can"] = item.can(editor) ?? false;
```
chrisvxd: I find this "Can" state stuff a bit confusing. Is this a fair comparison? `.state()` is active state, and `.can()` is whether the control is currently enabled?

packages/core/components/RichTextEditor/styles.module.css
```diff
@@ -0,0 +1,86 @@
...
+.Input--readOnly > .Input-input,
+.Input--readOnly > select.Input-input {
+  background-color: var(--puck-color-grey-11);
+  border-color: var(--puck-color-grey-09);
+  color: var(--puck-color-grey-04);
+  cursor: default;
+  opacity: 1;
+  outline: 0;
+  transition: none;
+}
```
chrisvxd: I didn't expect to see these `.Input` styles here, as they impact a component outside of the `RichTextEditor` scope. Generally, the styles are co-located with the component of the same name (although, the line is a little blurry with fields).

Are these changes specifically for the RTE?

packages/core/components/RichTextEditor/styles.module.css
```diff
@@ -0,0 +1,86 @@
...
+  font-family: inherit;
+  font-size: 12px;
+  padding: 4px 8px;
+  padding-right: 20px;
+  transition: border-color 50ms ease-in;
+}
+
+select.Input-input {
+  appearance: none; /* Safari */
+  background: url("data:image/svg+xml;utf8,<svg xmlns='http://www.w3.org/2000/svg' width='100' height='100' fill='%23c3c3c3'><polygon points='0,0 100,0 50,50'/></svg>")
```
chrisvxd: What's this guy?

packages/core/lib/field-transforms/default-transforms/rich-text-transform.tsx
```diff
@@ -0,0 +1,96 @@
...
+  }) => {
+    const portalRef = useRef<HTMLDivElement>(null);
+    const appStoreApi = useAppStoreApi();
+
+    // Register portal once
+    useEffect(() => {
+      if (!portalRef.current) return;
+      const cleanup = registerOverlayPortal(portalRef.current);
+      return () => cleanup?.();
+    }, []);
```
chrisvxd: Don't want to update if the ref changes?

packages/core/lib/field-transforms/default-transforms/rich-text-transform.tsx
```diff
@@ -0,0 +1,96 @@
...
+
+InlineEditorWrapper.displayName = "InlineEditorWrapper";
+
+export const getRichTextTransform = (): FieldTransforms => ({
+  richtext: ({ value, componentId, field, propPath, isReadOnly }) => {
+    const { contentEditable, config, extensions } = field;
+    if (contentEditable === false || isReadOnly) {
+      return <Render content={value} />;
+    }
+    return (
```
chrisvxd:
```suggestion
   const { contentEditable, config, extensions } = field;

    if (contentEditable === false || isReadOnly) {
      return <Render content={value} />;
    }

    return (
```

chrisvxd (changes requested):
Epic! Left some initial thoughts. Some general tidy ups, but main area I think we need to discuss is the public API.

I wasn't able to test it yet due to failing build.

packages/core/tsconfig.json
xaviemirmon: So that @tiptap/react/menus resolves.

packages/core/components/RichTextEditor/lib/build-editor-state.ts
xaviemirmon: State:  whether or not the selected text is active e.g. bold is selected
Can: whether or not the action can be performed by the loaded extensions

packages/core/components/RichTextEditor/styles.module.css
xaviemirmon: The heading selector dropdown

packages/core/components/RichTextEditor/styles.module.css
xaviemirmon: This is all for the heading selector dropdown

packages/core/components/DropZone/index.tsx
xaviemirmon: I am looking at field type now

packages/core/components/RichTextEditor/components/BlockStyleSelect.tsx
xaviemirmon: I've fixed the formatting here now.

packages/core/components/RichTextEditor/Editor.tsx
xaviemirmon: Fixed

packages/core/types/index.ts
xaviemirmon: Moved some types to internal scope

packages/core/components/RichTextEditor/styles.module.css
xaviemirmon: I've moved to BlockStyleSelect

packages/core/components/RichTextEditor/components/MenuBar/index.tsx
xaviemirmon: I've now refactored InlineMenu so that it just wraps the BubbleMenu around MenuBar and updates styles

packages/core/components/RichTextEditor/Render.tsx
xaviemirmon: Not required

packages/core/components/RichTextEditor/config.tsx
xaviemirmon: Are you happy with `BubbleMenu` or should I look at using `ActionBar` instead?

FedericoBonel:
This is awesome! Super glad to see this feature finally being added 🙏, thank you.

I tested it on my machine and here are my thoughts:

- Clicking on the field inline deselects the component. This is inconsistent with the other inline field, and inconvenient because the menu in the sidebar provides more options, so I have to constantly click just outside the inline field area to reselect the component and access those options. Also it does feel a bit weird as a user, I would expect the component to get selected when I click on it.

https://github.com/user-attachments/assets/81146f6c-9f3a-49dd-80ef-65cfdbeb41a1

- Clicking on the menu options currently deselects the text, both inline and in the sidebar. This makes it a bit tricky to apply multiple options since I have to keep reselecting the text each time. 

https://github.com/user-attachments/assets/d0a358f1-2599-45d8-b888-7eb1f7de8e22

- Editing inline adds some extra spacing on the left that isn’t there when you’re not editing inline.

https://github.com/user-attachments/assets/e8c6cdc1-45e7-467d-815c-8e4e718cd8f9

- Pressing Ctrl + I triggers interactive mode instead of applying italics.

https://github.com/user-attachments/assets/96aa82de-e256-4dc3-b4ed-efc881a8f5ad

- When editing inline, selecting text and trying to apply styling from the sidebar menu doesn’t affect the selected text. It seems the sidebar options are tied only to the sidebar field. As a user, I’d expect the selected text, no matter where it is, to be updated when I apply a style. I also think people might want to edit inline, but use the sidebar options since there are more options there and inline editing feels more intuitive.

https://github.com/user-attachments/assets/a2819ca1-0a22-4b9f-8ec7-e85d74ce5b96

- I’d expect the sidebar field to behave more like a textarea that I can resize, or at least expand up to a certain point and then scroll.  Right now, if I write a long article in the RTE, the field becomes very tall, and any fields below it require a lot of scrolling to reach. It also makes applying formatting options difficult since I have to scroll all the way up to use the toolbar and then all the way down to keep editing (check the video). To fix this, I think we could set a max height so that the field is never taller than the viewport.

https://github.com/user-attachments/assets/0fe73cfb-5004-49e3-8bb0-c721b9afea49

- If you resize the sidebar to its maximum width and keep dragging, the field overflows past its container.

https://github.com/user-attachments/assets/adce4f7c-f6a1-4026-98c0-91794b858b8d

- Code block styling doesn’t look quite right on new lines, the left margin disappears, making it look misaligned with the first line. Also, if you add two new empty lines, the code block breaks with a gap.

https://github.com/user-attachments/assets/b2f91712-e1e5-48ef-ad35-48658746cf27

apps/demo/config/blocks/RichText/index.tsx
```diff
@@ -0,0 +1,25 @@
...
+    richtext: {
+      type: "richtext",
+    },
+  },
+  render: ({ richtext }) => {
+    return <Section>{richtext}</Section>;
+  },
+  defaultProps: {
+    richtext:
+      "<p style='text-align: center;'><strong>✍️ <s>Plain-text</s> Rich Text Editor! 🎉</strong></p>",
```
chrisvxd: I think we should use something a bit simpler as the default. Like

```
## Heading

Body
```

apps/demo/config/initial-data.ts
```diff
@@ -127,6 +127,20 @@ export const initialData: Record<string, UserData> = {
           direction: "vertical",
         },
       },
+      {
+        type: "RichText",
+        props: {
+          richtext:
+            "<p style='text-align: center;'><strong>✍️ <s>Plain-text</s> Rich Text Editor! 🎉</strong></p>",
```
chrisvxd: Let's pull this from the demo page unless we can find a real use-case for it. The demo page gets decent SEO and so it needs to actually make some sense as a piece of content

apps/demo/tsconfig/base.json
```diff
@@ -9,7 +9,7 @@
     "forceConsistentCasingInFileNames": true,
     "inlineSources": false,
     "isolatedModules": true,
-    "moduleResolution": "node",
+    "moduleResolution": "bundler",
```
chrisvxd: These changes still concern me, as I don't want to cause bundling issues for our users. Would like to see if there's another way to handle it.

apps/docs/pages/docs/api-reference/fields/richtext.mdx
```diff
@@ -0,0 +1,733 @@
...
+import { Callout } from "nextra/components";
+import Superscript from "@tiptap/extension-superscript";
+import { SuperscriptControl } from "@/docs/components/Richtext/controls";
+import { superscriptSelector } from "@/docs/components/Richtext/selector";
+
+# RichText
+
+Render a `richtext` input. Extends [Base](base).
+
+<ConfigPreview
```
chrisvxd, line 10: The inline preview for this isn't actually rendering on the right hand side

image

apps/docs/pages/docs/api-reference/fields/richtext.mdx
```diff
@@ -0,0 +1,733 @@
...
+      richtext:
+        "<p style='text-align: center;'><strong>✍️ Hello,<s>Plain-text</s> Rich Text! 🎉</strong></p>",
+    },
+  }}
+/>
+
+<Callout>
+  `defaultProps` currently supports an `html` string value. The editor renders
+  the content as `html`.
+</Callout>
```
chrisvxd: It's confusing to me that the field is creating JSON here (visible under "Show output"), but defaultProps receives HTML. Field data should be identical to the default prop data.

Are you saying defaultProps support an HTML string _or_ the JSON generated by the field?

if we need this call out, it would be better placed after the code snippet.

apps/docs/pages/docs/api-reference/fields/richtext.mdx
```diff
@@ -0,0 +1,733 @@
...
+| [`inlineMenu`](#inlineMenu)               | `inlineMenu: { typography: ["Bold", "Italic", "Underline", "Strikethrough"] }`      | `Record<string,string[]>`                                      | -        |
+| [`controls`](#controls)                   | `controls: { MyControl: {render: (editor, editorState) => <div>My Control</div>} }` | `Record<string, {render: (editor, editorState) => ReactNode}>` | -        |
+| [`textSelectOptions`](#textselectoptions) | `textSelectOptions: ["p","h2"]`                                                     | `("p" \| "h2" \| "h3" \| "h4" \| "h5" \| "h6")[]`              | -        |
+| [`extensions`](#extensions)               | `extensions: [StarterKit, Link]`                                                    | `Extensions[]` <i>from TipTap</i>                              | -        |
+| [`selector`](#selector)                   | `selector: (ctx) => ({ ... })`                                                      | `RichTextSelector`                                             | -        |
```
chrisvxd: Some of these examples can be stripped back, and we don't need to provide full TS types.

They're just meant to be indicative.

I'm happy to make some changes here.

apps/docs/pages/docs/api-reference/fields/richtext.mdx
```diff
@@ -0,0 +1,733 @@
...
+Enable inline text editing for this field.
+
+```tsx {7} copy
+const config = {
+  components: {
+    Example: {
+      fields: {
+        description: {
+          type: "richtext",
+          contentEditable: false,
```
chrisvxd:
```suggestion
          contentEditable: true,
```

apps/docs/pages/docs/api-reference/fields/richtext.mdx
```diff
@@ -0,0 +1,733 @@
...
+    },
+  },
+};
+```
+
+## Optional params
+
+### `contentEditable`
+
+Enable inline text editing for this field.
```
chrisvxd:
```suggestion
Enable inline text editing for this field. Defaults to `false`.
```

apps/docs/pages/docs/api-reference/fields/richtext.mdx
```diff
@@ -0,0 +1,733 @@
...
+      },
+      // ...
+    },
+  },
+};
+```
+
+### `menu`
+
+Provide an override for the grouped button layout rendered in the editor toolbar. Each key in your `menu` matches a menu group and maps to an array of control keys.
```
chrisvxd:
```suggestion
Configure the toolbar menu layout for the available [controls](#controls). Items are placed into groups.
```

apps/docs/pages/docs/api-reference/fields/richtext.mdx
```diff
@@ -0,0 +1,733 @@
...
+        },
+      },
+    },
+  },
+};
+```
+
+### `inlineMenu`
+
+Provide an override for the grouped button layout rendered in the `BubbbleMenu` when inline editing of your field. Each key in your `inlineMenu` matches a menu group and maps to an array of control keys.
```
chrisvxd:
```suggestion
Configure the toolbar menu layout for the available [controls](#controls) shown inline when enabling [`contentEditable`](#contenteditable). Items are placed into groups.
```

apps/docs/pages/docs/api-reference/fields/richtext.mdx
```diff
@@ -0,0 +1,733 @@
...
+        },
+      },
+    },
+  },
+};
+```
+
+### `controls`
+
+Controls let you swap or extend the buttons available to each menu key. A control is an object with a `render(editor, editorState)` function, so you can plug in any React UI while still receiving the merged TipTap state produced by our selector utilities.
```
chrisvxd: The control object needs breaking down into another level of docs to cover the render function and its args.

Additionally, I suggest the following for this:

```suggestion
Controls are actions shown in the [`menu`](#menu) and [`inlineMenu`](#inlinemenu). You can define your own controls, or override the [default ones](#default-controls).
```

apps/docs/pages/docs/api-reference/fields/richtext.mdx
```diff
@@ -0,0 +1,733 @@
...
+
+```tsx {1-12, 20-22} copy
```
chrisvxd: Line numbers are incorrect

apps/docs/pages/docs/api-reference/fields/richtext.mdx
```diff
@@ -0,0 +1,733 @@
...
+      },
+    },
+  },
+};
+```
+
+<Callout>
+  Note: The `e.stopPropagation();` is required in Puck in order to avoid issues
+  with the portal on inlineMenu.
+</Callout>
```
chrisvxd: This could be more informative (and I personally cannot remember why this is necessary)

apps/docs/pages/docs/api-reference/fields/richtext.mdx
```diff
@@ -0,0 +1,733 @@
...
+            blocks: ["BulletList", "OrderedList"],
+          },
+        },
+      },
+    },
+  },
+};
+```
+
+### `controls`
```
chrisvxd: Suggest we need a list of "Default controls" underneath this, rather than under "Defaults"

apps/docs/pages/docs/api-reference/fields/richtext.mdx
```diff
@@ -0,0 +1,733 @@
...
+          type: "richtext",
+          selector: highlightSelector,
+        },
+      },
+    },
+  },
+};
+```
+
+## Default values
```
chrisvxd: I think this section is generally unnecessary, and most defaults can be folded into the parameters. We don't do this elsewhere and the level of detail is a bit on the verbose side.

apps/docs/pages/docs/api-reference/fields/richtext.mdx
```diff
@@ -0,0 +1,733 @@
...
+  },
+};
+```
+
+<Callout>
+  Note: The `e.stopPropagation();` is required in Puck in order to avoid issues
+  with the portal on inlineMenu.
+</Callout>
+
+### `textSelectOptions`
```
chrisvxd: The user is able to set tags regardless what value is provided to textSelectOptions.

For example, using markdown, the user can set an H1. But because "textSelectOptions" doesn't contain an H1, it shows "Paragraph".

I haven't tested it, but I imagine that this might extend beyond this API. I.e. if the user sets bold but there is no bold control.

We likely need to tweak the API here. As a developer, I would expect to be able to control exactly what tags are available in my RTE. I would also expect Puck to automatically show the controls (and adjust values in the text select) based on what tags I allow.

So something like this: `validTags: ["h1", "h2", "p", "b"]`.

apps/docs/pages/docs/api-reference/fields/richtext.mdx
```diff
@@ -0,0 +1,733 @@
...
+<Callout type="info">
+  TipTap maintains extensive [extension
+  guides](https://tiptap.dev/docs/editor/extensions/overview). Any configuration
+  recommended there works inside your `extensions` array and via `controls`,
+  `selector` + `menu`/`inlineMenu` overrides.
+</Callout>
+
+### `selector`
+
+Pass a custom selector `(ctx) => ({ ... })` to add booleans into the merged editor state. This is useful when supplying bespoke controls that need to react to TipTap attributes or mark status beyond what the defaults provide.
```
chrisvxd: I think it's not clear this is a TipTap API

apps/docs/pages/docs/api-reference/fields/richtext.mdx
```diff
@@ -0,0 +1,733 @@
...
+
+### `inlineMenu`
+
+```tsx
+{
+  text: ["Bold", "Italic", "Underline", "Strikethrough"],
+}
+```
+
+### `controls`
```
chrisvxd: These aren't necessary to document in such detail. We just need to provide a human-friendly list of the available controls, not internal code snippets.

apps/docs/pages/docs/api-reference/fields/richtext.mdx
```diff
@@ -0,0 +1,733 @@
...
+
+    isBlockquote: editor.isActive("blockquote"),
+    canBlockquote: editor.can().chain().toggleBlockquote().run(),
+
+    canHorizontalRule: editor.can().chain().setHorizontalRule().run(),
+  };
+};
+```
+
+## Extending the Editor
```
chrisvxd: These usage guidelines should be in "Integrating/Extending Puck", not in the API reference.

packages/core/components/DropZone/index.tsx
```diff
@@ -47,7 +48,9 @@ import { expandNode } from "../../lib/data/flatten-node";
 import { useFieldTransforms } from "../../lib/field-transforms/use-field-transforms";
 import { getInlineTextTransform } from "../../lib/field-transforms/default-transforms/inline-text-transform";
 import { getSlotTransform } from "../../lib/field-transforms/default-transforms/slot-transform";
+import { getRichTextTransform } from "../../lib/field-transforms/default-transforms/rich-text-transform";
 import { FieldTransforms } from "../../types/API/FieldTransforms";
+import { Render } from "../RichTextEditor/Render";
```
chrisvxd: Can you name this RichTextRender? Render normally refers to the Puck Render component

packages/core/components/DropZone/index.tsx
```diff
@@ -528,10 +546,18 @@ const DropZoneRenderItem = ({
     [props]
   );
 
+  const richtextKey = findRichtextKey(Component.fields);
+
+  const richTextRenderer =
+    richtextKey && richtextKey in props
+      ? { richtext: <Render content={props[richtextKey]} /> }
+      : {};
```
chrisvxd: This object is going to to get re-created on every render and needs memoizing

packages/core/components/DropZone/index.tsx
```diff
@@ -528,10 +546,18 @@ const DropZoneRenderItem = ({
     [props]
   );
 
+  const richtextKey = findRichtextKey(Component.fields);
+
+  const richTextRenderer =
+    richtextKey && richtextKey in props
+      ? { richtext: <Render content={props[richtextKey]} /> }
```
chrisvxd: Shouldn't this be

```suggestion
      ? { [richtextKey]: <Render content={props[richtextKey]} /> }
```

packages/core/components/DropZone/index.tsx
```diff
@@ -528,10 +546,18 @@ const DropZoneRenderItem = ({
     [props]
   );
 
+  const richtextKey = findRichtextKey(Component.fields);
```
chrisvxd: This will only work with a single rich text field. It's possible the user will have multiple.

It also needs memoizing

packages/core/components/RichTextEditor/Render.tsx
```diff
@@ -0,0 +1,64 @@
...
+  content,
+  extensions = [],
+}: {
+  content: string | JSONContent;
+  extensions?: Extensions;
+}) {
+  const loadedExtensions = useMemo<Extensions>(
+    () => [...defaultExtensions, ...extensions],
+    [extensions]
+  );
```
chrisvxd: Hm, so does the user have to provide their tiptap extensions into the rendered user environment?

It's possible they might want to strip out unnecessary internal packages.

packages/core/components/RichTextEditor/lib/use-synced-editor.ts
```diff
@@ -0,0 +1,58 @@
...
+    content,
+    editable,
+    immediatelyRender: false,
+    parseOptions: { preserveWhitespace: "full" },
+  });
+
+  const [debouncedJson, setDebouncedJson] = useDebounce<JSONContent | string>(
+    "",
+    200
+  );
```
chrisvxd: Currently there is a perceived delay when the user modifies the form field, caused by this default.

The debounce behaviour should be tweaked to be leading, with max delay, which should make the field feel more responsive.

packages/core/components/RichTextEditor/styles.module.css
```diff
@@ -0,0 +1,52 @@
...
+  transition: border-color 50ms ease-in;
+  width: 100%;
+  max-width: 100%;
+}
+
+.RichTextEditor--editor :global(.ProseMirror-focused) {
+  border-color: var(--puck-color-grey-05);
+  outline: 2px solid var(--puck-color-azure-05);
+  transition: none;
+}
```
chrisvxd, line 104: Note to self to review the practicality of these

packages/core/package.json
```diff
@@ -95,6 +95,9 @@
   "dependencies": {
     "@dnd-kit/helpers": "0.1.18",
     "@dnd-kit/react": "0.1.18",
+    "@tiptap/extension-text-align": "^3.6.2",
+    "@tiptap/react": "^3.6.2",
+    "@tiptap/starter-kit": "^3.6.2",
```
chrisvxd: We should discuss whether or not to include these as deps or peer deps

packages/core/components/RichTextEditor/config.ts
```diff
@@ -0,0 +1,14 @@
+import { RichTextMenuConfig } from "./types";
+
+export const defaultMenu: RichTextMenuConfig = {
```
chrisvxd: I think we can remove the following controls from the default menu as they are unlikely to be widely applicable. We can keep the underlying default controls so it's easy for people to add them.

1.  hr
2. inline code
3. code block
4. block quote

apps/demo/config/blocks/RichText/index.tsx
xaviemirmon: Done!

apps/demo/config/initial-data.ts
xaviemirmon: I've moved to hero

apps/demo/tsconfig/base.json
xaviemirmon: Removed as I've changed it to the actionBar

packages/core/types/Fields.ts
xaviemirmon: Done!

packages/core/tsconfig.json
xaviemirmon: Removed as we're now using actionBar

packages/core/package.json
xaviemirmon: Added as a peer dependencies!

packages/core/components/RichTextEditor/config.tsx
xaviemirmon: Moved to actionBar

apps/docs/pages/docs/api-reference/fields/richtext.mdx
xaviemirmon: Fixed rendering

apps/docs/pages/docs/api-reference/fields/richtext.mdx
xaviemirmon: Yeah if you wouldn't mind taking a stab at this one that'd be very helpful!

packages/core/components/DropZone/index.tsx
xaviemirmon: This is outdated now

packages/core/components/DropZone/index.tsx
xaviemirmon: This is outdated now

packages/core/components/DropZone/index.tsx
xaviemirmon: This is outdated now

packages/core/components/RichTextEditor/lib/use-synced-editor.ts
xaviemirmon: Tweaked to 50ms

packages/core/package.json
xaviemirmon: Also added peer deps

packages/core/components/RichTextEditor/config.ts
xaviemirmon: Done

apps/docs/pages/docs/api-reference/fields/richtext.mdx
xaviemirmon: Let's dicuss this one!

apps/docs/pages/docs/api-reference/fields/richtext.mdx
xaviemirmon: Don't totally follow.  Lets discuss

apps/docs/pages/docs/api-reference/fields/richtext.mdx
xaviemirmon: I'll remove.

apps/demo/config/initial-data.ts
```diff
@@ -8,7 +8,7 @@ export const initialData: Record<string, UserData> = {
         props: {
           title: "This page was built with Puck",
           description:
-            "Puck is the self-hosted visual editor for React. Bring your own components and make site changes instantly, without a deploy.",
+            "<p><strong>Puck is the self-hosted visual editor for React.</strong> Bring your own components and make site changes instantly, without a deploy.</p>",
```
chrisvxd: I actually think we should drop the `strong` here for now

```suggestion
            "<p><strong>Puck is the self-hosted visual editor for React.</strong> Bring your own components and make site changes instantly, without a deploy.</p>",
```

apps/docs/package.json
```diff
@@ -17,6 +17,7 @@
     "typescript": "^5.5.4"
   },
   "dependencies": {
+    "@tiptap/extension-superscript": "^3.6.5",
```
chrisvxd: Is this in the right place?

packages/core/components/DraggableComponent/index.tsx
```diff
@@ -33,6 +33,7 @@ import { useSortable } from "@dnd-kit/react/sortable";
 import { accumulateTransform } from "../../lib/accumulate-transform";
 import { useContextStore } from "../../lib/use-context-store";
 import { useOnDragFinished } from "../../lib/dnd/use-on-drag-finished";
+import { useActiveEditor } from "../RichTextEditor/context";
```
chrisvxd: this isn't used

packages/core/components/RichTextEditor/Render.tsx
```diff
@@ -0,0 +1,55 @@
...
+      return content;
+    }
+
+    if (typeof content === "string") {
+      const isHtml = /<\/?[a-z][\s\S]*>/i.test(content);
+
+      if (isHtml) {
+        return generateJSON(content, loadedExtensions);
+      }
+
```
chrisvxd: Could be nice to add a markdown case here, or will that just work?

packages/core/components/RichTextEditor/components/InlineMenu/index.tsx
```diff
@@ -0,0 +1,75 @@
...
+      parentAction: ReactNode;
+    }) => (
+      <ActionBar label={label}>
+        <ActionBar.Group>
+          <MenuBar
+            menuConfig={menuConfig}
+            editor={editor}
+            selector={selector}
+            inline
+          />
```
chrisvxd: How easy is it for the user to implement these controls in their own action bar?

apps/docs/pages/docs/integrating-puck/rich-text-editing.mdx
```diff
@@ -0,0 +1,342 @@
...
+  label="Example"
+  componentConfig={{
+    fields: {
+      body: {
+        type: "richtext",
+        options: { bold: false },
+      },
+    },
+    defaultProps: {
+      body: "<p>Hello, <strong>world</strong></p>",
```
xaviemirmon: I am confused by this not sure if it's a gap in my understanding or the docs.  If bold is disabled shouldn't all of bold be disabled not just the control?

apps/docs/pages/docs/api-reference/fields/richtext.mdx
```diff
@@ -0,0 +1,557 @@
...
+  componentConfig={{
+    fields: {
+      body: {
+        type: "richtext",
+        renderMenu: ({ children }) => (
+          <RichTextMenu>
+            {children}
+            <RichTextMenu.Group>
+              <RichTextMenu.Control icon={<Quote />} />
+              <RichTextMenu.Control icon={<Highlighter />} />
```
xaviemirmon: Should highlighter do anything or is it just indicative of adding a control?

apps/docs/components/Richtext/controls.tsx
xaviemirmon, line 1: I think this is safe to remove now?

apps/docs/components/Richtext/selector.ts
xaviemirmon, line 1: I think this is safe to remove now?

apps/docs/package.json
xaviemirmon, line 1: I think this is safe to remove now?

apps/demo/config/blocks/Hero/client.tsx
```diff
@@ -81,7 +81,36 @@ export const Hero: ComponentConfig<{
...
+        return (
+          <RichTextMenu>
+            <RichTextMenu.Group>
+              <RichTextMenu.Bold />
+              <RichTextMenu.Italic />
+              <RichTextMenu.Underline />
+            </RichTextMenu.Group>
+            <RichTextMenu.Group>
+              <RichTextMenu.ListSelect />
+              <RichTextMenu.Control
```
xaviemirmon, line 101: How come we aren't using `<RichTextMenu.Blockquote>` and a custom control instead here?

apps/docs/pages/docs/api-reference/fields/richtext.mdx
```diff
@@ -0,0 +1,557 @@
...
+          body: "<p>Hello, <strong>world</strong></p>",
+        },
+      },
+    ],
+  }}
+>
+  <Puck.Preview />
+</PuckPreview>
+
+### `tiptap.extensions`
```
xaviemirmon, line 364: No one for now but I think it would be good to have an example of everything working together, controls, options, menus, and extensions.

packages/core/components/RichTextMenu/index.tsx
xaviemirmon, line 1: Any reason we're not exporting `ControlSelect`?

packages/core/components/Select/index.tsx
```diff
@@ -0,0 +1,112 @@
...
+          <div>
+            <div className={getClassName("button")}>
+              <span className={getClassName("buttonIcon")}>{children}</span>
+              <ChevronDown size={12} />
+            </div>
+          </div>
+        )}
+
+        {options.length > 0 && (
+          <PopoverPortal>
```
xaviemirmon, line 80: How come the popover and not a home built option?

FedericoBonel:
This is looking great! I really like that we provide utilities while still exposing TipTap's APIs. I also really like the UI and overall experience.

Here are a few small bugs/limitations I noticed:

- If I apply a bullet or ordered list from the sidebar (without selecting the text inline), the changes don't commit until I modify the text content. This seems related to the list menu buttons not updating the page data.

  https://github.com/user-attachments/assets/828e5b54-144a-4f97-baf7-53bbebf0dc98

- Similarly, if I apply alignment from the sidebar (without selecting the text inline), the changes also don't commit until I modify the text content. Might be related to the issue above.

  https://github.com/user-attachments/assets/b13f2cab-5f8d-44b8-95ad-f7d95ae9c3b2

- I'd suggest adding `resize: vertical` to the field container so users can resize it, similar to how native textareas behave. Some browsers don’t support it, but the `max-height` fallback should still apply.

  https://github.com/user-attachments/assets/a7708239-335f-4de5-9cfe-635b56f758a4

- I can't select the rendered text on the final page, is that intentional? Is there a way to toggle that behavior?

  https://github.com/user-attachments/assets/75bd1223-b7f8-464c-8ca9-10d4215b8554

- There's no easy way to type the `renderInlineMenu` or `renderMenu` params when adding new extensions. I think we should provide some types, generics, or module augmentation to help with that.

apps/docs/pages/docs/integrating-puck/rich-text-editing.mdx
chrisvxd: It's a bug. We weren't including the options or extensions in the sanitisation for the rendered output. Good catch!

apps/docs/pages/docs/api-reference/fields/richtext.mdx
chrisvxd: It was just meant to be indicative. I've added an `alert` on click.

apps/docs/pages/docs/integrating-puck/rich-text-editing.mdx
chrisvxd: It's a bug. We weren't including the options or extensions in the sanitisation for the rendered output. Good catch!

apps/docs/pages/docs/api-reference/fields/richtext.mdx
chrisvxd: It was just meant to be indicative. I've added an `alert` on click.

apps/docs/components/Richtext/controls.tsx
chrisvxd: Done!

apps/docs/components/Richtext/selector.ts
chrisvxd: Done!

apps/demo/config/blocks/Hero/client.tsx
chrisvxd: Just to demonstrate the Control working, and ensure the build is working as expected.

apps/docs/package.json
chrisvxd: Done!

apps/demo/config/blocks/Hero/client.tsx
chrisvxd: Just to demonstrate the Control working, and ensure the build is working as expected.

apps/docs/pages/docs/api-reference/fields/richtext.mdx
chrisvxd: There is a much more comprehensive guide under "Integration Guides > Rich Text Editing" under "Adding custom extensions", but I imagine we'll also do some additional content around it

packages/core/components/RichTextMenu/index.tsx
chrisvxd: Just because it's a bit fiddly from an API and docs point of view, and I thought we could see if there's a need and do it later if necessary.

packages/core/components/Select/index.tsx
chrisvxd: I did originally do that but quickly hit issues with overflowing the viewport, and getting clipped by `overflow-x: auto` on the parent. Radix handles all of that by using `fixed` and tracking the position appropriately.

chrisvxd:
@xaviemirmon / @FedericoBonel I've addressed all of your comments. Please let me know if there's anything else.

@FedericoBonel responses to yours

1. Fixed in [2b19517](https://github.com/puckeditor/puck/pull/1341/commits/2b195175d206e16ffbdf85603a88c50269b7167f). 
2. Same bug
3. TIL! Swapped `maxHeight` for `initialHeight` and added resize.
4. Bug, fixed in [3aa27d1](https://github.com/puckeditor/puck/pull/1341/commits/3aa27d1d2587e855d0760bff43aafefbf28e4193)

apps/docs/pages/docs/api-reference/fields/richtext.mdx
FedericoBonel: Chiming in, I followed the guide from the docs, I didn't really have any exposure to the API until this point and I found it clear enough. I do think we'll create some blog posts and videos explaining how to fully customize it to create something like an article editor or similar.

packages/core/components/RichTextMenu/index.tsx
FedericoBonel: I do think it would be really useful for custom controls, but I think we could release this without exposing it at the moment. Might need to be wrapped with an abstraction that makes it simpler.

packages/core/components/Select/index.tsx
FedericoBonel: Ah, popovers are always tricky like this. As a side note, the native [popover](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Global_attributes/popover) is now baseline, but I don't think we need to re-implement this.

packages/core/components/SlotRender/server.tsx
```diff
@@ -35,9 +36,12 @@ const Item = ({
     <SlotRenderPure {...slotProps} config={config} metadata={metadata} />
   )) as WithPuckProps<ComponentData["props"]>;
 
+  const richTextRenderer = useRichtextRenderer(Component.fields, props);
+
   return (
     <Component.render
       {...props}
+      {...richTextRenderer}
```
FedericoBonel: We need to do this in [`ServerRender`](https://github.com/puckeditor/puck/blob/3aa27d1d2587e855d0760bff43aafefbf28e4193/packages/core/components/ServerRender/index.tsx#L63-L76) also, otherwise server-side rendering won't work for top-level rich text fields.

Also, this `Item` component is starting to resemble the top-level item rendering we're doing in `ServerRender` (inside the map) since they are conceptually the same. It might be worth refactoring into a shared component, though that’s probably better left for a separate PR.

FedericoBonel (changes requested):
Awesome, thank you for solving all those bugs. It's looking great.

One last one I found out with SSR/RSC.

Wondering if we should create an issue to track small refactor ideas like the one above, we'll end up forgetting.

DamianKocjan:
I spotted a issue while testing this in Firefox.

In Chrome, caret navigation (using arrow keys) and text selection (Shift + Arrow keys) work perfectly within the rich text field. However, in Firefox 145.0.2, these actions don't behave like in Chrome.

FedericoBonel:
Good catch @DamianKocjan! Thanks for reporting it

chrisvxd:
Fixed that issue @DamianKocjan, thanks for reporting.

@FedericoBonel ready for final pass 🎉  canary available at `0.21.0-canary.b353d921`.

FedericoBonel:
Beautiful work ✨

- Firefox issue is fixed now
- Bundle size went down when not using RTE. I think this is good!

PR #1343, opened 2025-10-02
Check for `disableDrag` first instead of `disableDragOnFocus`

### LINKED ISSUE #1283: The `disableDrag` from `registerOverlayPortal` doesn't work when `disableDragOnFocus` is enabled
## Description

If both `disableDrag` and `disableDragOnFocus` are enabled, then only `disableDragOnFocus` is applied, while `disableDrag` is completely ignored.  
This is especially problematic because `disableDragOnFocus` is enabled by default, making it unclear to developers that both flags are active.

## Environment

- Puck version: 0.20.2

## Steps to reproduce

1. Add a component to the Puck config that uses overlay portals and enables `disableDrag`:

```tsx
const config = {
  components: {
    TestBlock: {
      fields: {
        title: {
          type: "text",
        },
        button: {
          type: "text",
        },
      },
      defaultProps: {
        title: "Hello there",
        button: "From portal",
      },
      render: ({ title, button }) => {
        return (
          <div style={{ padding: 16 }}>
            <div
              ref={(ref) => {
                return registerOverlayPortal(ref, {
                  disableDrag: true, // <- Enable this flag only
                });
              }}
              style={{ padding: 20 }}
            >
              <div>
                <h1>{title}</h1>
              </div>
              <button>{button}</button>
            </div>
          </div>
        );
      },
    },
  },
};
```

2. Open the editor  
3. Add the component with the overlay portal to the canvas
4. Try dragging from the portal content  

## What happens

The component can still be dragged around.

## What I expect to happen

The component should **not** be draggable because `disableDrag` is enabled.

## Additional media

https://github.com/user-attachments/assets/97cee173-c73f-4a2b-a5b7-43e6f58c595c

## Why this happens and workaround

- This occurs because of [this `else` check](https://github.com/puckeditor/puck/blob/c6d4e0ecff66f04a79708bd1bd805e0471c729bf/packages/core/lib/overlay-portal/index.tsx#L33).  
  Whenever `disableDragOnFocus` is enabled, only the first branch of the `if` statement executes, and the `disableDrag` handling is ignored.  

- To work around this, you need to explicitly disable `disableDragOnFocus` while keeping `disableDrag` enabled, which forces the `disableDrag` handling to execute:

```tsx
render: ({ title, button }) => {
  return (
    <div style={{ padding: 16 }}>
      <div
        ref={(ref) => {
          return registerOverlayPortal(ref, {
            disableDrag: true,
            disableDragOnFocus: false, // <-- Disable this flag
          });
        }}
        style={{ padding: 20 }}
      >
        <div>
          <h1>{title}</h1>
        </div>
        <button>{button}</button>
      </div>
    </div>
  );
},
```

FedericoBonel:
The `disableDrag` option from `registerOverlayPortal` had no effect if `disableDragOnFocus` was also enabled. This happened because of the order of the `if` checks inside `registerOverlayPortal`: `disableDragOnFocus` always took precedence over `disableDrag`.

## Changes made

- Reordered the checks so that `disableDrag` is evaluated first. If `disableDrag` is enabled, its handling runs. If not, then the `disableDragOnFocus` handling runs.  
- Fixed another bug in the cleanup function: the `onFocus` callback was incorrectly removed for the `blur` event instead of `onBlur`.

## Manual testing

- Enabling the `disableDrag` option in `registerOverlayPortal` now correctly disables dragging components when dragging within the portal.  
- Disabling `disableDrag` and:  
  - Enabling `disableDragOnFocus` disables dragging when focusing on elements within the portal.  
  - Disabling `disableDragOnFocus` does not disable dragging when focusing on elements within the portal and the component can freely be repositioned.

Closes #1283

### DIFF
--- a/packages/core/lib/overlay-portal/index.tsx
+++ b/packages/core/lib/overlay-portal/index.tsx
@@ -30,13 +30,13 @@ export const registerOverlayPortal = (
     });
   };
 
-  if (disableDragOnFocus) {
-    el.addEventListener("focus", onFocus, { capture: true });
-    el.addEventListener("blur", onBlur, { capture: true });
-  } else if (disableDrag) {
+  if (disableDrag) {
     el.addEventListener("pointerdown", stopPropagation, {
       capture: true,
     });
+  } else if (disableDragOnFocus) {
+    el.addEventListener("focus", onFocus, { capture: true });
+    el.addEventListener("blur", onBlur, { capture: true });
   }
 
   el.setAttribute("data-puck-overlay-portal", "true");
@@ -46,13 +46,13 @@ export const registerOverlayPortal = (
       capture: true,
     });
 
-    if (disableDragOnFocus) {
-      el.removeEventListener("focus", onFocus, { capture: true });
-      el.removeEventListener("blur", onFocus, { capture: true });
-    } else if (disableDrag) {
+    if (disableDrag) {
       el.removeEventListener("pointerdown", stopPropagation, {
         capture: true,
       });
+    } else if (disableDragOnFocus) {
+      el.removeEventListener("focus", onFocus, { capture: true });
+      el.removeEventListener("blur", onBlur, { capture: true });
     }
 
     el.removeAttribute("data-puck-overlay-portal");


packages/core/lib/overlay-portal/index.tsx
```diff
@@ -30,13 +30,13 @@ export const registerOverlayPortal = (
     });
   };
```
FedericoBonel, line 31: @chrisvxd 

In theory this should execute inside a useEffect only once, but we might want to move these callbacks outside of the function itself to avoid recreating them and loosing reference, or using useCallback.

What do you think?

chrisvxd: I don't think it will have a significant impact but feel free to make those refactors in a follow up

chrisvxd (approved):

PR #1346, opened 2025-10-02
feat: disable cache for external fields via cache prop

### LINKED ISSUE #1339: Add cache control in external field
## Description

Currently, there is no way to control the internal cache that external fields use for fetching their lists. This leads to outdated data when the listed content is frequently updated, requiring a page refresh to get the latest data.

If Puck could provide a way to manage that internal cache so developers can control when it should be invalidated and refetched, it would avoid this issue and make external fields suitable for more use cases.

## Considerations

- The cache is [implemented in this file](https://github.com/puckeditor/puck/blob/16a3eee1a81f4df275dfbbb0903d4274ecaeb400/packages/core/components/ExternalInput/index.tsx) and used within [this function](https://github.com/puckeditor/puck/blob/16a3eee1a81f4df275dfbbb0903d4274ecaeb400/packages/core/components/ExternalInput/index.tsx#L78C3-L95C5).

### Proposal 1

- Provide options to set the time the cache should be reused in milliseconds. This would prevent the cache from becoming outdated during long editing sessions.

```tsx
fields: {
  myField: { 
    type: "external", 
    cacheTime: 60000, // set a cache lifetime of 1 minute
  },
}
```

### Proposal 2

- Provide an option to enable or disable the cache. This would allow use cases where the data changes frequently in short timeframes, making caching less useful. It would also allow use cases where people might already have a caching system in place.

```tsx
fields: {
  myField: { 
    type: "external", 
    cache: false, // Defaults to true
  },
}
```

It could also be implemented through Proposal 1 by setting a `cacheTime` of 0 milliseconds.

```tsx
fields: {
  myField: { 
    type: "external", 
    cacheTime: 0, // Disable the cache
  },
}
```

### Proposal 3

- Provide a function that returns the cache key to use internally for the current query and filters. If the returned value is nullish, the cache should be disabled. This would allow more control over which cached data is used and when.

```tsx
fields: {
  myField: { 
    type: "external", 
    cache: ({ id, query, filters }) => `${id}-${query}-${JSON.stringify(filters)}`, // Or null to disable the cache
  },
}
```

It could also be combined with Proposal 1.

```tsx
fields: {
  myField: { 
    type: "external", 
    cacheTime: 60000, // Invalidate any key after 60 seconds since the last fetch
    cache: ({ id, query, filters }) => `${id}-${query}-${JSON.stringify(filters)}`,
  },
}
```

Another option would be to return an object that includes the cache time for that specific key. In this case, the cache would be disabled when the `time` property is set to 0.

```tsx
fields: {
  myField: { 
    type: "external", 
    cache: ({ id, query, filters }) => ({ key: `${id}-${query}-${JSON.stringify(filters)}`, time: 60000 })
  },
}
```

matthewlynch-wise:
This PR closes #1339  ~~via Proposal 2~~ .

I've also added one additional change that stops the form event from searching via`<ExternalInput />` submitting a parent form/page. We were running into this issue with payload. Happy to spin the change into a separate issue/PR if you wish.

Wise internal refs: OGE-5640 / OGE-5641.

### DIFF
--- a/apps/docs/pages/docs/api-reference/fields/external.mdx
+++ b/apps/docs/pages/docs/api-reference/fields/external.mdx
@@ -55,7 +55,7 @@ const config = {
 | ----------------------------------------- | -------------------------------------------- | ---------- | -------- |
 | [`type`](#type)                           | `type: "external"`                           | "external" | Required |
 | [`fetchList()`](#fetchlistqueryparams)    | `fetchList: async () => []`                  | Function   | Required |
-| [`cache`](#cache)                         | `cache: { enabled: true }`                   | Object    | -        |
+| [`cache`](#cache)                         | `cache: { enabled: true }`                   | Object     | -        |
 | [`filterFields`](#filterfields)           | `{ "rating": { type: "number" } }`           | Object     | -        |
 | [`getItemSummary()`](#getitemsummaryitem) | `getItemSummary: async ({ title }) => title` | Function   | -        |
 | [`initialFilters`](#initialfilters)       | `{ "rating": 1 }`                            | Object     | -        |


packages/core/components/ExternalInput/index.tsx
```diff
@@ -39,6 +39,7 @@ export const ExternalInput = ({
   const {
     mapProp = (val: any) => val,
     mapRow = (val: any) => val,
+    cache: shouldCacheData = true,
```
chrisvxd: It wasn't included in the proposals, but I wonder if we should actually have a `CacheOpts` shape here, to provide a path to future caching functionality

```
cache: {
  enabled: true
}
```

What do you think? If you do it, would you be able to fix-up your first commit?

matthewlynch-wise: Happy to do this

FedericoBonel (approved):
Tested it out, it's working nicely! I also like the object, bc it gives us room to add timeouts and other configuration options in the future 👍.

apps/docs/pages/docs/api-reference/fields/external.mdx
```diff
@@ -55,6 +55,7 @@ const config = {
 | ----------------------------------------- | -------------------------------------------- | ---------- | -------- |
 | [`type`](#type)                           | `type: "external"`                           | "external" | Required |
 | [`fetchList()`](#fetchlistqueryparams)    | `fetchList: async () => []`                  | Function   | Required |
+| [`cache`](#cache)                         | `cache: { enabled: true }`                   | Boolean    | -        |
```
chrisvxd:
```suggestion
| [`cache`](#cache)                         | `cache: { enabled: true }`                   | Object    | -        |
```

matthewlynch-wise:
Thanks for the docs type fix @chrisvxd 🙏

chrisvxd:
@matthewlynch-wise yw, but I broke the formatting! 🙈 Would you mind formatting and pushing up?

matthewlynch-wise:
@chrisvxd fixed 🏓

chrisvxd (approved):

PR #1347, opened 2025-10-02
fix: updating root component Id using populateIds

### LINKED ISSUE #1221: Using `defaultProps` for slots at the root level results in error

xaviemirmon:
fixes #1221 

Supercedes #1303 

Updating root component ids with `populateIds` helper function to avoid touching `walk-app-state`.

### DIFF
--- a/packages/core/components/Puck/index.tsx
+++ b/packages/core/components/Puck/index.tsx
@@ -232,13 +232,14 @@ function PuckProvider<
 
     const defaultedRootProps = {
       ...config.root?.defaultProps,
-      ...rootProps,
+      ...(rootProps as AsFieldProps<DefaultComponentProps> | AsFieldProps<any>),
     };
 
     const root = populateIds(
       toComponent({ ...initialData?.root, props: defaultedRootProps }),
       config
     );
+
     const newAppState = {
       ...defaultAppState,
       data: {


packages/core/components/Puck/index.tsx
```diff
@@ -230,6 +254,8 @@ function PuckProvider<
       ...rootProps,
     };
 
+    updateIds(defaultedRootProps, config);
```
chrisvxd: Can we not just use `populateIds` here? It's recursive so should handle everything you've implemented in `updateIds`

```ts
const root = populateIds(
  toComponent({ ...initialData?.root, props: defaultedRootProps }),
  config
)
```

xaviemirmon: Oh cool I didn't know about the toComponent helper!

Issue #1238, opened 2025-08-13
Puck Issue: dragRef loses required attributes when component re-renders or changes


muchisx:
## Summary

When using `puck.dragRef` dynamically (when the component owning the ref changes or re-renders), the subsequent component doesn't receive all the necessary attributes, causing visual and functional issues.

## Current Behavior

After a component re-render or when the ref is moved to a different component, the draggable element loses critical `data-puck-*` attributes, specifically:

- `data-puck-disabled`
- `data-puck-component`
- `data-puck-dnd`

This results in the element becoming invisible due to Puck's CSS rule:

```css
._DropZone_1i2sv_1 > *:not([data-puck-component]) {
  opacity: 0;
}
```

In addition to not having the necessary attributes, I could only assume that the entire engine behind the ref gets desynced? But I didn't test that.

https://github.com/user-attachments/assets/2479e3b3-f137-443f-aa7c-e228b2aba0f0

## Expected Behavior

The `dragRef` should maintain all necessary Puck attributes when:

1. The component owning the ref re-renders
2. The ref is transferred to a different component instance
3. Any other dynamic usage scenarios

## Reproduction Example

**Initial render (working):**

```html
<div
  class="..."
  tabindex="0"
  role="button"
  aria-roledescription="draggable"
  aria-describedby="dnd-kit-description-12"
  aria-pressed="false"
  aria-grabbed="false"
  aria-disabled="true"
  data-puck-disabled=""
  data-puck-component="53740f76-8f83-4bb4-bd9c-cf30cb5cbf14"
  data-puck-dnd="53740f76-8f83-4bb4-bd9c-cf30cb5cbf14"
  style="position: relative"
>
  <!-- content -->
</div>
```

**After re-render (broken):**

```html
<div
  class="..."
  tabindex="0"
  role="button"
  aria-roledescription="draggable"
  aria-describedby="dnd-kit-description-10"
  aria-pressed="false"
  aria-grabbed="false"
  aria-disabled="true"
>
  <!-- content -->
</div>
```

## Reproduction

This issue can be reproduced with a simple component that conditionally renders different elements with the same `dragRef`:

```tsx
inline: true,
render: function Render({ puck, xSetting }) {
  const isX = xSetting === 'valueX';

  return isX ? <div ref={puck.dragRef}>true</div> : <span ref={puck.dragRef}>false</span>;
},
```

[...]

PR #1351, opened 2025-10-03
fix: change ref to state so that the component re-renders if the element is changed

xaviemirmon:
fixes: #1238

### DIFF
--- a/packages/core/components/DraggableComponent/index.tsx
+++ b/packages/core/components/DraggableComponent/index.tsx
@@ -214,11 +214,11 @@ export const DraggableComponent = ({
       sortable.droppable.disabled = !s.enabledIndex[zoneCompound];
     });
 
-    if (ref.current && !permissions.drag) {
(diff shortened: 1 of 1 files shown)


packages/core/components/DraggableComponent/index.tsx
```diff
@@ -214,26 +214,26 @@ export const DraggableComponent = ({
...
+        elState?.removeAttribute("data-puck-disabled");
         cleanup();
       };
     }
 
     return cleanup;
   }, [permissions.drag, zoneCompound]);
 
-  const ref = useRef<HTMLElement>(null);
+  const [elState, setElState] = useState<HTMLElement | null>(null);
```
chrisvxd: The "State" suffix here seems unnecessary. Suggest just going with `el` or `targetEl`

xaviemirmon:
closing in favour of #1360

Issue #1354, opened 2025-10-06
Title undefined in RR7 recipe


chrisvxd:
## Description

When generating an RR7 app, the page "title" is undefined

## Environment

- Puck version: 0.20.2

## Steps to reproduce

1. Install the recipe using `npx create-puck-app@latest react-router-demo`
2. Run the application in development mode with `npm run dev`
3. Visit the homepage

## What happens

The page title is "undefined".

## What I expect to happen

The page title to represent the root title in the data.

PR #1355, opened 2025-10-06
fix(meta): undefined title in React Router v7 recipe

ghost:
Use optional chaining to safely access nested props and avoid undefined values.

Resolves: #1354

### DIFF
--- a/recipes/react-router/app/routes/_index.tsx
+++ b/recipes/react-router/app/routes/_index.tsx
@@ -21,7 +21,7 @@ export async function loader() {
 export function meta({ data: loaderData }: Route.MetaArgs) {
   return [
     {
-      title: loaderData?.data?.root?.props?.title ?? '',
+      title: loaderData.data.root.props?.title ?? "",
     },
   ];
 }
--- a/recipes/react-router/app/routes/puck-splat.tsx
+++ b/recipes/react-router/app/routes/puck-splat.tsx
@@ -40,9 +40,9 @@ export async function loader({ params }: Route.LoaderArgs) {
 export function meta({ data: loaderData }: Route.MetaArgs) {
   return [
     {
-      title: loaderData?.isEditorRoute
+      title: loaderData.isEditorRoute
         ? `Edit: ${loaderData.path}`
-        : loaderData?.data?.root?.props?.title ?? '',
+        : loaderData.data.root.props?.title ?? "",
     },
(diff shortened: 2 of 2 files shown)


recipes/react-router/app/routes/_index.tsx
```diff
@@ -21,7 +21,7 @@ export async function loader() {
 export function meta({ data: loaderData }: Route.MetaArgs) {
   return [
     {
-      title: loaderData.data.root.title,
+      title: loaderData?.data?.root?.props?.title ?? '',
```
FedericoBonel:
```suggestion
      title: loaderData.data.root.props?.title ?? "",
```
The actual optional property here is `props`, so I don't think we need full optional chaining. Using it might end up hiding other issues that aren't related to this specific case.

recipes/react-router/app/routes/puck-splat.tsx
```diff
@@ -40,9 +40,9 @@ export async function loader({ params }: Route.LoaderArgs) {
 export function meta({ data: loaderData }: Route.MetaArgs) {
   return [
     {
-      title: loaderData.isEditorRoute
+      title: loaderData?.isEditorRoute
         ? `Edit: ${loaderData.path}`
-        : loaderData.data.root.title,
+        : loaderData?.data?.root?.props?.title ?? '',
```
FedericoBonel:
```suggestion
        : loaderData.data.root.props?.title ?? "",
```
Same as above

recipes/react-router/app/routes/puck-splat.tsx
```diff
@@ -40,9 +40,9 @@ export async function loader({ params }: Route.LoaderArgs) {
 export function meta({ data: loaderData }: Route.MetaArgs) {
   return [
     {
-      title: loaderData.isEditorRoute
+      title: loaderData?.isEditorRoute
```
FedericoBonel:
```suggestion
      title: loaderData.isEditorRoute
```

The loader should always return an object for `loaderData`, so the optional chaining here shouldn’t be necessary.

recipes/react-router/app/routes/_index.tsx
ghost: yeah @FedericoBonel, I was skeptical at first too, but thanks! I’m applying the changes.

PR #1364, opened 2025-10-08
Feature: Add maxItems Property to Slot Component

nazrdogan:
**Problem**

  Currently, Puck's Slot component allows unlimited items to be added. In many real-world use cases, developers need to limit the number of items that can be
  placed in a slot to maintain design integrity and usability.

  **Solution**

  This PR introduces an optional maxItems property to the Slot field configuration:

  ```
{
    type: 'slot',
    label: 'Action Buttons',
    allow: ['Button'],
    maxItems: 3  // Limits slot to maximum 3 buttons
  }
```

  **Why This Matters**

  **Without maxItems:**
  - Users can add unlimited items, potentially breaking layouts
  - No way to enforce design system constraints
  - Content editors lack guidance on recommended limits

  **With maxItems:**
  - Enforces component limits at the framework level
  - Provides immediate visual feedback when limit is reached
  - Maintains consistent user experience across the application

  **Common Use Cases**

  - Hero CTAs: Limit to 2-3 buttons to avoid decision paralysis
  - Navigation Bars: Cap at 5-7 items for optimal usability
  - Card Grids: Restrict to specific numbers for responsive layouts
  - Testimonials: Limit carousel items for better performance
  - Footer Links: Control number of link columns

### DIFF
--- a/packages/core/components/DropZone/index.tsx
+++ b/packages/core/components/DropZone/index.tsx
@@ -278,6 +278,7 @@ export const DropZoneEdit = forwardRef<HTMLDivElement, DropZoneProps>(
       zone,
       allow,
       disallow,
+      maxItems,
       style,
       className,
       minEmptyHeight: userMinEmptyHeight = 128,
@@ -360,6 +361,11 @@ export const DropZoneEdit = forwardRef<HTMLDivElement, DropZoneProps>(
           return true;
         }
 
+        // Check maxItems limit
+        if (maxItems !== undefined && contentIds.length >= maxItems) {
+          return false;
+        }
+
         if (disallow) {
           const defaultedAllow = allow || [];
 
@@ -379,7 +385,7 @@ export const DropZoneEdit = forwardRef<HTMLDivElement, DropZoneProps>(
 
         return true;
       },
-      [allow, disallow]
+      [allow, disallow, maxItems, contentIds.length]
     );
 
     const targetAccepted = useContextStore(ZoneStoreContext, (s) => {
@@ -471,6 +477,7 @@ export const DropZoneEdit = forwardRef<HTMLDivElement, DropZoneProps>(
           isAreaSelected,
           hasChildren: contentIds.length > 0,
           isAnimating,
+          isMaxed: maxItems !== undefined && contentIds.length >= maxItems,
         })}${className ? ` ${className}` : ""}`}
         ref={(node) => {
           assignRefs<HTMLDivElement>([ref, dropRef, userRef], node);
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: packages/core/components/DropZone/index.tsx [hunk 1]
```diff
@@ -500,6 +507,27 @@ export const DropZoneEdit = forwardRef<HTMLDivElement, DropZoneProps>(
...
+              fontSize: "12px",
+              fontWeight: "500",
+              pointerEvents: "none",
+              zIndex: 10,
+              whiteSpace: "nowrap",
+            }}
+          >
+            Max items reached ({maxItems}/{maxItems})
+          </div>
+        )}
```
FedericoBonel, line 530: This never really shows up in the canvas. I believe because of the stacking context, but not 100% sure. I'm also not 100% about this kind of "invasive" feedback that bleeds into the user page preview all the time.

Image

Also, is there any reason why you decided to use inline styles here and not CSS classes, given that these styles are not reacting to anything, I'd expect a class here, unless there's a clear reason why.

packages/core/components/DropZone/styles.module.css
```diff
@@ -60,6 +60,15 @@
...
 
+.DropZone--isMaxed {
+  opacity: 0.6;
+}
+
+[data-puck-dragging] .DropZone--isMaxed {
+  outline: 2px dashed var(--puck-color-grey-06);
+  background: color-mix(in srgb, var(--puck-color-grey-09) 20%, transparent);
+}
+
```
FedericoBonel, line 71: These styles have some issues. 

With this, the visual feedback to know I can drag and drop something inside a nested slot within the slot that is maxed out never shows up, until I fully drag the element on top.

Here's a video demonstrating this, note that, when not maxed out, I see a blue outline in the grid to know there's a slot there, but when the container is maxed out the grid doesn't have that outline any longer:


https://github.com/user-attachments/assets/7456ebb2-d18e-4e60-bdfd-17859c967720

packages/core/components/DropZone/index.tsx
```diff
@@ -360,6 +361,11 @@ export const DropZoneEdit = forwardRef<HTMLDivElement, DropZoneProps>(
           return true;
         }
 
+        // Check maxItems limit
+        if (maxItems !== undefined && contentIds.length >= maxItems) {
+          return false;
+        }
+
```
FedericoBonel, line 368: This is a good start! But this check is not enough for us to know if something can be dropped in a slot or not, because with this, I can't reorder something that is already within a maxed-out slot.

Here's a video showing this, note how the heading that is already inside the maxed-out slot can't be reordered and gets dropped outside the slot:

https://github.com/user-attachments/assets/c47939c3-adce-4261-9eba-a97a31490e57

We should also check if the element being dragged is already positioned within the maxed out slot or not.

FedericoBonel:
Since there haven’t been any updates in over a month, I’ll be closing this PR as stale.

Feel free to pick it back up and re-open the PR anytime!

PR #1371, opened 2025-10-10
chore: fixing dependabot warnings

xaviemirmon:
Attempting to manually fixing dependabot warning which can't be done automatically

### DIFF
--- a/packages/eslint-config-custom/index.js
+++ b/packages/eslint-config-custom/index.js
@@ -1,5 +1,7 @@
+const turboConfig = require("eslint-config-turbo").default.extends;
+
 module.exports = {
-  extends: ["next", "turbo", "prettier"],
+  extends: ["next", ...turboConfig, "prettier"],
   rules: {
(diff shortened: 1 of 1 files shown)


.prettierignore
```diff
@@ -1 +1,3 @@
 CHANGELOG.md
+
+/.nx/workspace-data
```
chrisvxd: Was this addition necessary? Regardless, it needs a new line

chrisvxd (changes requested):
Build is failing unfortunately

chrisvxd:
@xaviemirmon I think the eslint config might need updating. Checking the build logs:

```
Oops! Something went wrong! :(

ESLint: 7.32.0

Error: ESLint configuration in ../../.eslintrc.js » eslint-config-custom » eslint-config-turbo is invalid:
	- Unexpected top-level property "default".
```

chrisvxd (approved):

PR #1376, opened 2025-10-12
Remove unused private APIs from resolveAllData

### LINKED ISSUE #899: resolveAllData has deprecated APIs
The `resolveAllData` function has deprecated APIs that were [previously used](https://github.com/measuredco/puck/blob/358e61995aabc889ab0df60704d78c91b3004855/packages/core/components/Puck/index.tsx#L155-L176) for managing component load state internally. Loading states are now handled differently.

These snuck into the public API and are currently undocumented.

Before removing, check whether anyone is using it.

FedericoBonel:
Closes #899

The `resolveAllData` function contained deprecated APIs that were [previously used](https://github.com/measuredco/puck/blob/358e61995aabc889ab0df60704d78c91b3004855/packages/core/components/Puck/index.tsx#L155-L176) to manage component load state internally.

These APIs were never documented or intended for external use outside of Puck’s internals.

So far, no one has reported using these APIs, but we should document this as a breaking change in the migration guide.

### DIFF
--- a/packages/core/lib/resolve-all-data.ts
+++ b/packages/core/lib/resolve-all-data.ts
@@ -16,19 +16,11 @@ import { mapFields } from "./data/map-fields";
 export async function resolveAllData<
   Components extends DefaultComponents = DefaultComponents,
   RootProps extends Record<string, any> = DefaultRootFieldProps
->(
-  data: Partial<Data>,
-  config: Config,
-  metadata: Metadata = {},
-  onResolveStart?: (item: ComponentData) => void,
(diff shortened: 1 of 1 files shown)


chrisvxd (approved):

FedericoBonel:
I think you mentioned this was never meant to be exposed and wasn't really being used or how it would be used by people outside Puck.

Want me to change it and add a deprecated note?

chrisvxd:
Retaining APIs as they're the only way to identify when a node is resolving in the tree. They need documenting.

PR #1378, opened 2025-10-15
Custom field resolution in resolvefields

### LINKED ISSUE #1288: Custom field resolution in `resolveFields`

FedericoBonel:
The current implementation of the `AutoField` component doesn't remount fields unless their `field.type` prop changes. This makes it impossible to resolve two different custom fields under the same prop.

To fix this, I've added a `key` prop to custom fields. You can now provide a unique key that, when changed on the same custom field instance, forces the field to remount and apply any new rendering you might need.

Closes #1288

### DIFF
--- a/packages/core/types/Fields.ts
+++ b/packages/core/types/Fields.ts
@@ -134,7 +134,7 @@ export interface CustomField<Value extends any> extends BaseField {
   render: CustomFieldRender<Value>;
   contentEditable?: boolean;
   key?: string;
-};
+}
 
 export interface SlotField extends BaseField {
   type: "slot";


apps/docs/pages/docs/api-reference/fields/custom.mdx
```diff
@@ -211,6 +212,156 @@ const config = {
   <Puck.Preview />
 </ConfigPreview>
 
+### key
+
+Assign a unique key to force the field to remount when resolving different custom field types on the same prop.
```
chrisvxd:
```suggestion
Assign a unique key to force the field to remount when using [`resolveFields`](/docs/api-reference/configuration/component-config#resolvefieldsdata-params) to resolve different custom field types on the same prop.
```

apps/docs/pages/docs/api-reference/fields/custom.mdx
```diff
@@ -211,6 +212,156 @@ const config = {
...
+              key: "custom-text",
+              render: ({ onChange, value }) => (
+                <label>
+                  Text Input
+                  <input
+                    value={value}
+                    name={name}
+                    onChange={(e) => onChange(e.currentTarget.value)}
+                  />
+                </label>
```
chrisvxd: Hm this example is good but pretty verbose. Can we simplify it a bit?

How about removing the label, or the render methods entirely (and replacing with `// ...`)

```suggestion
                  <input
                    value={value}
                    name={name}
                    onChange={(e) => onChange(e.currentTarget.value)}
                  />
```

apps/docs/pages/docs/api-reference/fields/custom.mdx
```diff
@@ -211,6 +212,156 @@ const config = {
...
+            { label: "Yes", value: true },
+            { label: "No", value: false },
+          ],
+        },
+      },
+      resolveFields: (data, params) => {
+        if (data.props.isText) {
+          return {
+            ...params.fields,
+            title: {
```
chrisvxd: Suggest you use a different name for the prop, as a number "title" doesn't make sense. Perhaps `prop` is fine.

apps/docs/pages/docs/api-reference/fields/custom.mdx
FedericoBonel: Little side quest: I also made the heading inline code and added code line numbers in the `contentEditable` section to keep it consistent with the rest of the page.

apps/docs/pages/docs/api-reference/fields/custom.mdx
```diff
@@ -211,6 +212,136 @@ const config = {
   <Puck.Preview />
 </ConfigPreview>
 
+### `key`
+
+Assign a unique key to force the field to remount when using [`resolveFields`](/docs/api-reference/configuration/component-config#resolvefieldsdata-params) to resolve different custom field types on the same prop.
+
+```tsx {17-21, 26-30} showLineNumbers copy
```
chrisvxd: One final comment - we should only be highlighting the lines with `key` in the code snippets

chrisvxd (changes requested):

chrisvxd:
@FedericoBonel approved! But can you rebase this? I think the merge introduced conflicts.

FedericoBonel:
Rebased!

PR #1380, opened 2025-10-16
feat: add disable hotkeys feature

do4Mother:
when using a wysiwyg editor to the puck, using the CTRL+Z or CTRL+R keys will trigger undo or redo simultaneously.

### DIFF
--- a/packages/core/components/Puck/index.tsx
+++ b/packages/core/components/Puck/index.tsx
@@ -160,7 +160,7 @@ function PuckProvider<
     metadata,
     onAction,
     fieldTransforms,
-    hotkeys,
+    hotkeys = { enabled: true },
   } = usePropsContext();
 
   const iframe: IframeConfig = useMemo(
@@ -244,7 +244,6 @@ function PuckProvider<
 
     const newAppState = {
       ...defaultAppState,
-      hotkeys,
       data: {
         ...initialData,
         root: { ...initialData?.root, props: root.props },
@@ -333,6 +332,7 @@ function PuckProvider<
         onAction,
         metadata,
         fieldTransforms: loadedFieldTransforms,
+        hotkeys: hotkeys,
       };
     },
     [
@@ -345,6 +345,7 @@ function PuckProvider<
       onAction,
       metadata,
       loadedFieldTransforms,
+      hotkeys
     ]
   );
 
@@ -434,7 +435,7 @@ function PuckLayout<
   const rightSideBarVisible = useAppStore(
     (s) => s.state.ui.rightSideBarVisible
   );
-  const enabledHotkeys = useAppStore((s) => s.state.hotkeys.enabled);
+  const enabledHotkeys = useAppStore((s) => s.hotkeys.enabled);
 
   const {
     width: leftWidth,
--- a/packages/core/components/Puck/index.tsx
+++ b/packages/core/components/Puck/index.tsx
@@ -345,7 +345,7 @@ function PuckProvider<
       onAction,
       metadata,
       loadedFieldTransforms,
-      hotkeys
+      hotkeys.enabled
     ]
   );
 
--- a/packages/core/components/Puck/index.tsx
+++ b/packages/core/components/Puck/index.tsx
@@ -27,7 +27,7 @@ import type {
   Metadata,
   AsFieldProps,
   DefaultComponentProps,
-  ComponentData,
+  HotkeyState,
 } from "../../types";
 
 import { SidebarSection } from "../SidebarSection";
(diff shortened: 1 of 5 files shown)


FedericoBonel:
Hello @do4Mother!

Thank you very much for your contribution 🙏

For the record, you can actually avoid triggering Puck hotkeys by using an `onKeyDown` event listener on your RTE or WYSIWYG component and stopping propagation. We’re currently implementing a Puck RTE, and we needed to do this as well to prevent Puck from switching into interactive mode when pressing `cmd + i`.

You can check out the implementation here:  
https://github.com/puckeditor/puck/pull/1341/commits/90f3cf5fe0f1f5c84baa2d8ca81d9ea81c55292f#diff-ae9dac1ee1543faeeb141edd12591423d6ecfa47b4cf79c5875e825f475ed95cR1-R108  

In that case, we stop propagation in the capturing phase (`onKeyDownCapture`), but you might be better off stopping it in the bubbling phase (`onKeyDown`).

We normally prefer having a feature request or bug report before opening PRs. This helps us understand the need and discuss the best approach before implementation.  

That said, having a set of props to control Puck hotkeys would definitely be beneficial, so I have a small request to better align it with what we’d expect to support:

Could you update the API to work as a `hotkeys` prop on the `Puck` component that receives an object, with an `enabled` property inside it? Something like this:

```tsx
<Puck
  hotkeys={{ enabled: false }}
/>
```

That would allow us to add more controls in the future.

Thank you again for taking the time to contribute!

FedericoBonel (changes requested):
Check above!

do4Mother:
Sorry, I was so excited to contribute to this awesome project that I didn't read the documentation on how to contribute. 

Additionally, in the use case project I'm working on, 80% of the work is done using the WYSIWYG editor, so I don't disable hotkeys in the editor.

And I have been updated the state. please help to check it. thanks!

chrisvxd (approved):

packages/core/types/AppState.tsx
```diff
@@ -39,6 +43,7 @@ export type UiState = {
 };
 
 export type AppState<UserData extends Data = Data> = {
+  hotkeys: HotkeyState;
```
chrisvxd: Can you move this to the [App**Store**](https://github.com/puckeditor/puck/blob/main/packages/core/store/index.ts#L51) rather than AppState?

chrisvxd (changes requested):

packages/core/types/AppState.tsx
do4Mother: okay! i have done a refactor update to the appstore

packages/core/components/Puck/index.tsx
```diff
@@ -330,6 +332,7 @@ function PuckProvider<
         onAction,
         metadata,
         fieldTransforms: loadedFieldTransforms,
+        hotkeys: hotkeys,
```
FedericoBonel:
```suggestion
        hotkeys,
```

packages/core/components/Puck/index.tsx
```diff
@@ -342,6 +345,7 @@ function PuckProvider<
       onAction,
       metadata,
       loadedFieldTransforms,
+      hotkeys
```
FedericoBonel:
```suggestion
      hotkeys.enabled
```

I don't know if we want to subscribe to the whole object here (which could be re-created through re renders), it might be better to start off with hotkeys.enabled, just in case.

packages/core/components/Puck/index.tsx
```diff
@@ -124,6 +124,7 @@ type PuckProps<
   };
   initialHistory?: InitialHistory;
   metadata?: Metadata;
+  hotkeys?: HotkeyState
```
FedericoBonel: I think we need to use a `Partial` here and handle that optionality (applying defaults) before passing it down to the `appStore`.

We might introduce more options in the future, and requiring `enabled` could become inconvenient, some users might only want to enable a single option.

FedericoBonel (changes requested):
Thanks again for your contribution @do4Mother!

Awesome stuff, left couple of comments above.

do4Mother:
@FedericoBonel Thanks for suggestions and build cool project!

FedericoBonel:
Hey @do4Mother! Thanks again for the contribution, and I'm incredibly sorry for the massive delay on this. For some reason, it slipped under our radar.

I think this looks good now, but it needs two final things:

- Rebasing onto `main` to pick up the latest changes
- Some documentation for the Puck props

If you don't want to continue with this PR, that's totally understandable, and we can pick it up. Just let me know so I know how to proceed.

PR #1385, opened 2025-10-21
React Native rendering support

aleksivuorela:
We would like to use Puck in React Native project that uses [react-native-web](https://github.com/necolas/react-native-web). Requirements were that editor works on web only, but rendering the JSON with `<Render>` should work cross-platform.

Issue is that internally Puck renders `<div>` elements that do not work on native platforms. This was solved by adding `overrides` to config for rendering div – similar to how you can already override parts of the editor UI.

Now divs can be replaced with `<View>` on native platforms by providing config like this to `<Render>`

```
import { View } from 'react-native';

const config = {
  components: { ... },
  overrides: {
    slot: props => {
      return <View {...props} />;
    }
  }
}
```

Another issue that caused app to crash on native platforms when importing anything from `@measured/puck` was in `bubble-pointer-events.ts` which tried to use `Event` interface, but it's not defined in native environment. This interface was simply shimmed to prevent crash since I believe it's relevant for editor only.

These changes allowed us to use Puck in our React Native project successfully.

### DIFF
--- a/packages/core/types/Config.tsx
+++ b/packages/core/types/Config.tsx
@@ -5,7 +5,7 @@ import { ComponentData, ComponentMetadata, RootData } from "./Data";
 import { AsFieldProps, WithChildren, WithId, WithPuckProps } from "./Utils";
 import { AppState } from "./AppState";
 import { DefaultComponentProps, DefaultRootFieldProps } from "./Props";
-import { Permissions } from "./API";
+import { Permissions, RenderOverrides } from "./API";
 import { DropZoneProps } from "../components/DropZone/types";
 import {
   AssertHasValue,
@@ -175,6 +175,7 @@ type ConfigInternal<
     >;
   };
   root?: RootConfigInternal<RootProps, UserField>;
+  overrides?: RenderOverrides;
 };
 
 // This _deliberately_ casts as any so the user can pass in something that widens the types
--- a/packages/core/components/DropZone/index.tsx
+++ b/packages/core/components/DropZone/index.tsx
@@ -487,7 +487,14 @@ export const DropZoneEdit = forwardRef<HTMLDivElement, DropZoneProps>(
       [dropRef]
     );
 
-    const El = as ?? "div";
+    const config = useAppStore((s) => s.config);
+
+    const Wrapper = useMemo(
+      () => config.overrides?.slot || "div",
+      [config.overrides]
+    );
+
+    const El = as ?? Wrapper;
 
     return (
       <El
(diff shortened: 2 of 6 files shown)


FedericoBonel:
Hello @aleksivuorela!

Thank you very much for your contribution 🙏

We usually prefer having a feature request or bug report before opening a PR. This helps us understand the use case and agree on the best approach before implementation. You'll also have the full context, and we’ll be better prepared to review your changes.

You can read more about our process in the [Contributing Guide](https://github.com/puckeditor/puck/blob/main/CONTRIBUTING.md).

All this means, it might take us a bit longer to review and decide how to move forward with this PR, and we might have some review comments that are higher level before going through implemention details.

aleksivuorela:
> Edit: I've just opened #1418. I think your PR would work in congruence if you can rename the `div` override to `slot`.

@chrisvxd I've renamed the override to `slot` and rebased the branch with your latest changes.

Basically it allows overriding all the default divs via config at once, for example in React Native context where divs aren't available.

packages/core/types/Config.tsx
```diff
@@ -174,6 +174,7 @@ type ConfigInternal<
     >;
   };
   root?: RootConfigInternal<RootProps, UserField>;
+  overrides?: RenderOverrides;
```
chrisvxd, line 178: This change doesn't seem necessary

chrisvxd (changes requested):
Apologies for the delay but this looks good to me!

Could you add to the docs [here](https://github.com/puckeditor/puck/blob/main/apps/docs/pages/docs/api-reference/overrides.mdx) and [here](https://github.com/puckeditor/puck/tree/main/apps/docs/pages/docs/api-reference/overrides)?

_Edit: Please also rebase and resolve conflicts_

packages/core/types/Config.tsx
aleksivuorela: It is necessary – render `overrides` is provided via the config object. That way both `<Puck>` and `<Render>` use the same overrides as they share the config. I suppose the correct place for documentation would be [here](https://github.com/puckeditor/puck/blob/main/apps/docs/pages/docs/api-reference/configuration/config.mdx).

Perhaps the prop should be renamed e.g. `renderOverrides` so it's not confused with [editor overrides](https://github.com/puckeditor/puck/blob/main/apps/docs/pages/docs/api-reference/overrides.mdx).

chrisvxd: Apologies for the delay on this one. Just catching up after a busy start to the year and checking back.

I hesitated on this one because adding `overrides` (or `renderOverrides`) to the config is quite a significant change and needs further thought.

We've hit things like this a few times, where there's a good use-case for editor APIs to be available within the renderer, such as with field transforms.

However, I think having `overrides` on the `config` is definitely confusing when we already have `overrides` as a prop.

I think the easier thing might be to use [`as`](https://puckeditor.com/docs/api-reference/fields/slot#as) for now. I'm going to close this one out, but feel free to open another one if you want to get the event shim in, or drop a reply.

chrisvxd: Worth adding, we're also experimenting with a [useSlot hook](https://github.com/puckeditor/puck/pull/1546)

PR #1387, opened 2025-10-22
Add parent data to resolve permissions

### LINKED ISSUE #1338: Add `params.parent` to `resolvePermissions`
## Description

Currently, you can use the `params.parent` prop in the `resolveFields` API, which is useful for defining specific fields when a component is nested in a particular component. 

It would also be helpful to add this prop to the `resolvePermissions` API, allowing you to lock down components when they are dropped inside a specific slot or modify permissions as needed. This would be especially powerful if the prop is added alongside its inclusion in `resolveData`, as requested in #1359.

## Considerations

- Issue #1359 requests this same prop in the `resolveData` API, which could complement this proposal.

## Proposals

Adding the `params.parent` prop to `resolvePermissions` might look like the following:

```tsx
resolvePermissions: (data, params) => {
  if (params.parent.type === "Grid") {
    return { drag: false, delete: false };
  }
  return { drag: true, delete: true };
}
```

FedericoBonel:
Closes #1338.

This PR adds parent data to `resolvePermissions`, matching the behavior of `resolveFields` as required in #1338.

## Changes made

- The `resolvePermissions` config type now includes a `parent` parameter.
  - The permissions slice `resolvePermissions` was updated to retrieve the parent data and pass it to the config-defined function.
- `resolvePermissions` now also runs when components are moved between slots/dropzones.
  - The permissions slice `cache` now stores the parent ID for each execution and re-execute when it changes.
- Added documentation for the new `parent` prop and execution flow.

## Manual tests

- Confirmed that `resolvePermissions` runs the same number of times on editor load, insertion, deletion, replacement, and reordering within the same slot using sample data and config.
- Confirmed that `resolvePermissions` runs once per component when a component is reordered across slots (i.e. when its parent changes).
- Confirmed that conditional permissions based on the parent work as expected:

```tsx
resolvePermissions: (data, params) => {
  if (params.parent.type === "Grid") {
    return { drag: false, delete: false };
  }
  return { drag: true, delete: true };
}
```

### DIFF
--- a/apps/docs/pages/docs/api-reference/configuration/component-config.mdx
+++ b/apps/docs/pages/docs/api-reference/configuration/component-config.mdx
@@ -660,7 +660,7 @@ The last permissions object created by the previous run of this function.
 
 The [component data](/docs/api-reference/data-model/component-data) of the parent.
 
-Resolvers are executed bottom-up when [`<Puck>`](/docs/api-reference/components/puck) renders, so the parent component will only [resolve its data](#resolvedatadata-params) after the child resolver has completed when this happens.
+If the parent implements [resolveData](#resolvedatadata-params), this parameter will return the unresolved data on initial render.
 
 #### Returns
 


apps/docs/pages/docs/api-reference/configuration/component-config.mdx
```diff
@@ -590,7 +591,7 @@ The parent data object if this item is within a [DropZone](/docs/api-reference/c
 
 Dynamically set the [permissions](/docs/api-reference/permissions) for this component to toggle functionality. Can be used to control the permissions on a specific component instance. Inherits [`permissions`](#permissions). Supports asynchronous calls.
 
-This function is triggered when the component data changes.
+This function is triggered when the component is inserted or loaded in the editor, when its data changes, or when it's moved to a different parent.
```
chrisvxd: What does "or loaded in the editor" mean here? Feels a bit awkward. Might be better to just say

```
This function is triggered when the component is inserted, when its data changes, or when it's moved to a different parent.
```

packages/core/store/slices/__tests__/permissions.spec.tsx
```diff
@@ -425,6 +433,197 @@ describe("permissions slice", () => {
...
+      );
+
+      expect(resolvePermissions.mock.calls[1][1].parent).toEqual({
+        type: "root",
+        props: { id: "root" },
+      });
+      expect(resolvePermissions.mock.results[1].value).toEqual(
+        outsideParentPermissions
+      );
+    });
```
chrisvxd, line 602: Would it make sense to actually check the updated permissions cache here? Might be easier than checking the mocks

packages/core/store/slices/__tests__/permissions.spec.tsx
```diff
@@ -425,6 +433,197 @@ describe("permissions slice", () => {
       // Watcher will trigger this
       expect(resolveCalls).toBe(4);
     });
+
```
chrisvxd, line 436: I think you're missing the basic test case for `it("should receive the parent component")`, without moving

apps/docs/pages/docs/api-reference/configuration/component-config.mdx
```diff
@@ -656,6 +657,15 @@ The data object from the previous run of this function.
 
 The last permissions object created by the previous run of this function.
 
+##### `params.parent`
+
+The parent [data object](/docs/api-reference/data-model/component-data) if this item is within a [DropZone](/docs/api-reference/components/drop-zone) or a [Slot](/docs/api-reference/fields/slot).
```
chrisvxd: This can be simplified, as it always receives the data even if not inside a slot. (Aside, but we don't need to document for DropZones, as these are deprecated.) I note that the resolveFields docs are out-of-date in this regard.

```suggestion
The [component data](/docs/api-reference/data-model/component-data) of the parent.
```

apps/docs/pages/docs/api-reference/configuration/component-config.mdx
```diff
@@ -656,6 +657,15 @@ The data object from the previous run of this function.
...
+
+<Callout type="warning">
+The latest parent's [resolved data](#resolvedatadata-params) won't be available on initial load because resolvers run from bottom to top.  
+To ensure it's ready, use [`resolveAllData`](/docs/api-reference/functions/resolve-all-data) before passing [initial data](/docs/api-reference/components/puck#data) to the `<Puck>` component.
+</Callout>
```
chrisvxd: This callout is a little scary, and I'm not sure it's necessary to cause alarm here.

I think we this can be replaced with a less scary line outside of a Callout.

```
Resolvers are executed bottom-up, so the parent component will only resolve after the child resolver has completed.
```

apps/docs/pages/docs/api-reference/configuration/component-config.mdx
```diff
@@ -656,6 +657,15 @@ The data object from the previous run of this function.
 
 The last permissions object created by the previous run of this function.
 
+##### `params.parent`
```
chrisvxd, line 659: Aside, but this file is getting ridiculously long. We should break it down. Mind opening a ticket?

chrisvxd (changes requested):
Great stuff! Left some comments.

apps/docs/pages/docs/api-reference/configuration/component-config.mdx
FedericoBonel: Loaded means when a component gets initially loaded with the editor. I see why it's awkard, will update it!

FedericoBonel: Updated it! I followed the same wording you used for resolveData, let me know what you think.

apps/docs/pages/docs/api-reference/configuration/component-config.mdx
FedericoBonel: Re-wrote it, I followed your suggestion, but added that this is only an issue when `<Puck>` renders since in any other execution you should have the latest resolved data of the parent.

apps/docs/pages/docs/api-reference/configuration/component-config.mdx
FedericoBonel: Done: https://github.com/puckeditor/puck/issues/1396 👍

packages/core/store/slices/__tests__/permissions.spec.tsx
FedericoBonel: Done! I didn't know if you also wanted me to use the cache for checking the resolvePermissions params, so I did it, let me know if you prefered how it was and I can roll that back to use `mock.calls[].parent` 😁

packages/core/store/slices/__tests__/permissions.spec.tsx
FedericoBonel: It has somewhat of an overlap with the `it("provides correct args to component.resolvePermissions on subsequent calls")` above, but I added it. 

I'm actually checking the mock params in that one since we probably want to make sure the function itself receives the expected param and not just that we store its ID in the cache.

apps/docs/pages/docs/api-reference/configuration/component-config.mdx
chrisvxd: Wait, is that correct? Aren't resolvers were executed bottom-up at all times?

packages/core/store/slices/__tests__/permissions.spec.tsx
chrisvxd: overlap is okay!

packages/core/store/slices/__tests__/permissions.spec.tsx
chrisvxd: Oh yeah, cache is a side effect. Silly to test that. Sorry, perhaps let's roll back.

apps/docs/pages/docs/api-reference/configuration/component-config.mdx
FedericoBonel: They are, but with resolvePermissions, the only time you actually execute all the resolvers is when you load the data in the editor, in principle, any other execution, should only trigger when the data, or the parent id changes, because of the cache (or when it's forced by resolveData via refreshPermissions. 
Does that make sense? Maybe the wording is off.
What I'm trying to say is that you won't have the latest data when resolvePermissions runs on the initial puck render.

packages/core/store/slices/__tests__/permissions.spec.tsx
FedericoBonel: All good! That was what I thought as well originally, here we really just care that the resolvePermissions executes with the correct parent, and when it needs to. Rolled it back!

We can add tests specifically for the cache if we want to, but I think counting number of executions already tests that as a black box.

apps/docs/pages/docs/api-reference/configuration/component-config.mdx
```diff
@@ -656,6 +656,12 @@ The data object from the previous run of this function.
...
+
+The [component data](/docs/api-reference/data-model/component-data) of the parent.
+
+Resolvers are executed bottom-up when [`<Puck>`](/docs/api-reference/components/puck) renders, so the parent component will only [resolve its data](#resolvedatadata-params) after the child resolver has completed when this happens.
```
chrisvxd:
```suggestion
If the parent implements [resolveData](#resolvedatadata-params), this parameter will return the unresolved data on initial render.
```

chrisvxd (changes requested):
Suggested wording for the problematic line.

apps/docs/pages/docs/api-reference/configuration/component-config.mdx
chrisvxd: Suggested some wording below

apps/docs/pages/docs/api-reference/configuration/component-config.mdx
FedericoBonel: Nice okay!

apps/docs/pages/docs/api-reference/configuration/component-config.mdx
FedericoBonel: Great! I commited your suggestion. Thank you 🙏

FedericoBonel:
Added your suggestion to the docs!

PR #1382, opened 2025-10-20
feat: add pagination and sortable columns to ExternalInput component

hugodebard:
This PR introduces optional pagination and sortable columns to ExternalInput fields, along with UI improvements and API documentation updates.

### Features:

- Optional pagination for external fields (pagination prop).
- Dynamic column sorting via new sortableColumns prop.
- Outline style for Button.

### Details:

- Added pagination and sortableColumns to external fields (no breaking changes).
- Introduced Pagination component (fully responsive).
- Refactored ExternalInput for enhanced pagination/sorting logic.
- Improved UX with page navigation, sorting indicators, and state persistence.
- New CSS for outline button variant.
- Updated documentation and API reference.

### Motivation:

- Improve handling of large external data sets and make the experience more ergonomic.

### Exemple:

Capture d’écran 2025-10-20 à 16 44 52

PR #1388, opened 2025-10-22
feat: enhance ExternalInput with pagination and sortable columns support + improve docs

hugodebard:
This is a re-created PR following the closure of [#1382](https://github.com/puckeditor/puck/pull/1382), which was closed due to implementation issues and limitations in the approach.
This new PR keeps the core features from the previous one, but fully integrates the feedback received — particularly regarding server-side pagination and sorting.

### Server-side pagination:
- `fetchList` now accepts and uses `page` and `limit` parameters to request paginated data from the server.
- Removed the previous client-side simulation of pagination.
- Integrated pagination state into the cache to maintain consistency and improve performance.

### Server-side sorting:

- Sorting is no longer simulated on the client.
- `fetchList` now accepts a `sort` parameter (`{ column: string; direction: "asc" | "desc" }`) to enable server-side sorting at the data source.
- Sorting state is also cached for consistency across interactions.

### UI improvements:

- Pagination controls reflect actual server-side data boundaries.
- Sorting indicators are now tied to actual sort order from the server.

### Styling and components:

- Introduced a responsive `Pagination` component.
- Added an outline style variant for `Button`.

### Documentation:

- Updated the API reference to include new props and expected `fetchList` behavior.
- Added examples for server-side pagination and sorting configuration.

### Demo:
https://github.com/user-attachments/assets/261a1246-6533-4394-9895-71a06353ff1c

### DIFF
--- a/apps/docs/pages/docs/api-reference/fields/external.mdx
+++ b/apps/docs/pages/docs/api-reference/fields/external.mdx
@@ -780,7 +780,7 @@ The initial number of items per page to display when the modal opens. If not set
 
 Show a search input, the value of which will be passed to `fetchList` as the `query` param.
 
-```tsx {16} copy
+```tsx {17} copy
 const config = {
   components: {
     Example: {


FedericoBonel:
Hey @hugodebard, thanks for reviewing my comments and opening this again 🫡.

Just for anyone reviewing this, these were the comments from the last PR that was closed:

> Hello @hugodebard,
> 
> Thanks for the contribution 🙏! I have a couple of comments we should address before proceeding with a more thorough review.
> 
> ### Pagination Approach
> The current implementation doesn’t actually perform pagination, it simulates it on the client side. Typically, pagination happens on the server, with the client requesting each page as needed. With the API and implementation you're proposing, we would still fetch all available data from the server, then paginate through it on the client.
> 
> This creates a few problems:
> 
> * If the content backend has a default `page` and `limit` (e.g., page 1 and limit 50), `fetchList` will only retrieve the first 50 items. The modal would display pagination controls, but you'd only be paging through those 50 items from the first page.
> * If the backend has no pagination defaults, the situation could be worse. We would be encouraging users to fetch all items at once. If the server contains, say, 500,000 items, this would result in fetching hundreds of megabytes, or even gigabytes, of data, which is not viable.
> 
> **Suggestion:** Could we instead pass `page` and `limit` to the `fetchList` function, just like we already do with filters and the search query? We’d also need to integrate these into the internal cache (similar to how the other PR handles pagination). This would allow users to control which page to fetch and would also simplify your changes to the `ExternalField` component, removing the need for simulated paging.
> 
> ### Sorting Approach
> Similar to pagination, sorting is currently being simulated. We'd recommend extending the `fetchList` function to accept sorting parameters as well, and storing them in the cache.
> 
> This should also be documented (how we express the sorting order).
> 
> Once we’ve addressed these points, we’ll take a deeper look at the PR. Thanks again for your effort, work, and contribution!

apps/docs/pages/docs/integrating-puck/external-data-sources.mdx
```diff
@@ -12,38 +12,7 @@ If you want to provide the user a way to select the data, you can use the [`exte
...
-      return (
-        <>
-          <b>{data.title}</b>
-          <p>{data.description}</p>
-        </>
-      );
-    },
-
-}}
-/>
```
chrisvxd: Is there a reason for removing this example?

FedericoBonel:
Hey @hugodebard!

I just noticed you closed this and a few other PRs, you're absolutely free to do that, of course! I just wanted to check in and see if it was intentional or if there was a specific reason.

Thanks again for your time and contributions 🙏

Issue #1394, opened 2025-10-23
Manage expansion state in Outline separately from selection state


wlenig:
## Description

Currently, within the Outline in the editor, the tree's expansion state is coupled to the editor's selection state.

The tree only expands the selected component and its immediate parents, making navigating even shallowly nested component trees somewhat jarring. Deselecting any child component collapses the entire tree (since nothing is selected anymore), which was a slight productivity killer in my own experience, and the catalyst for making this issue.

## Proposal

- Add some "expanded" state to `Layer`s inside `LayerTree`, and update the styling of the chevron and children to use the new expanded state rather than `isSelected`
- Update the list items' `onClick`s to update both the selection state and update state

I appreciate this approach because the expanded state is bound to the lifetime of each `Layer`, instead of managing some complex parallel data structure that describes outline expansion state, or otherwise intermingling UI state with data store.

chrisvxd:
Thanks for the ticket (and PR) @wlenig! This is a change in core behaviour that is probably unexpected for many Puck users. 

We plan to redo the outline completely in #730 - feel free to drop a suggestion in there. I'm going to close this (and the PR).

PR #1395, opened 2025-10-23
feat: Outline expansion state independent from selection state

wlenig:
Addresses #1394 

Decouples outline tree expansion state from editor selection state

## Changes

- Added local `isExpanded` state to `Layer` components (defaults `false`)
- Updated chevron rotation and children visibility to use `isExpanded` instead of `isSelected`/`childIsSelected`
- Made chevron clickable to toggle expansion independently of selection
- Clicking a selected item now deselects and collapses it (and only it!)

## Before
Only the selected component and its parents would be expanded in the component outline. The tree would collapse entirely when deselecting child components.

## After
Tree expansion occurs independently of selection state, so you can select different components without affecting others' expansion state.

### DIFF
--- a/packages/core/components/LayerTree/index.tsx
+++ b/packages/core/components/LayerTree/index.tsx
@@ -78,7 +78,7 @@ const Layer = ({
     config.components[nodeData.data.type];
   const label = componentConfig?.["label"] ?? nodeData.data.type.toString();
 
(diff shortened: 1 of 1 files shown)


chrisvxd:
Closing per https://github.com/puckeditor/puck/issues/1394#issuecomment-3443140635.

PR #1398, opened 2025-10-24
docs: cleanup + homogenization

hugodebard:
This pull request performs a general cleanup and homogenization of the documentation files.


hugodebard:
> Thanks @hugodebard! This generally looks great, but I would rather turn `showeLineNumbers` _on_ for everything, rather than turning them off.
> 
> I would prefer to do that automatically, but can't see of a way to do it. Could you either remove those changes, or updated all instances to include `showLineNumbers`?

I made a choice to make the whole docs homogeneous, I had to decide between with or without `showLineNumbers`. I understand that it is better with, I will take care of it soon to correct it on the whole docs.

hugodebard:
@chrisvxd Done.

PR #1404, opened 2025-10-28
Feature/puck issue 1403 custom overrides

shannonhochkins:
A simple change to allow "custom" to be overwritten from the overrides section.

Closes: [issue](https://github.com/puckeditor/puck/issues/1403)

Now you can simply provide an override for the "custom" field type, and then re-render it manually:

```
export const createPuckOverridesPlugin = () => {
  return {
    overrides: {
      fieldTypes: {
        custom:(props) => {
             return <>Field for: {props.name} -  {props.field.render(props)}</>
        },
    },
  };
};
```

### DIFF
--- a/packages/core/components/AutoField/index.tsx
+++ b/packages/core/components/AutoField/index.tsx
@@ -221,10 +221,9 @@ function AutoFieldInternal<
     return (_props: any) => null;
   }, [field.type]);
 
-
   let FieldComponent: React.ComponentType<any> = useMemo(() => {
     // if there's an override provided for custom fields, fallback to standard behavior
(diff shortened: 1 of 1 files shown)


packages/core/components/AutoField/index.tsx
```diff
@@ -158,6 +158,7 @@ function AutoFieldInternal<
   const render = useMemo(
     () => ({
       ...overrides.fieldTypes,
+      custom: overrides.fieldTypes?.custom,
       array: overrides.fieldTypes?.array || defaultFields.array,
```
shannonhochkins, line 162: This was added just to satisfy typescript ^

PR #1408, opened 2025-10-30
Add parent data to resolve data params

### LINKED ISSUE #1359: Add `params.parent` to `resolveData`
## Description

Currently, you can use the `params.parent` prop in the `resolveFields` API, which is useful for defining specific fields when a component is nested in a particular component. 

It would also be helpful to add this prop to the `resolveData` API, allowing you to define specific component props when they are dropped inside a specific slot. This would be especially powerful if the prop is added alongside its inclusion in `resolvePermissions`, as requested in #1338.

## Considerations

- Issue #1338 requests this same prop in the `resolvePermissions` API, which could complement this proposal.

## Proposals

Adding the `params.parent` prop to `resolveData` might look like the following:

```tsx
resolveData: (data, params) => {
  if (params.parent.type === "Grid") {
    return { props: { title: "Hello from Grid child" } };
  }
  return data;
}
```

FedericoBonel:
Closes #1359.

This PR adds parent data to `resolveData`, matching the behavior of `resolveFields` as required in #1359.

## Changes made

- The `resolveData` config type now includes a `parent`.
  - The app store `resolveComponentData` was updated to retrieve the parent data from state and pass it to the lib `resolveComponentData` function.
  - The lib `resolveComponentData` was updated to receive the parent data from its caller and pass it to the config-defined function.
  - The lib `resolveAllData` was updated to pass parent data on each recursive call to the config-defined function.
    - `resolveAllData` now fully iterates top to bottom. Previously, this was done for slots but not for zones, which made efficient parent resolution difficult. A new utility, `groupZonesByComponent`, was introduced to handle this.
    - `resolveAllData` was also updated to fix a bug where unresolved `zones` promises were returned in the data.
    - `resolveAllData` now triggers all child `resolveData` calls in parallel, for both zones and slots.
- The `resolveData` config type now includes a `moved` trigger, and the store's `resolveComponentData` now runs when components are moved between slots or dropzones.
  - The `resolveComponentData` cache now stores the parent ID for each execution and re-executes when it changes (for `moved` triggers).
  - A new hook, `useResolveDataOnMoved`, was added to track component drops and trigger `resolveComponentData` with a `moved` trigger. The cache ensures it only runs if the parent ID changed.
- Added documentation for the new `parent` param, `moved` trigger, and updated execution flow.
- Added tests for all new behavior.

## Manual tests

- Confirmed that `resolveData` runs the same number of times on editor load, insertion, deletion, replacement, and reordering within the same slot, using sample data and config.
- Confirmed that `resolveData` runs once per component when a component is reordered across slots (i.e. its parent changes).
- Confirmed that conditional data resolution based on the parent works as expected:

```tsx
resolveData: (data, params) => {
  if (params.parent.type === "Grid") {
    return { props: { title: "Hello from Grid child" } };
  }
  return data;
}
```

https://github.com/user-attachments/assets/f1d9eaca-4a71-4f42-ab2c-a4a23075b281

### DIFF
--- a/packages/core/components/DragDropContext/index.tsx
+++ b/packages/core/components/DragDropContext/index.tsx
@@ -348,7 +348,7 @@ const DragDropContextClient = ({
                   );
                 }
               }
-              
+
               dispatch({
                 type: "setUi",
                 ui: {
@@ -356,7 +356,7 @@ const DragDropContextClient = ({
                   isDragging: false,
                 },
               });
-              
+
               dragListeners.dragend?.forEach((fn) => {
                 fn(event, manager);
               });


apps/docs/pages/docs/api-reference/configuration/component-config.mdx
```diff
@@ -355,10 +355,10 @@ const config = {
...
 #### Args
 
-| Prop     | Example                                                  | Type   |
-| -------- | -------------------------------------------------------- | ------ |
-| `data`   | `{ props: { title: "Hello, world" }, readOnly: {} }`     | Object |
-| `params` | `{ changed: { title: true }, metadata: { foo: "bar" } }` | Object |
+| Prop     | Example                                                                                          | Type   |
+| -------- | ------------------------------------------------------------------------------------------------ | ------ |
+| `data`   | `{ props: { title: "Hello, world" }, readOnly: {} }`                                             | Object |
+| `params` | `{ changed: { title: true }, metadata: { foo: "bar" }, parent: { type: "Heading", props: {} } }` | Object |
```
chrisvxd: These examples don't need to include _everything_. It's more important they just give a high-level overview. I'd prefer to keep them tight onto one line. I'd probably just do `parent: {}`

apps/docs/pages/docs/api-reference/configuration/component-config.mdx
```diff
@@ -408,12 +408,19 @@ const resolveData = async ({ props }, { metadata }) => {
 };
 ```
 
+##### `params.parent`
+
+The [component data](/docs/api-reference/data-model/component-data) of the parent.
+
+Resolvers are executed bottom-up when [`<Puck>`](/docs/api-reference/components/puck) renders, so the parent component will only [resolve its data](#resolvedatadata-params) after the child resolver has completed when this happens.
```
chrisvxd:
```suggestion
If the parent implements [resolveData](#resolvedatadata-params), this parameter will return the unresolved data on initial render, as resolvers are executed bottom-up.
```

apps/docs/pages/docs/api-reference/configuration/component-config.mdx
```diff
@@ -408,12 +408,19 @@ const resolveData = async ({ props }, { metadata }) => {
...
+
 ##### `params.trigger`
 
 The event that triggered this resolveData call. One of:
 
 - `insert` - the component was inserted
 - `replace` - the component was replaced, such as when a field value is changed
+- `moved` - the component moved between parents
```
chrisvxd: This should be `move` not `moved` to align with the other triggers

packages/core/lib/use-resolve-data-on-moved.ts
```diff
@@ -0,0 +1,45 @@
...
+import { useAppStoreApi } from "../store";
+import { getSelectorForId } from "./get-selector-for-id";
+import { rootDroppableId } from "./root-droppable-id";
+import { getItem } from "./data/get-item";
+
+export const useResolveDataOnMoved = (
+  appStore: ReturnType<typeof useAppStoreApi>
+) => {
+  useEffect(() => {
+    return appStore.subscribe(
```
chrisvxd: Subscribing to the store here feels like an inappropriate pattern.

Instead, I think this would be better off being explicitly called in DragDropContext [after the move dispatch](https://github.com/puckeditor/puck/blob/main/packages/core/components/DragDropContext/index.tsx#L359-L366)

chrisvxd (changes requested):
This seems to work well! Great work!

I've left a handful of comments, and I have some feedback on the scope of the PR.

I'm seeing the following changes:

1. Addition of `parent` to `resolveData` (feature 1)
2. Triggering of resolvers whenever an item is moved (feature 2)
3. Various refactors and bug fixes

Combining multiple features, refactors and bug fixes did make reviewing a little difficult as I couldn't easily tell which changes were related to the core functionality and which were fixes or refactors without stepping through the commits.

Breaking down of the commits was really helpful for reviewing, but since some of the conventional commit entries weren't quite right (some incorrect labelling), we can't merge the PR without squashing the history, but we also can't squash because you have multiple features and fixes that we need to track separately in the changelog.

As a rule of thumb, one feature, fix, or refactor per PR is good because we can easily squash. The alternative is to get really good at conventional commits, but this requires getting comfortable with rewriting history (happy to talk through that).

apps/docs/pages/docs/api-reference/configuration/component-config.mdx
FedericoBonel: Ah great, makes sense! I changed it

packages/core/lib/use-resolve-data-on-moved.ts
FedericoBonel: Makes complete sense. I did consider adding it there but assumed we wanted to keep that part lean and follow the same approach as `resolvePermissions` or `resolveFields`.

But yeah, in retrospect, 100% agree, that might also remove the need to check if cached parent was null. Thank you!

apps/docs/pages/docs/api-reference/configuration/component-config.mdx
FedericoBonel: Done! Great catch, thank you

FedericoBonel:
Applied the requested changes, only need to go through the commit history to fix it up with you!

FedericoBonel:
Hey @chrisvxd!

I rebased to `main` and rewrote the commits to avoid pushing anything to the changelog that shouldn't be there. It now adds three features: support for `parent` in `resolveAllData` and `resolveData`, and resolving data when swapping parents.

I think this change should probably be merged before the resolver utility in #1427 so I can rebase that one and make any needed changes.

Let me know if this is enough, or if you'd like me to squash the commits further or rewrite them entirely. I get your point about the PR covering multiple things, so I'm totally fine cleaning it up more, just want to make sure I understand what you're looking for (i.e, less more targeted commits or better tagging).

PR #1415, opened 2025-11-02
Add `getParentByChildId` to Puck API and make `dev` script cross-platform

### LINKED ISSUE #1337: useParent is no longer available.
Has the new version of "useParent" been removed?
How can we obtain the parent component by using the id of the current component now?

eyueldk:
Closes #1337

This PR simplifies the existing workaround for retrieving the parent data of a component by ID by introducing a new `getParentByChildId` helper in the Puck API. It also updates the `dev` script to use `rimraf` instead of `rm -rf`, making it compatible with Windows-only environments.

## Changes Made

- Replaced the `rm -rf` command in the `dev` script with `rimraf` for cross-platform compatibility.
  - Added `rimraf` as a dev dependency.
- Extended the `UsePuckData` type to include `getParentByChildId`, which returns the parent component data.
  - Updated the `generateUsePuck` function to implement the new helper.

### DIFF
--- a/packages/core/lib/use-puck.ts
+++ b/packages/core/lib/use-puck.ts
@@ -75,13 +75,10 @@ export const generateUsePuck = (store: PickedStore): UsePuckStore => {
     getSelectorForId: (id) => getSelectorForId(store.state, id),
     getParentById: (id) => {
       const node = store.state.indexes.nodes[id];
-      
       const parentId = node.parentId;
       if (parentId === null) return;
-      
       const parentNode = store.state.indexes.nodes[parentId];
       if (!parentNode) return;
-      
       return parentNode.data;
     },
   };


FedericoBonel:
Hey @eyueldk!

Thanks for opening this. For the record, we normally prefer to have a GitHub issue opened before a PR. 
This allows us to fully understand the feature request or bug, discuss the best path forward, and share any requirements you’d need to meet for your contribution to be merged. You can read more about our contribution process in the [contributing guide](https://github.com/puckeditor/puck/blob/main/CONTRIBUTING.md).

That said, this PR appears to address an open issue we already have: #1337, which requests a utility for accessing a component's parent data.

I don’t think exposing the node itself makes much sense for end users. It includes a lot of internal details (like `flattenData`, `zone`, `path`, etc.) that likely wouldn’t be meaningful outside the core and hard to document. Additionally, as Puck evolves, we may add more internal data to the node object that we wouldn’t want to expose publicly. So, providing a more abstract and targeted utility that clearly communicates the intent to access the parent data makes more sense in this case.

Before I give this a thorough review, would you be open to updating this PR to more directly address the original issue? If so, I’d be happy to give it a deeper look.

FedericoBonel (changes requested):
Check the comment above!

Remember to use this in your PR description: 

```markdown
Closes #1337
```
So that our tooling can automatically close the issue when this gets merged.

eyueldk:
Made an update. Please let me know if that sufficient.

Closes #1337

packages/core/lib/use-puck.ts
```diff
@@ -71,6 +73,7 @@ export const generateUsePuck = (store: PickedStore): UsePuckStore => {
     getItemBySelector: (selector) => getItem(selector, store.state),
     getItemById: (id) => store.state.indexes.nodes[id].data,
     getSelectorForId: (id) => getSelectorForId(store.state, id),
+    getParentById: (id) => store.state.indexes.nodes[id].data,
```
FedericoBonel: Hm, this is just returning an item by ID, similar to what `getItemById` above is doing. I think we should update it to work as follows:

1. Rename it to `getParentByChildId`.
2. Use the provided ID to get the child node.
3. Retrieve the parent ID from the child node.
4. Use that parent ID to get the parent node from `indexes.nodes`.
5. Return the parent node's data.

Keep in mind that any of these steps could result in `undefined` or `null`, so be sure to validate that each item (child node, parent ID, and parent node) exists before proceeding.

eyueldk: woops. My bad.

eyueldk: On second thought, should I return the data or the parentId? Technically the data can be derived by calling again getItemById. I'm thinking, what is the minimal function needed. Maybe the developer might be just interested in the id, thus not needing to do a second lookup of the data. What do you think?
Should we:
A) lookup the parentId then lookup the node's data 
B) lookup just the parentId and return it

chrisvxd: Chiming in here, I think it's okay to return the whole parent data as this will be most common use-case.

eyueldk: Ok. I'll perform the lookup.

packages/core/lib/use-puck.ts
```diff
@@ -71,6 +73,12 @@ export const generateUsePuck = (store: PickedStore): UsePuckStore => {
     getItemBySelector: (selector) => getItem(selector, store.state),
     getItemById: (id) => store.state.indexes.nodes[id].data,
     getSelectorForId: (id) => getSelectorForId(store.state, id),
+    getParentById: (id) => {
+      const node = store.state.indexes.nodes[id];
+      const parentId = node.parentId;
+      if (parentId === null) return;
+      return store.state.indexes.nodes[parentId].data;
```
FedericoBonel:
```suggestion
      const node = store.state.indexes.nodes[id];
      
      const parentId = node.parentId;
      if (parentId === null) return;
      
      const parentNode = store.state.indexes.nodes[parentId];
      if (!parentNode) return;
      
      return parentNode.data;
```

Could you also move this function into a module under `/lib`, similar to how we’re handling `getSelectorForId`?

eyueldk:
Applied your suggestions.

FedericoBonel:
Epic! Thank you. Now Prettier is failing, would you mind running Prettier and committing the output?

eyueldk:
Build passed

chrisvxd (approved):

chrisvxd:
Thanks @eyueldk! 🎉

PR #1418, opened 2025-11-09
feat: add "as" API to change slot element type

### LINKED ISSUE #965: Change DropZone/Slot element type during render
Before 0.18, DropZones used divs in edit mode (`<Puck>`), but didn't a fragment in render mode (`<Render>`). This was a bug, and never intentional. In 0.18 we addressed this behaviour to ensure consistent rendering between both environments. 

Some users relied on that difference in behaviour. We should explore ways to address this need.

## Considerations

* Default behaviour should be consistency between Puck and Render modes.
* This is a feature request, not a bug fix

### Proposal 1 - Add an `as` prop to the DropZone

We can't use a fragment in <Puck>, so consider an `as` prop to change the underlying element

```tsx
<DropZone as="table" />
```

### Proposal 2 - Add render only prop

We could also consider something similar to your proposal, but ensure it only triggers inside <Render>

```tsx
<DropZone renderAs="" /> {/* Blank implies fragment */}
<DropZone inlineRender /> {/* Alternative, uses fragment inside <Render> */}
```

### Proposal 2 - Leverage slots

We might also be able to take advantage of the new `slots` API that is being introduced as part of #255, but it's not entirely clear what the API would be yet. This suggestion combines slots and DropZones to achieve different rendering results.

```tsx
fields: {
  children: {
    type: "slot",
    mode: "inline" // Current working implementation treats all slots as a render function, which would not solve the problem alone. Introducing a `mode` to slot fields could tell Puck to render the DropZone children inline as a ReactNode, instead of a render function
  },
},
render: ({ children, puck }) => {
  if (puck.isEditing) {
    return <DropZone zone="children" />;
  }

  return <table>{children}</table>;
}
```

---

Original message:

> Hello,
> 
> I have been looking into new Puck Editor changes in 18.0.0 however there is one big issue with our system that we use for Email rendering. We are using Puck editor on Email templating system, however the introduced changes makes upgrading to new version imposible for us.
> 
> Our system uses dropzones on email compatible HTML (Outlook 2016 and up), that means we do not use div elements in email code and use plain table /table html structures. We use Puck to insert tables to other tables and elements however the introduction of forced div element breaks our system as we now cannot insert table inside table without having additional unnecessary div elements that bloat our previously efficient and compatible code and breaks email clients. We are fully relating on fragments without additional div elements.
> 
> Please add a property to <DropZone useFragment={true} > that will on render use a  fragment instead of forced div as an option. 
> 
> `<DropZone useFragment={true} >`

chrisvxd:
Closes #965

### DIFF
--- a/apps/demo/config/blocks/Grid/index.tsx
+++ b/apps/demo/config/blocks/Grid/index.tsx
@@ -13,6 +13,10 @@ export type GridProps = {
   items: Slot;
 };
 
+const CustomSlot = (props: any) => {
+  return <span {...props} />;
+};
+
 export const GridInternal: ComponentConfig<GridProps> = {
   fields: {
     numColumns: {
@@ -39,6 +43,7 @@ export const GridInternal: ComponentConfig<GridProps> = {
     return (
       <Section>
         <Items
+          as={CustomSlot}
           disallow={["Hero", "Stats"]}
           className={getClassName()}
           style={{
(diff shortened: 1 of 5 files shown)


FedericoBonel (approved):
Epic!

Tested it out locally with a component with `useState`, it seems to work well!

PR #1432, opened 2025-11-14
Export reducer for internal use

chrisvxd:


FedericoBonel:
I think this looks good, should we add a warning if you try to access __private or internal?

PR #1444, opened 2025-11-26
fix: memoize assignRefs to prevent max-depth error

### LINKED ISSUE #1440: Using refs for slots with inline mode doesn't work in the Puck AI canary
## Description

When using inline mode in components in the Puck AI canary and pass the `dragRef` to a slot render function, Puck throws an error. 
This does not happen in the latest stable Puck version.

## Environment

- Puck version: 0.21.0-canary.c0db75c1

## Steps to reproduce

1. Render the Puck editor with a config that uses a slot field and inline mode, passing the `dragRef` to the slot render function or component:

```tsx
export const config = {
  components: {
    SomeSlot: {
      fields: {
        content: { type: "slot" },
      },
      inline: true,
      render: ({ content: Content, puck }) => (
        <Content style={{ height: "fit-content" }} ref={puck.dragRef} />
      ),
    },
  },
};
```

2. Navigate to the Puck editor.  
3. Drag and drop a `SomeSlot` component.

## What happens

An error is thrown:

```bash
Maximum update depth exceeded. This can happen when a component repeatedly calls setState inside componentWillUpdate or componentDidUpdate. React limits the number of nested updates to prevent infinite loops.
```

This does not happen when the `dragRef` is passed to a different component or element that is not the slot render function.

## What I expect to happen

The slot should render as expected, the same way it does in the latest stable version.

## Additional Media

https://github.com/user-attachments/assets/bbac20af-6804-4458-93c1-b37057d5d48f

xaviemirmon:
Closes #1440 

## Description

This PR adds fixes the max-depth error when adding a drag ref to slots on the canary.

```
Maximum update depth exceeded. This can happen when a component repeatedly calls setState inside componentWillUpdate or componentDidUpdate. React limits the number of nested updates to prevent infinite loops.
```

## Changes made

-  Wrapped assignRefs for slots in a useCallback.

## How to test

- Render the Puck editor with a config that uses a slot field and inline mode, passing the dragRef to the slot render function or component:

```tsx
export const config = {
  components: {
    SomeSlot: {
      fields: {
        content: { type: "slot" },
      },
      inline: true,
      render: ({ content: Content, puck }) => (
        <Content style={{ height: "fit-content" }} ref={puck.dragRef} />
      ),
    },
  },
};
```
Drag and drop a SomeSlot component and see no errors.

### DIFF
--- a/packages/core/components/DropZone/index.tsx
+++ b/packages/core/components/DropZone/index.tsx
@@ -463,6 +463,13 @@ export const DropZoneEdit = forwardRef<HTMLDivElement, DropZoneProps>(
       ref,
     });
 
+    const setRefs = useCallback(
+      (node: any) => {
+        assignRefs<any>([ref, dropRef, userRef], node);
+      },
+      [dropRef]
+    );
+
     const El = as ?? "div";
 
     return (
@@ -475,9 +482,7 @@ export const DropZoneEdit = forwardRef<HTMLDivElement, DropZoneProps>(
           hasChildren: contentIds.length > 0,
           isAnimating,
         })}${className ? ` ${className}` : ""}`}
-        ref={(node: any) => {
-          assignRefs<any>([ref, dropRef, userRef], node);
-        }}
+        ref={setRefs}
         data-testid={`dropzone:${zoneCompound}`}
         data-puck-dropzone={zoneCompound}
         style={


FedericoBonel (approved):
Great stuff! Thanks 🙏

PR #1447, opened 2025-11-27
Eliminate re-rendering of components when hovered in the canvas

### LINKED ISSUE #1442: Slot components re-render on hover
## Description

If we use slots to nest components, whenever a slot or its nested content is hovered, the entire tree re-renders.  
This happens because the component’s render function is not memoized in `DropZoneChild` (both in the [definition here](https://github.com/puckeditor/puck/blob/af48a4011d9a1ff589b5465b9a2684e4ba00f6aa/packages/core/components/DropZone/index.tsx#L221) and the invocation [here](https://github.com/puckeditor/puck/blob/af48a4011d9a1ff589b5465b9a2684e4ba00f6aa/packages/core/components/DropZone/index.tsx#L253-L267)).  
As a result, whenever `DropZoneChild` or `DraggableComponent` re-renders, the render function does too.  
In this case, the component that re-renders on hover is `DraggableComponent`.

## Environment

- Puck version: 0.20.2

## Steps to reproduce

1. Render the Puck editor with a config that uses slots:

```tsx
const config = {
  components: {
    Slot: {
      fields: { items: { type: "slot" } },
      render: ({ items: Items }) => {
        console.log("Re-rendering Slot");
        return <Items />;
      }
    },
    Heading: {
      fields: { title: { type: "text" } },
      defaultProps: { title: "Heading" },
      render: ({ title }) => {
        console.log("Re-rendering Heading");
        return <h1>{title}</h1>;
      }
    }
  }
};
```

2. Navigate to the editor.  
3. Open the devtools console.  
4. Drag and drop a `Slot`.  
5. Drag and drop two `Heading` components inside the `Slot`.  
6. Hover over the `Slot` component.  
7. Watch the console.

## What happens

Multiple logs from both the `Slot` and `Heading` components appear.

## What I expect to happen

No logs should appear for either the `Slot` or the `Heading` when hovering over them.

## Additional Media

https://github.com/user-attachments/assets/533e9122-88a8-4ea6-8692-3d3641d62ea4

FedericoBonel:
Closes #1442

## Description

This PR fixes a performance issue where components re-rendered when hovered over in the canvas.

## Changes made

- The `renderPreview` function in `DropZoneChild` was moved to its own component to simplify the code and avoid remounting and unnecessary re-renders.
- The `DropZoneChild` component now uses `MemoizeComponent` to render user components. This component accepts two props: the component to render, and the props to pass to it. It memoizes the rendering by only re-rendering when the props change, using shallow comparison for general props and deep comparison for the `puck` prop. This approach was chosen because user props can be large and deeply nested, making deep equality checks costly. For solving the hover-related re-renders, this level of optimization was enough.
- A new `shallowEqual` utility was introduced to check for shallow equality between two objects.

## How to test

1. Render the Puck editor with a config that uses slots:

```tsx
const config = {
  components: {
    Slot: {
      fields: { items: { type: "slot" } },
      render: ({ items: Items }) => {
        console.log("Re-rendering Slot");
        return <Items />;
      }
    },
    Heading: {
      fields: { title: { type: "text" } },
      defaultProps: { title: "Heading" },
      render: ({ title }) => {
        console.log("Re-rendering Heading");
        return <h1>{title}</h1>;
      }
    }
  }
};
```

2. Navigate to the editor.  
3. Open the devtools console.  
4. Drag and drop a `Slot`.  
5. Drag and drop two `Heading` components inside the `Slot`.  
6. Hover over the `Slot` component.  
7. Watch the console.  
8. Confirm that no re-rendering logs are printed.

https://github.com/user-attachments/assets/72099349-fc6f-4831-9a3c-fbe318dea87f

### DIFF
--- a/packages/core/components/DropZone/index.tsx
+++ b/packages/core/components/DropZone/index.tsx
@@ -52,6 +52,7 @@ import { getSlotTransform } from "../../lib/field-transforms/default-transforms/
 import { getRichTextTransform } from "../../lib/field-transforms/default-transforms/rich-text-transform";
 import { FieldTransforms } from "../../types/API/FieldTransforms";
 import { useRichtextProps } from "../RichTextEditor/lib/use-richtext-props";
+import { MemoizeComponent } from "../MemoizeComponent";
 
 const getClassName = getClassNameFactory("DropZone", styles);
 
@@ -223,7 +224,7 @@ const DropZoneChild = ({
 
   if (!item) return;
 
-  let Render = componentConfig
+  const Render = componentConfig
     ? componentConfig.render
     : () => (
         <div style={{ padding: 48, textAlign: "center" }}>
@@ -253,15 +254,13 @@ const DropZoneChild = ({
       {(dragRef) => {
         if (componentConfig?.inline && !isInserting) {
           return (
-            <>
-              <Render
-                {...transformedProps}
-                puck={{
-                  ...transformedProps.puck,
-                  dragRef,
-                }}
-              />
-            </>
+            <MemoizeComponent
+              Component={Render}
+              componentProps={{
+                ...transformedProps,
+                puck: { ...transformedProps.puck, dragRef },
+              }}
+            />
           );
         }
 
@@ -276,7 +275,10 @@ const DropZoneChild = ({
                 }
               />
             ) : (
-              <Render {...transformedProps} />
+              <MemoizeComponent
+                Component={Render}
+                componentProps={transformedProps}
+              />
             )}
           </div>
         );
--- a/packages/core/components/MemoizeComponent/index.tsx
+++ b/packages/core/components/MemoizeComponent/index.tsx
@@ -1,0 +1,27 @@
+import { deepEqual } from "fast-equals";
+import { ComponentType, memo } from "react";
+import { shallowEqual } from "../../lib/shallow-equal";
+
+const RenderComponent = ({
+  Component,
+  componentProps: renderProps,
+}: {
+  Component: ComponentType<any>;
+  componentProps: any;
+}) => {
+  return <Component {...renderProps} />;
+};
+
+/**　Renders the Component and only re-renders when its props change using shallow comparison. Uses deep comparison for the "puck" prop. */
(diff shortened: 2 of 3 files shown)


packages/core/components/DropZone/index.tsx
```diff
@@ -65,6 +67,25 @@ export type DropZoneDndData = {
...
+  if (element) {
+    return (
+      // Safe to use this since the HTML is set by the user
+      <div dangerouslySetInnerHTML={{ __html: element.outerHTML }} />
+    );
+  }
+
+  return <DrawerItemInner name={label}>{override}</DrawerItemInner>;
+};
+
```
FedericoBonel, line 91: Tempted to pull this into its own component, this file is getting pretty long. Any thoughts?

I'm also really tempted to refactor the file to break it down a bit. I found it hard to navigate when I first looked at it. This second refactor is probably better left for another PR, though.

packages/core/components/MemoizeComponent/index.tsx
```diff
@@ -0,0 +1,27 @@
...
+  if ("puck" in prev.componentProps && "puck" in next.componentProps) {
+    puckEquals = deepEqual(prev.componentProps.puck, next.componentProps.puck);
+  }
+
+  return (
+    prev.Component === next.Component &&
+    shallowEqual(prev.componentProps, next.componentProps, ["puck"]) &&
+    puckEquals
+  );
+});
```
FedericoBonel, line 27: Was thinking that maybe we could rename this to `MemoizeComponentShallow` and export another `MemoizeComponentDeep` as well (that uses `deepEqual` for all props), it might prove useful in the future. 
But I might be overdoing it.

packages/core/components/DropZone/index.tsx
chrisvxd: I think it's fine for now

packages/core/components/MemoizeComponent/index.tsx
chrisvxd: we can do that when we need it

Issue #1456, opened 2025-12-10
Highlighted component - Breaks when component is fixed


shannonhochkins:
## Description

I have a "Popup" component that is fixed position with css, unfortunately when i scroll within the canvas, the "highlighted" area scrolls with it and doesn't respect the actual positioning of the element. 

I believe the problem is with the [getStyle](https://github.com/puckeditor/puck/blob/c78dc826ab201e57b1f807a9ed75d216bd8b7aca/packages/core/components/DraggableComponent/index.tsx#L257), which will work for any normal component, aside from fixed components as it offsets with the scroll position which fixed elements do not need.

## Environment

- Puck version: [0.19.2]
- Browser version: any

## Steps to reproduce

1. Set position: fixed on the draggable component
2. Create overflow on the page
3. Select the component, then scroll, you'll notice the selector highlight isn't respecting the position

## What happens

Highlighted selector isn't tracking component correctly

## What I expect to happen

Highlighted selector should track component

## Suggested fix

```tsx
const getStyle = useCallback(() => {
  const el = ref.current;
  if (!el) return;

  const rect = el.getBoundingClientRect();
  const deepScrollPosition = getDeepScrollPosition(el);

  const portalContainerEl = iframe.enabled
    ? null
    : el.closest<HTMLElement>("[data-puck-preview]");

  const portalContainerRect = portalContainerEl?.getBoundingClientRect();
  const portalScroll = portalContainerEl
    ? getDeepScrollPosition(portalContainerEl)
    : { x: 0, y: 0 };

  const scroll = {
    x:
      deepScrollPosition.x -
      portalScroll.x -
      (portalContainerRect?.left ?? 0),
    y:
      deepScrollPosition.y -
      portalScroll.y -
      (portalContainerRect?.top ?? 0),
  };

  const untransformed = {
    height: el.offsetHeight,
    width: el.offsetWidth,
  };

  const transform = accumulateTransform(el);
  const computedStyle = window.getComputedStyle(el);
  // for fixed elements, we shouldn't offset with scroll, we should instead match
  // the coordinate system
  const isFixed = computedStyle.position === "fixed";

  const offsetX = isFixed ? 0 : scroll.x;
  const offsetY = isFixed ? 0 : scroll.y;

  const style: CSSProperties = {
    left: `${(rect.left + offsetX) / transform.scaleX}px`,
    top: `${(rect.top + offsetY) / transform.scaleY}px`,
    height: `${untransformed.height}px`,
[...]

PR #1461, opened 2025-12-11
Puck issue 1456 overlay position

shannonhochkins:
Closes [#1456](https://github.com/puckeditor/puck/issues/1456)

## Description

Fix overlay positioning to fully cover components even when parents/targets apply CSS transforms or fixed positioning. We now rely on the browser’s transformed bounding boxes, add fixed-target handling, and resync overlays when style/class or geometry changes. Removed the legacy inverse-scale correction because getBoundingClientRect already reports transformed sizes/positions, so dividing by accumulated transforms caused misalignment under parent transforms.

## Changes made

- Use the element’s transformed getBoundingClientRect plus scroll offsets to set overlay left/top/width/height; drop accumulateTransform inverse-scaling
  - getBoundingClientRect() already returns the post-transform box: position, width, and height are in the final rendered coordinate space, including all parent transforms (scale/translate/rotate). Dividing by an accumulated scale double-corrects and shrinks/shifts the overlay when parents are transformed.
- Detect fixed ancestors and set overlay position: fixed while skipping scroll offsets in that case.
- Add a MutationObserver watching style and class to resync overlays when styles change.
- Add a lightweight rAF rect watcher to resync when geometry moves due to transforms (e.g., carousel track translate/scale) without waiting for scroll/resize.
- Keep ResizeObserver and ref reattachment logic; sync now depends only on getStyle.
- Supports css transforms completely (scale, rotation, translation, even skew) (example screenshot below, try the same css in production ;) )

image

## How to test

- Update inline css of any component, change to position fixed, overlay should be positioned correctly
- Update inline css of any component, or even a parent, add css transform `scale3d(0.79, 0.79, 1) translate3d(20%, 0, 0);` and you should see the overlay move to match the new position when hovering, even other sub components if added to a parent with transforms like scale3d(0.79, 0.79, 1) translate3d(20%, 0, 0); overlay should match the rendered element bounds.
- Set a draggable ancestor to position: fixed; overlay should also be fixed and align while scrolling.
- Apply style/class changes (e.g., add/remove CSS that shifts size/position) and confirm the overlay updates immediately.
- transitions that manipulate inline styles even not on the target will now respect the new position and overlays will update (carousel example)
- Test disableIframe=true to ensure all the above still works when no iframe is provided

I have done EXTENSIVE testing on this and it's pretty solid! Can't find any more issues

- Multiple tests were done, comparing production to this branch, screen recordings below: 

https://github.com/user-attachments/assets/590a3312-990c-440c-97f4-382a2205ec16

^ Fixes a bug in production in the puck demo that's present

And here's the difference in my software:

- EDIT - This screen recording was taken before i added the raf helper, you can see the overlay sometimes jumps when the carousel moves, until you move your cursor or scroll/resize - this is fixed now, additionally the "fixed" popup overlay that animates, is also patched and now is pinned to the element directly (this was a separate side effect i removed with a way over-engineered solution) - now much simpler and very little changes to puck

https://github.com/user-attachments/assets/2d4699ab-9c19-4290-b7cf-e63bb2cc2f3d

### DIFF
--- a/packages/core/components/DraggableComponent/index.tsx
+++ b/packages/core/components/DraggableComponent/index.tsx
@@ -563,52 +563,52 @@ export const DraggableComponent = ({
   // PERFORMANCE: when visible, respond to scroll/resize + track layout shifts without a global rAF loop
   useEffect(() => {
     if (!dragFinished || !(isSelected || thisIsDragging)) return;
-  
+
     const el = ref.current;
     if (!el) return;
-  
+
     const doc = el.ownerDocument;
     const view = doc.defaultView;
     if (!view) return;
-  
+
     lastMeasureRef.current = 0;
     scheduleSync(); // immediate position on show
-  
+
     const onScroll = () => scheduleSync();
     const onResize = () => scheduleSync();
-  
+
     doc.addEventListener("scroll", onScroll, true);
     view.addEventListener("resize", onResize);
-  
+
     let frame = 0;
     const tick = (t: number) => {
       if (t - lastMeasureRef.current >= MEASURE_EVERY_MS) {
         lastMeasureRef.current = t;
-  
+
         const node = ref.current;
         if (node) {
           const rect = node.getBoundingClientRect();
           const prev = lastRectRef.current;
-  
+
           const changed =
             !prev ||
             Math.abs(rect.x - prev.x) > 0.5 ||
             Math.abs(rect.y - prev.y) > 0.5 ||
             Math.abs(rect.width - prev.width) > 0.5 ||
             Math.abs(rect.height - prev.height) > 0.5;
-  
+
           if (changed) {
             lastRectRef.current = rect;
             scheduleSync();
           }
         }
       }
-  
+
       frame = requestAnimationFrame(tick);
     };
-  
+
     frame = requestAnimationFrame(tick);
-  
+
     return () => {
       doc.removeEventListener("scroll", onScroll, true);
       view.removeEventListener("resize", onResize);


packages/core/components/DraggableComponent/index.tsx
```diff
@@ -311,7 +324,61 @@ export const DraggableComponent = ({
...
+  useEffect(() => {
+    if (!ref.current) return;
+
+    const observer = new MutationObserver((mutations) => {
+      if (
+        mutations.some((mutation) =>
+          ["style", "class"].includes(mutation.attributeName || "")
+        )
+      ) {
+        sync();
```
chrisvxd: We should probably put this inside a `requestAnimationFrame`

packages/core/components/DraggableComponent/index.tsx
```diff
@@ -257,49 +256,63 @@ export const DraggableComponent = ({
...
+            (portalContainerRect?.top ?? 0),
+        };
 
     const style: CSSProperties = {
-      left: `${(rect.left + scroll.x) / transform.scaleX}px`,
-      top: `${(rect.top + scroll.y) / transform.scaleY}px`,
-      height: `${untransformed.height}px`,
-      width: `${untransformed.width}px`,
+      left: `${rect.left + scroll.x}px`,
+      top: `${rect.top + scroll.y}px`,
```
chrisvxd, line 303: Hm bit confused why the transforms have been removed here, and how they're getting applied

packages/core/components/DraggableComponent/index.tsx
```diff
@@ -311,7 +324,61 @@ export const DraggableComponent = ({
...
+          Math.abs(rect.height - prev.height) > 0.5;
+
+        if (changed) {
+          lastRectRef.current = rect;
+          sync();
+        }
+      }
+
+      frame = requestAnimationFrame(tick);
+    };
```
chrisvxd: I'm not sure this animation loop is viable. The demo app now idles my MBP (M1 Max) CPU utilisation at 10%, up from under 2%. Can you find another solution?

chrisvxd (changes requested):
Sorry for slow review. Can confirm it works, but have performance concerns with the animation loop. See comment.

packages/core/components/DraggableComponent/index.tsx
shannonhochkins: I did explain this in the main description of the PR a little more in detail, but a bounding box value is provided with transforms applied, including its parent(s), so theres no need to apply the transforms here

packages/core/components/DraggableComponent/index.tsx
shannonhochkins: We can throttle it, the main benefit of the RAF is animations realistically, but theres something many other operations thst weren't not listening to that can affect position so i added the raf, would a throttled raf suffice?

shannonhochkins:
@chrisvxd I've made a few changes, are you able to check them again for me?

**Perf change:** removed the always-on overlay sync loop and switched to event-driven + throttled measurement.

Previously, each DraggableComponent started a requestAnimationFrame loop that ran continuously, even when the app was otherwise idle. That meant we were calling getBoundingClientRect() every frame and occasionally updating overlay state, which kept the renderer active and raised baseline CPU usage.

In this update:

- Overlay sync is now coalesced via scheduleSync() (single rAF per frame). This prevents multiple rapid triggers from causing multiple React updates.

- Resize handling stays event-driven via ResizeObserver, which calls scheduleSync() when the element’s dimensions change.

- Scroll / window resize are handled via listeners (scroll uses capture to catch nested scroll containers), triggering scheduleSync() only when those events occur.

- To still support position changes caused by animations/layout shifts/transforms (i.e. movement that doesn’t emit a scroll/resize/resizeObserver event), we added a throttled rect sampling loop while the component is active (measuring at ~10fps via MEASURE_EVERY_MS). When a rect delta is detected, we trigger scheduleSync().

**Net effect:** we keep overlay positioning correct during real movement (scroll/resize/animations), but avoid the previous “always hot” 60fps loop that increased idle CPU.

The throttled measurement loop is not started on hover, only when selected/dragging, to avoid many components spinning up trackers as the mouse moves through the editor.

chrisvxd:
I haven't yet reviewed the latest changes, but wanted to know if you'd considered using https://github.com/Shopify/position-observer for this?

shannonhochkins:
> I haven't yet reviewed the latest changes, but wanted to know if you'd considered using https://github.com/Shopify/position-observer for this?


Haven't seen that before, happy to wire it in if you'd prefer

shannonhochkins:
@chrisvxd can this be merged the way it is without me having to wire up this other package? It works as is and I dont know if this package is overkill for this bugfix

PR #1479, opened 2025-12-22
fix: prevent empty objects from causing parent re-render

### LINKED ISSUE #1471: Inline RichText field does not update within slots
## Description

When an item inside a slot contains a rich text field with `contentEditable: true`, the inline mode does not accept input and causes the browser to crash.

## Environment

- Puck version: 0.21 canary

## Steps to reproduce

1. Configure a slot and child, giving the child a richtext field with contentEditable

```tsx
const config = {
  components: {
    Parent: {
      fields: {
        content: {
          type: "slot",
        },
      },
      render: ({ content: Content }) => {
        return <Content />;
      },
    },
    Child: {
      fields: {
        title: { type: "richtext", contentEditable: true }
      },
      defaultProps: { title: "<h2>Title</h2>" },
      render: ({ title }) => <div>{title}</div>,
    },
  },
};
```

2. Nest the components
3. Try to edit the title field of the child inline

## What happens

The value does not update, and the overlay flickers, indicating a re-render.

## What I expect to happen

The value updates and the overlay does not flicker.

chrisvxd:
Closes #1471 

May also address other slot performance issues. Needs validating.

## Description

Upon investigation, certain field configurations on a component resulted in an empty object being returned by the `flattenNode` method. If this component also had a slot, changing any other prop or child prop would trigger a full re-render of the parent because the flattened data is shallow compared by the [selector in DropZoneChild](https://github.com/puckeditor/puck/blob/main/packages/core/components/DropZone/index.tsx#L120). If an object is included in the flattened data, the shallow comparison would always return a new value and trigger a re-render.

This PR addresses this by stripping any empty objects created in the `flattenNode` method.

This issue has existed since the introduction of slots in 0.19.0.

## Changes made

- Stripping of empty objects in the `flattenNode` method
- Alignment of inline rich text field behaviour with inline text field behaviour

## How to test

1. Add an `array` field to a component, but don't add `defaultItemProps` (as this will result in an empty object)
2. Add a `slot` field to the same component
3. Add a `Math.random()` to the render method of the parent component
4. Open puck
5. Add a new item to the array field but don't touch the subfields (to trigger an empty object)
6. Add a child component
7. Change a prop on the child component
8. Ensure Math.random() doesn't change

### DIFF
--- a/packages/core/lib/data/flatten-node.ts
+++ b/packages/core/lib/data/flatten-node.ts
@@ -5,6 +5,36 @@ import { stripSlots } from "./strip-slots";
 // Explicitly destructure to account for flat module issues: https://github.com/puckeditor/puck/issues/1089
 const { flatten, unflatten } = flat;
 
+function isEmptyArrayOrObject(val: any): boolean {
+  if (Array.isArray(val)) {
+    return val.length === 0;
+  }
+
+  if (
+    val != null &&
+    Object.prototype.toString.call(val) === "[object Object]"
+  ) {
+    return Object.keys(val).length === 0;
+  }
+
+  return false;
+}
+
+function stripEmptyObjects(props: Record<string, any>) {
+  const result: Record<string, any> = {};
+
+  for (const key in props) {
+    if (!Object.prototype.hasOwnProperty.call(props, key)) continue;
+
+    const val = props[key];
+    if (isEmptyArrayOrObject(val)) continue;
+
+    result[key] = val;
+  }
+
+  return result;
+}
+
 export const flattenNode = <
   UserConfig extends Config = Config,
   G extends UserGenerics<UserConfig> = UserGenerics<UserConfig>
@@ -14,7 +44,7 @@ export const flattenNode = <
 ) => {
   return {
     ...node,
-    props: flatten(stripSlots(node, config).props),
+    props: stripEmptyObjects(flatten(stripSlots(node, config).props)),
   };
 };
 


packages/core/lib/data/flatten-node.ts
```diff
@@ -5,6 +5,19 @@ import { stripSlots } from "./strip-slots";
...
+
+    // Filter out empty objects
+    if (val && typeof val === "object" && Object.keys(val).length === 0) {
+      return acc;
+    }
+
+    return { ...acc, [key]: val };
+  }, {});
+}
+
```
FedericoBonel, line 37:
```suggestion
function stripEmptyObjects(object: Record<string, any>) {
  const result: Record<string, any> = {};

  for (const key in object) {
    const val = object[key];

    // Filter out empty objects
    if (val && typeof val === "object" && Object.keys(val).length === 0) {
      continue;
    }

    result[key] = val;
  }

  return result;
}
```

I think we're better off using a traditional for loop here. 
The reason being that the reduce is creating a new object on every one of the calls using spreading. This has 2 side effects (where `n` is the number of keys in the object):
- A time complexity of `O(n^2)`: For every key, we're going through every other key (from `1`, all the way to `n-1`)
- Memory usage: We're creating `n` new objects (with `1` to `n-1` keys), these will be garbage collected, but it will create a spike on memory, and then it could affect performance since the garbage collector needs to do more work.

I'm not sure how many times this gets called, but I think we should start migrating the reducers where possible (I'd be happy to do this), if I remember correctly this was one of the optimizations that made TanStack table 1000x faster.

Using a for loop in contrast has:
- A time complexity of `O(n)`
- Less memory usage since it only creates a single object

chrisvxd: Adjusted

PR #1480, opened 2025-12-23
perf: don't remount slot when parent changes

### LINKED ISSUE #1208: Performance issues when nesting multiple components in a slot
## Description

Updating a component's prop that uses a slot with multiple nested components causes severe lag in the editor. This is reproducible in the demo.

## Environment

- Puck version: 0.19.3

## Steps to reproduce

1. Navigate to the Puck demo app: https://demo.puckeditor.com/edit
2. Drag and drop a `Grid` component into the editor
3. Populate the grid with multiple `Card` components (it depends on hardware, but more than 60 to 80 should start to show the problem)
4. Change the "gap" field in the grid

## What happens

The editor becomes sluggish and unresponsive when modifying fields in the parent component with the slot that has multiple children.

## What I expect to happen

Changes to any component's fields should apply smoothly, regardless of the number of children rendered in their slots.

## Additional Media

https://github.com/user-attachments/assets/d7331980-ceba-4248-8f27-ec615815439a

chrisvxd:
Closes #1208 

## Description

This PR adds a shallow comparison when transforming props to prevent remounting of slots.

## Changes made

- Added a new "tracked" version of the `useFieldTransform` hook that performs a shallow diff so it only runs on the changed props since last run
- The `mapFields` function now has an opt-out for field defaulting, as it was reintroducing any slots that were removed during the shallow diffing

## How to test

Follow steps outlined in https://github.com/puckeditor/puck/issues/1208

### DIFF
--- a/packages/core/lib/field-transforms/build-mappers.ts
+++ b/packages/core/lib/field-transforms/build-mappers.ts
@@ -1,0 +1,58 @@
+import {
+  ComponentData,
+  Config,
+  ExtractField,
+  Field,
+  UserGenerics,
+} from "../../types";
+import { MapFnParams, Mappers } from "../data/map-fields";
+import {
+  FieldTransformFn,
+  FieldTransforms,
+} from "../../types/API/FieldTransforms";
+
+/**
+ * Converts transformers to mappers
+ *
+ * Transformers are the same as mappers, except they receive the additional `isReadOnly` param.
+ * This converts transformers to mappers by adding the `isReadOnly` param.
+ */
+export function buildMappers<
+  T extends ComponentData,
+  UserConfig extends Config,
+  G extends UserGenerics<UserConfig>
+>(
+  transforms: FieldTransforms,
+  readOnly?: T["readOnly"],
+  forceReadOnly?: boolean
+) {
+  return Object.keys(transforms).reduce<Mappers>((acc, _fieldType) => {
+    const fieldType = _fieldType as Field["type"]; // Not strictly true, as could include user fields, but this should be safe enough
+
+    return {
+      ...acc,
+      [fieldType]: ({
+        parentId,
+        ...params
+      }: MapFnParams<ExtractField<G["UserField"], Field["type"]>>) => {
+        const wildcardPath = params.propPath.replace(/\[\d+\]/g, "[*]");
+
+        const isReadOnly =
+          readOnly?.[params.propPath] ||
+          readOnly?.[wildcardPath] ||
+          forceReadOnly ||
+          false;
+
+        const fn = transforms[fieldType] as FieldTransformFn<
+          ExtractField<G["UserField"], Field["type"]>
+        >;
+
+        return fn?.({
+          ...params,
+          isReadOnly,
+          componentId: parentId,
+        });
+      },
+    };
+  }, {});
+}
--- a/packages/core/lib/field-transforms/use-field-transforms-tracked.tsx
+++ b/packages/core/lib/field-transforms/use-field-transforms-tracked.tsx
@@ -1,0 +1,75 @@
+"use client";
+
+import { ComponentData, Config } from "../../types";
+import { useMemo, useRef } from "react";
+import { mapFields, Mappers } from "../data/map-fields";
+import { FieldTransforms } from "../../types/API/FieldTransforms";
+import { buildMappers } from "./build-mappers";
+
+export function useFieldTransformsTracked<
+  T extends ComponentData,
+  UserConfig extends Config
+>(
+  config: UserConfig,
+  item: T,
+  transforms: FieldTransforms,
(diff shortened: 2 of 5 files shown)


packages/core/lib/field-transforms/build-mappers.ts
```diff
@@ -0,0 +1,52 @@
...
+  ExtractField,
+  Field,
+  UserGenerics,
+} from "../../types";
+import { MapFnParams, Mappers } from "../data/map-fields";
+import {
+  FieldTransformFn,
+  FieldTransforms,
+} from "../../types/API/FieldTransforms";
+
```
FedericoBonel, line 13:
```suggestion

/**
 * Converts transformers to mappers
 * 
 * Transformers are the same as mappers, except they receive the additional `isReadOnly` param.
 * This converts transformers to mappers by adding the `isReadOnly` param.
 */
```

packages/core/lib/field-transforms/use-field-transforms-tracked.tsx
```diff
@@ -0,0 +1,74 @@
...
+  item: T,
+  transforms: FieldTransforms,
+  readOnly?: T["readOnly"],
+  forceReadOnly?: boolean
+): T["props"] {
+  const prevProps = useRef<Record<string, any>>(null);
+  const prevResult = useRef<Record<string, any>>(item.props);
+
+  // Transformers are the same as mappers, except they receive the additional `isReadOnly` param.
+  // This converts transformers to mappers by adding the `isReadOnly` param
```
FedericoBonel:
```suggestion
```
Consolidated in the JSDoc above

packages/core/lib/field-transforms/use-field-transforms.tsx
```diff
@@ -25,42 +16,13 @@ export function useFieldTransforms<
 ): T["props"] {
   // Transformers are the same as mappers, except they receive the additional `isReadOnly` param.
   // This converts transformers to mappers by adding the `isReadOnly` param
```
FedericoBonel:
```suggestion
```
Consolidated in the JSDoc above

packages/core/lib/field-transforms/use-field-transforms-tracked.tsx
```diff
@@ -0,0 +1,74 @@
...
+    const propKeys = Object.keys(item.props);
+    let changeIncludesSlot = false;
+
+    for (let i = 0; i < propKeys.length; i++) {
+      const key = propKeys[i];
+
+      const fieldType = componentConfig?.fields?.[key]?.type;
+
+      if (!prevProps.current || item.props[key] !== prevProps.current[key]) {
+        changedProps[key] = item.props[key];
```
FedericoBonel: Little suggestion to make it more concise and clear:
- Rename `key` to `fieldName`.
- Using a for...in instead of Object.keys and indexes
```tsx
    // Filter to changed fields only (shallow comparison)
    const changedProps: Record<string, any> = {};

    const componentConfig =
      item.type === "root" ? config.root : config.components?.[item.type];

    let changeIncludesSlot = false;

    for (const fieldName in item.props) {
      const fieldType = componentConfig?.fields?.[fieldName]?.type;

      if (
        !prevProps.current ||
        item.props[fieldName] !== prevProps.current[fieldName]
      ) {
        changedProps[fieldName] = item.props[fieldName];

        if (fieldType === "slot") {
          changeIncludesSlot = true;
        }
      }
    }
```

chrisvxd: I generally stick with native `for` for performance and so you can exclude `hasOwnProperty`, but since this is only ever a plain object, it should be fine. Updated.

packages/core/lib/field-transforms/use-field-transforms.tsx
chrisvxd: Done

packages/core/lib/field-transforms/use-field-transforms-tracked.tsx
chrisvxd: Done

packages/core/lib/field-transforms/build-mappers.ts
chrisvxd: Done

chrisvxd:
Fixed the issue @FedericoBonel

packages/core/lib/field-transforms/use-field-transforms-tracked.tsx
FedericoBonel: Ahh makes sense! I just checked and it is quite slower. I thought the overhead wouldn't be that big but the `in` is way worse than the rest.
Hmm we can roll it back then if you want 🫠 sorry for the back and forth.

PR #1493, opened 2026-01-12
Add Puck AI to `create-puck-app`

FedericoBonel:
## Description

This PR adds an option for integrating Puck AI with the existing recipes.

## Changes made

- Added handling of sigterm and sigint for a graceful exit
- Added a check to see if the user provides a name for the app or not
- Added a puck AI option to `create-puck-app`, this option appends an `-ai` to the name of the recipe to spin up and links to docs at the end.
- Added a version with Puck AI integrated for each existing recipe with an `-ai` suffix.

## How to test

- Navigate to the `create-puck-app` package
- Run `yarn generate`
- Spin up an app using `node index.js`
- When prompted use Puck AI
- Navigate to the spinned up project
- Add a Puck API key to `.env`
- Execute the recipe
- Navigate to `/edit` and create a page with Puck AI


packages/create-puck-app/index.js
```diff
@@ -96,13 +101,32 @@ program
           },
         ],
       },
+      {
+        type: "confirm",
+        name: "puckAi",
+        message: "Would you like to use Puck AI?",
```
chrisvxd: Hm, I think we should introduce Puck AI before asking, otherwise the user has no idea what this is

packages/create-puck-app/index.js
```diff
@@ -201,6 +225,16 @@ program
...
+
+    if (usesPuckAi) {
+      console.log("Get started with Puck AI by checking out the docs:");
+      console.log(
+        `  - Onboarding: ${ansiColors.cyan}https://cloud.puckeditor.com/onboarding/integrate${ansiColors.reset}`
+      );
+      console.log(
+        `  - Getting started: ${ansiColors.cyan}https://puckeditor.com/docs/ai/getting-started${ansiColors.reset}\n`
+      );
+    }
```
chrisvxd, line 235: Two links is too many. I think we should probably just point to the onboarding.

I'm unclear what the ideal CTA and destination is since we're still finalising the onboarding, but I think something functional is fine:

> Configure Puck AI access: https://cloud.puckeditor.com/onboarding/integrate

recipes/next-ai/app/puck/[...puckPath]/client.tsx
```diff
@@ -0,0 +1,25 @@
+"use client";
```
Comment (chrisvxd) [inline recipes/next-ai/app/puck/[...puckPath]/client.tsx:1]: I wonder if we get rid of this catch-app route.

recipes/next-ai/app/api/pages/route.ts
```diff
@@ -0,0 +1,25 @@
+import { revalidatePath } from "next/cache";
```
chrisvxd, line 1: I know this is just consistent with the existing Next recipe, but having both `/api/puck` and `/puck/api` is very confusing. I think we could either move this to `/api/puck/page/route.ts`, or get rid of it entirely.

recipes/react-router-ai/.env.example
```diff
@@ -0,0 +1,2 @@
+# Add your key here, more information available at: https://puckeditor.com/docs/ai/getting-started#generate-an-api-key
+PUCK_API_KEY=
```
chrisvxd, line 2: I wasn't necessarily expecting us to duplicate the entire application for each recipe here, but I suppose it's quite clean and canonical

recipes/next-ai/app/api/pages/route.ts
FedericoBonel: Yeah! I agree, we should also do it in the og recipe I think, but will do on this one first!

recipes/react-router-ai/.env.example
FedericoBonel: I did have a think about this, I initially thought of either:

- Surgically insert the code and files required in the original recipe.
- Have the original recipe as the base and the new ones define only the files that need to be replaced

However other tools (vite, next, etc.), actually redefine the whole recipe for integrations like this, and I think it does make sense since it probably can change and become more complicated as we grow (could be we need completely different routing, setups with third party packages, or something else)

So I decided to go for the simplest solution, but happy to change it if you think the other two or another one would be best.

recipes/next-ai/app/puck/[...puckPath]/client.tsx
Comment (FedericoBonel) [reply recipes/next-ai/app/puck/[...puckPath]/client.tsx:1]: Happy to do it. 

We would lose the multi page generation (which could be super fine honestly, we could just create some docs for this approach if we wanted to), or we could provide some selector for that, since this would now be only rendered at /puck.

packages/create-puck-app/index.js
FedericoBonel: Done added! Let me know what you think, I tried keeping it minimal and just clarify what the result of enabling it would be.

packages/create-puck-app/index.js
FedericoBonel: Absolutely, I agree, added your suggestion, we can always fix the link if we decide to direct somewhere else.

recipes/next-ai/app/api/pages/route.ts
FedericoBonel: Done!

FedericoBonel:
@chrisvxd, applied your suggestions, I also moved the ai packages to point to the latest canary to avoid the dependencies to `@measured/puck`. 

Only remaining question is if we want the catch all route or not, happy to proceed either way.

recipes/next-ai/app/api/puck/[...all]/route.ts
```diff
@@ -0,0 +1,11 @@
+import { NextRequest } from "next/server";
+import { puckHandler } from "@puckeditor/cloud-client";
+
+export const POST = (request: NextRequest) => {
```
Comment (chrisvxd) [inline recipes/next-ai/app/api/puck/[...all]/route.ts]: Add a link to the AI docs here

chrisvxd (approved):

PR #1495, opened 2026-01-12
refactor: encode empty objects as strings to prevent stripping them

### LINKED ISSUE #1494: Empty array in nested array field default props gets assigned to undefined
## Description

When using two nested array fields, if the second one is assigned an empty array in the top-level default props, it gets set to `undefined` in the render function.

## Environment

- Puck version: canary

## Steps to reproduce

1. Add a component that uses two nested array fields and sets an empty array for the nested one at the default props level:

```tsx
const config = {
  components: {
    ArrayTest: {
      fields: {
        items: {
          type: "array",
          arrayFields: {
            nested: { type: "array", arrayFields: { text: { type: "text" } } },
          },
          defaultItemProps: {
            nested: [],
          },
        },
      },
      defaultProps: {
        items: [
          {
            text: "Item 1",
            nested: [], // This gets set to undefined
          },
        ],
      },
      render: ({ items }) =>
        items.map((item, index) => (
          <div key={index}>
            {/* This throws an error because nested is undefined */}
            <p>{item.nested.map((nestedItem) => nestedItem.text).join(", ")}</p>
          </div>
        )),
    },
  },
};
```

2. Run the application in development mode.  
3. Navigate to the editor.  
4. Add the component to the page.

## What happens

The nested array field (in the above example, the `nested` field) is set to `undefined`.

## What I expect to happen

The nested array field should be set to whatever is defined in `defaultProps`.

chrisvxd:
Closes #1494 

## Description

a4bfae4f202ddaab12294cfffd2664d809e7cb54 stripped empty objects to prevent infinite renders in certain scenarios, but accidentally removed empty objects/arrays from expected scenarios.

Instead of stripping them, this approach encodes/decodes the empty objects as strings, solving the original re-rendering problem without losing the value.

## How to test

- Follow steps from #1494
- Also ensure no regressions by following test steps in https://github.com/puckeditor/puck/pull/1479

### DIFF
--- a/packages/core/lib/data/flatten-node.ts
+++ b/packages/core/lib/data/flatten-node.ts
@@ -19,7 +19,7 @@ function encodeEmptyObjects(props: Record<string, any> = {}) {
 
     const val = props[key];
 
-    if (Array.isArray(val)) {
+    if (Array.isArray(val) && val.length === 0) {
       result[key] = emptyArrayStr;
     } else if (isPureObject(val) && Object.keys(val).length === 0) {
       result[key] = emptyObjectStr;


packages/core/lib/data/flatten-node.ts
```diff
@@ -5,31 +5,47 @@ import { stripSlots } from "./strip-slots";
...
+    if (!Object.prototype.hasOwnProperty.call(props, key)) continue;
 
-  if (
-    val != null &&
-    Object.prototype.toString.call(val) === "[object Object]"
-  ) {
-    return Object.keys(val).length === 0;
+    const val = props[key];
+
+    if (Array.isArray(val)) {
```
FedericoBonel: Shouldn't we check if the array is empty here?

```suggestion
    if (Array.isArray(val) && val.length === 0) {
```

FedericoBonel:
Awesome! Thanks for the fix, this still doesn't remount and now the nested arrays are not set to undefined. 

I left a comment because I thought it was needed, but when checking the execution flow it seems it never receives an empty array (empty slots get set to null, and arrays are flatten down when they have value). However we might still add the check just to be sure we don't empty an array.

packages/core/lib/data/flatten-node.ts
chrisvxd: Yes good catch!

PR #1514, opened 2026-01-19
fix(iframe): use filtered.length for style loading completion check

### LINKED ISSUE #1324: Iframe not receiving pointer events due to onStylesLoaded
## Description
We are using the compositional patterns to customise the editor to our needs and are using plugins for custom fields. I've run into a bug where the slots in components are not enabled. 

While debugging i came towards the following conclusion:

[useBubbleIFrameEvents](https://github.com/puckeditor/puck/blob/cf074bc6cb52e66e08bf6e36199ad9e89b0f19fb/packages/core/components/Puck/components/Preview/index.tsx#L21) is not receiving the status 'READY' because the [CopyHostStyles](https://github.com/puckeditor/puck/blob/cf074bc6cb52e66e08bf6e36199ad9e89b0f19fb/packages/core/components/AutoFrame/index.tsx#L269-L285) thinks not all styles are loaded yet.

This is a specific scenario happends because one of the plugins we created is a code editor field loading the monaco-editor which is adding styles while loading the field, increasing the elements list while the [Promise](https://github.com/puckeditor/puck/blob/cf074bc6cb52e66e08bf6e36199ad9e89b0f19fb/packages/core/components/AutoFrame/index.tsx#L243-L285) is resolving as a result the stylesLoaded might be equal to the mirrorStyles but in reality the elements list increased therefore onStylesLoaded is never called and the status is never set to ready, leaving the iframe without receiving events.

I think the stylesLoaded count should be validated against the filtered list instead of the elements list.

```
 if (stylesLoaded >= filtered.length) {
            onStylesLoaded();
 }

## Environment

- Puck version: [0.20.2]
- Browser version: [Chrome 140 (desktop)]
- Additional environment info: [MacOs]

<!--
  Detail the environment where the bug is occurring.
-->

## Steps to reproduce

1. Create a compositional puck editor
2. Build a plugin that dynamically loads css 
3. use the plugin in the root fields
4. use a component that has slot fields
5. try sorting the elements in the slot results in the slot not being sortable.

## What I expect to happen

I would expect the onStylesLoaded to be called so that the iframe receives the pointer events enabling the dropzones when hovering, so i can sort the components within a slot.

## Additional Media

bug:
https://github.com/user-attachments/assets/9023007e-d85b-4f77-b523-f831ed448822

expected:
https://github.com/user-attachments/assets/0b554fa4-18fd-4295-a180-f0524013e024

### PR DESCRIPTION
**Pull request by allandelmare**
## Summary

Fix style loading validation to compare against `filtered.length` instead of `elements.length`.

## Problem

When stylesheets are dynamically added to the DOM (e.g., Monaco editor CSS), the `elements` array grows but `filtered` remains the same size. This causes the completion check `stylesLoaded >= elements.length` to never trigger, leaving the iframe in a loading state.

## Solution

Change the comparison from `elements.length` to `filtered.length` since `filtered` represents the actual stylesheets that need to load.

## Changes

  - `packages/core/components/AutoFrame/index.tsx`: 2 lines changed

Closes #1324

### DIFF
--- a/packages/core/components/AutoFrame/index.tsx
+++ b/packages/core/components/AutoFrame/index.tsx
@@ -290,6 +290,17 @@ const CopyHostStyles = ({
       // Inject initial values in bulk
       doc.head.append(...filtered);
 
+      // Count <style> elements as immediately loaded (they don't fire onload)
+      filtered.forEach((mirror) => {
+        if (mirror.nodeName === "STYLE") {
+          stylesLoaded = stylesLoaded + 1;
+        }
+      });
+
+      if (stylesLoaded >= filtered.length) {
+        onStylesLoaded();
+      }
+
       observer.observe(parentDocument.head, { childList: true, subtree: true });
 
       filtered.forEach((el) => {

### CODE REVIEW

**Comment by kernel-io** [comment]
This patch requires more changes to fully solve the issue

After appending styles to doc.head - count how many <style> elements are loaded as they don't fire onload events, after this check if stylesLoaded >= filtered.length - this also handles an edge case where filtered.length == 0 

i.e on line 292.

```javascript
doc.head.append(...filtered)

// Count <style> elements as immediately loaded (they don't fire onload)
filtered.forEach((mirror) => {
  if (mirror.nodeName === "STYLE") {
    stylesLoaded = stylesLoaded + 1;
  }
});

if (stylesLoaded >= filtered.length) {
  onStylesLoaded();
}
```

allandelmare:
Good catch, thanks! Pushed a fix in 7f300f5c.

PR #1517, opened 2026-01-19
fix(richtext): prevent focus race condition dropping dropdown changes

### LINKED ISSUE #1516: bug(richtext): dropdown control changes not triggering dirty state due to focus race condition
### Description

When using Tiptap-based rich text fields, changes made via dropdown controls (font size, color picker, etc.) are not recognized by Puck's change detection. The editor only registers these changes after a subsequent content edit.

### Reproduction Steps

1. Add a component with a `richtext` field
2. Select some text in the rich text editor
3. Use a dropdown control (e.g., heading selector, font size) to change the text style
4. Observe: No "unsaved changes" indicator appears
5. Type any character
6. Observe: Now the changes are recognized

### Expected Behavior

Style changes via dropdown controls should immediately trigger dirty state, just like bold/italic/underline do.

### Actual Behavior

- Dropdown style changes: Not recognized until next edit
- Bold/italic/underline: Immediately recognized
- Typing: Immediately recognized

The style IS applied (HTML is correct), but Puck doesn't recognize the document as dirty.

### Root Cause

In `use-synced-editor.ts`, the `onUpdate` callback has a guard condition:

```typescript
onUpdate: ({ editor }) => {
  if (syncingRef.current || !isFocused) {
    appStoreApi.getState().setUi({ field: { focus: name } });
    return;  // Changes dropped!
  }
  // ... rest of update logic
}
```

When a dropdown control is used:
1. Focus moves from editor to dropdown
2. User selects option → command runs with `.focus()`
3. Tiptap's `onUpdate` fires synchronously
4. But React's `isFocused` state hasn't updated yet (batched)
5. Guard condition `!isFocused` is true → early return
6. Change is silently dropped

### Impact

- "Dirty" indicator doesn't show
- Undo history may not update
- Auto-save doesn't trigger
- Users may lose changes

### Affected Controls

Any control that moves focus away from the editor:
- Heading/paragraph selector
- Font size selector
- Color pickers
- Link dialogs
- Custom dropdown extensions

### Environment

- `@puckeditor/core`: 0.21.0
- Browser: All (race condition is framework-level)

### Suggested Fix

Remove the `isFocused` guard since `syncingRef.current` already prevents processing during sync operations:

```typescript
onUpdate: ({ editor }) => {
  if (syncingRef.current) {
    return;  // Only skip during sync
  }

  const html = editor.getHTML();
  // ... process update
}
```

Or use `queueMicrotask` to allow React state to settle before checking focus.

---

I have a fix ready and can submit a PR if this approach is acceptable.

allandelmare:
## Summary

Remove the `isFocused` guard from the `onUpdate` callback to fix a race condition where dropdown control changes were silently dropped.

## Problem

When using dropdown controls (font size selector, color picker, heading selector), changes are not recognized by Puck's dirty state tracking:

1. User clicks dropdown → focus leaves editor
2. User selects option → command runs with `.focus()`
3. Tiptap's `onUpdate` fires synchronously
4. React's `isFocused` state is still stale (batched update)
5. Guard condition `!isFocused` is true → early return
6. Change is dropped, dirty indicator doesn't show

## Solution

Remove `!isFocused` from the guard condition. The `syncingRef.current` check is sufficient to prevent processing during undo/redo history loads.

```diff
- if (syncingRef.current || !isFocused) {
-   appStoreApi.getState().setUi({ field: { focus: name } });
-   return;
- }
+ if (syncingRef.current) {
+   return;
+ }
+ appStoreApi.getState().setUi({ field: { focus: name } });
```

## Test plan

- [ ] Apply bold/italic via keyboard shortcut → dirty state triggers
- [ ] Apply bold/italic via toolbar button → dirty state triggers
- [ ] Change heading via dropdown → dirty state triggers (was broken)
- [ ] Change font size via dropdown → dirty state triggers (was broken)
- [ ] Undo/redo still works correctly
- [ ] External content sync still works correctly

Closes #1516

### DIFF
--- a/packages/core/components/RichTextEditor/lib/use-synced-editor.ts
+++ b/packages/core/components/RichTextEditor/lib/use-synced-editor.ts
@@ -56,12 +56,13 @@ export function useSyncedEditor({
     immediatelyRender: false,
     parseOptions: { preserveWhitespace: "full" },
     onUpdate: ({ editor }) => {
-      // This can trigger during undo/redo history loads
-      if (syncingRef.current || !isFocused) {
-        appStoreApi.getState().setUi({ field: { focus: name } });
-
+      // Skip during sync operations (undo/redo history loads)
+      if (syncingRef.current) {
         return;
       }
+
+      // Ensure focus state is tracked
+      appStoreApi.getState().setUi({ field: { focus: name } });
 
       const html = editor.getHTML();
 


FedericoBonel:
Hey @allandelmare! Since the original issue got closed, I'll close this PR

PR #1523, opened 2026-01-20
feat(iframe): add custom stylesheet support

### LINKED ISSUE #1515: feat(iframe): support custom stylesheet injection
### Problem

When embedding custom content in the Puck iframe (e.g., Tailwind CSS from an API endpoint, custom fonts), there's no way to inject stylesheets that participate in the `waitForStyles` mechanism.

Currently, the only workaround is using the `iframe` override to inject styles via `useEffect`, but this causes a flash of unstyled content (FOUC) because:
1. Custom stylesheets aren't in the host document
2. They're injected after the override mounts
3. `waitForStyles` completes before custom styles load

### Proposed Solution

Add two new properties to `IframeConfig`:

```typescript
interface IframeConfig {
  enabled?: boolean;
  waitForStyles?: boolean;
  // New:
  stylesheets?: string[];
  onStylesLoaded?: () => void;
}
```

**Usage:**

```tsx
<Puck
  iframe={{
    enabled: true,
    waitForStyles: true,
    stylesheets: [
      '/api/tailwind-styles',
      'https://fonts.googleapis.com/css2?family=Inter'
    ],
    onStylesLoaded: () => console.log('All styles ready'),
  }}
/>
```

### Use Case

I'm building a PayloadCMS plugin that integrates Puck. Users can configure custom stylesheets (Tailwind, custom fonts, design system CSS) that need to load in the preview iframe before rendering.

### Implementation

I have a working implementation ready if this approach is approved. It:
- Injects custom `<link>` elements after host styles
- Tracks loading via `onload`/`onerror` handlers
- Includes a timeout fallback for reliability
- Calls `onStylesLoaded` when all styles (host + custom) are ready

Happy to submit a PR if this direction makes sense!

allandelmare:
## Summary

Add support for injecting custom stylesheets into the iframe that participate in the `waitForStyles` mechanism.

## Changes

- Add `stylesheets` array to `IframeConfig` for custom CSS URLs
- Add `onStylesLoaded` callback to `IframeConfig`
- Inject custom `<link>` elements after host styles
- Track loading via `onload`/`onerror` handlers
- Add 10s timeout fallback for reliability
- Properly clean up timeout when styles load successfully

## Usage

```typescript
<Puck
  iframe={{
    enabled: true,
    waitForStyles: true,
    stylesheets: ['/api/styles', 'https://fonts.googleapis.com/...'],
    onStylesLoaded: () => console.log('Ready!'),
  }}
/>
```

## Test plan

- [ ] Verify custom stylesheets are injected into iframe `<head>`
- [ ] Verify `onStylesLoaded` callback fires after all styles load
- [ ] Verify timeout fallback works when a stylesheet fails to load
- [ ] Verify existing behavior unchanged when `stylesheets` not provided

Closes #1515

### DIFF
--- a/packages/core/components/AutoFrame/index.tsx
+++ b/packages/core/components/AutoFrame/index.tsx
@@ -66,14 +66,18 @@ const syncAttributes = (sourceElement: Element, targetElement: Element) => {
 
 const defer = (fn: () => void) => setTimeout(fn, 0);
 
+const STYLESHEET_LOAD_TIMEOUT = 10000; // 10 seconds
+
 const CopyHostStyles = ({
   children,
   debug = false,
   onStylesLoaded = () => null,
+  customStylesheets = [],
 }: {
   children: ReactNode;
   debug?: boolean;
   onStylesLoaded?: () => void;
+  customStylesheets?: string[];
 }) => {
   const { document: doc, window: win } = useFrame();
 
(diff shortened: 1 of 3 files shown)


chrisvxd:
Sorry for the delay, but I'm closing this as it's already possible via the [iframe override](https://puckeditor.com/docs/api-reference/overrides/iframe)

allandelmare:
Thanks for the look. Quick clarification — I don't think the override covers what this PR was after.

The override gives `document`, which lets us inject `<link>` tags in a `useEffect`. We already do this — it's how our IframeWrapper works today. The override is the substrate this PR builds on, not a substitute.

The gap is timing: injection in `useEffect` runs *after* Puck has declared the iframe ready and rendered children, which produces FOUC on every editor open. Working around that from userland requires manual relative→absolute URL resolution for srcdoc, `onload`/`onerror` tracking with a timeout fallback, and forced-repaint hacks (visibility toggle + reflow) to get already-rendered elements to pick up late-arriving styles. None of that is achievable from the override alone — it needs integration with the existing `waitForStyles` path so user stylesheets are awaited before render.

Happy to scope this down — just `iframe.stylesheets: string[]`, dropping the `onStylesLoaded` callback for a future PR if the surface area is the concern. If that's still not a direction you want to take Puck, totally fair — just wanted to flag that the closing rationale didn't quite match what was being proposed.

PR #1532, opened 2026-01-22
fix: prevent block deletion on backspace key press in rich text

### LINKED ISSUE #1529: Hitting the backspace key while inline editing a richtext field will delete component
## Description

When the user is editing the content of a richtext field with `contentEditable: true`, hitting the backspace key will delete the component from the layout.

Pucks custom keybindings interfer with native keyboard behaviour.

## Environment

- Puck version: 0.21.0
- Browser version: Firefox 146 Dekstop
- Additional environment info: Webpack, MacOS, MacBook Pro

## Steps to reproduce

1. Create an component that uses a richtext field with `contentEditable: true`
2. Add the component to the layout canvas
3. Enter any text using the inline functionality
4. Hit backspace key to delete text
5. See the component disappear from the layout canvas

The same will happen, when editing an already existing richtext component.

## What happens

Component will be deleted from the canvas.

## What I expect to happen

Text should be deleted.

## Additional Media

https://github.com/user-attachments/assets/1626111c-c520-497c-925b-e23704e329b3

DamianKocjan:
Closes #1529

### DIFF
--- a/packages/core/components/RichTextEditor/components/EditorInner.tsx
+++ b/packages/core/components/RichTextEditor/components/EditorInner.tsx
@@ -49,6 +49,11 @@ export const EditorInner = memo(
           event.preventDefault();
           editor?.commands.toggleItalic?.();
         }
+
+        // Prevent event propagation for backspace. When propagated, it will trigger block deletion
+        if (event.key.toLowerCase() === "backspace") {
+          event.stopPropagation();
+        }
       },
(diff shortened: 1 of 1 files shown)


FedericoBonel:
Hey @DamianKocjan!

Thanks for the contribution.
This works and is the same pattern we have for the inline text fields, so it looks good to me.

🫡

PR #1533, opened 2026-01-22
fix: prevent unintended history entry

### LINKED ISSUE #1521: Unintended history entry
## Description

When clicking outside of the DropZone, a new history record is added even if the state is [equals](https://ramdajs.com/docs/#equals) to the state from the previous history entry. 

## Environment

- Puck version: [0.20.2]
- Browser version: [Chrome 144 (desktop)]

## Steps to reproduce

This could be reproduced on the [demo page](https://demo.puckeditor.com/edit). On the fresh page (empty history), click outside of the DropZone (pink area on the screenshot).
<kbd>Image</kbd>

## What happens

The Undo button is active, and a new history record is added.

## What I expect to happen

The Undo button should remain disabled.

## Additional Media

[](https://github.com/user-attachments/assets/09c7b5fe-1ad0-408c-b19a-a096db54aadc)

DamianKocjan:
Closes #1521


packages/core/components/Puck/components/Canvas/index.tsx
```diff
@@ -234,10 +234,12 @@ export const Canvas = () => {
       })}
       onClick={(e) => {
         const el = e.target as Element;
+        const hasItemSelected = !!appStoreApi.getState().state.ui.itemSelector;
 
         if (
           !el.hasAttribute("data-puck-component") &&
-          !el.hasAttribute("data-puck-dropzone")
+          !el.hasAttribute("data-puck-dropzone") &&
+          hasItemSelected
```
FedericoBonel: I _believe_ this is not quite right, this fixes the problem, but it still creates an entry when I have something selected and click on the grey space around the canvas.

I think the actual fix might be setting `recordHistory` to false underneath. But haven't checked through if that has any other side effects yet.

DamianKocjan: Tested `recordHistory: false`, but it leads to weird history behavior. For example:
1. We select block.
2. We deselect it, by clicking empty canvas space.
3. We click history back button.
This action will do nothing from ui perspective. So it will be confusing from user perspective.

The current implementation avoids this side effect and ensures the UI state matches the history entry.

FedericoBonel: Fair, I think this gets the job done of not creating entries when nothing was initially selected. Thanks for the responsiveness! 🫡

FedericoBonel (approved):

packages/core/components/Puck/components/Canvas/index.tsx
chrisvxd: Chiming in here. This fix helps prevent duplicate entries once an item is deselected, but it is still causing an unexpected history item when the user deselects an item.

UI state changes alone (like in the steps you provided in your last comment @DamianKocjan) should _not_ cause history entries. Normally, setting`recordHistory: false` would force this (this is also the default for `setUi`).

However, it sounds like you're seeing an entry being added, so something else must be adding it.

I think this merits a bit further investigation. Perhaps you could try combining both solutions, or exploring further.

packages/core/components/RichTextEditor/lib/use-synced-editor.ts
```diff
@@ -91,6 +92,13 @@ export function useSyncedEditor({
   useEffect(() => {
     if (debouncedState) {
       const { ui } = appStoreApi.getState().state;
+      const currentData = appStoreApi.getState().getCurrentData();
+      const currentFieldValue = getDeep(currentData.props, name ?? "");
+
+      // Avoid echoing back same content
+      if (currentFieldValue === debouncedState.html) {
+        return;
+      }
```
chrisvxd: This looks like its leaked into this PR. If this is intentional, can you capture as a separate PR with relevant issue?

DamianKocjan: It was intentional. When I was testing against unintentional history entries I found that issue, so I fixed it. Thought it'd be ideal to merge it here with rest of related changes.

Commit addressing this issue with explanation: [fc58153](https://github.com/puckeditor/puck/pull/1533/commits/fc581535518fa3d5f0ba8de16e927622be56874c)

I'll create issue and PR later that day. Or we can add one more issue that this PR is addressing.

chrisvxd: Wonder if this is fixed in #1587

DamianKocjan: I've been thinking about it too. So I'll probably undo this commit from this pr and see if issue still persist after fix in #1587

DamianKocjan:
@chrisvxd You can merge. Tested once again to be sure

PR #1534, opened 2026-01-22
fix: disabling list item for richtext throws

### LINKED ISSUE #1468: Disabling `listItem` in richtext field throws an error
## Description

Setting the `listItem` option to `false` in the `richtext` field throws an error.

## Environment

- Puck version: 0.21.0-canary.fee2ba98

## Steps to reproduce

1. Use a config with a `richtext` field and set `listItem: false`.

```tsx
  components: {
    HeadingBlock: {
      fields: {
        title: {
          type: "richtext",
          options: {
            listItem: false,
          },
        },
      },
      render: ({ title }) => (
        <div style={{ padding: 64 }}>
          <h1>{title}</h1>
        </div>
      ),
    },
  },
};
```

2. Navigate to the editor.
3. Add a `HeadingBlock` component to the page.

## What happens

An error is thrown:

```zsh
No node type or group 'listItem' found (in content expression 'listItem+')
```

## What I expect to happen

The list item option should be disabled and no errors should be thrown.

DamianKocjan:
Closes #1468

### DIFF
--- a/packages/core/components/RichTextEditor/extension.ts
+++ b/packages/core/components/RichTextEditor/extension.ts
@@ -179,7 +179,9 @@ export const PuckRichText = Extension.create<PuckRichTextOptions>({
       extensions.push(Blockquote.configure(options.blockquote));
     }
 
-    if (options.bulletList !== false) {
+    // When we want bullet lists, we also need list items
+    // https://tiptap.dev/docs/editor/extensions/nodes/bullet-list#install
+    if (options.bulletList !== false && options.listItem !== false) {
       extensions.push(BulletList.configure(options.bulletList));
     }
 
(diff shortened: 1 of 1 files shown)


FedericoBonel:
Hey @DamianKocjan!

Thanks for the contribution!

packages/core/components/RichTextEditor/extension.ts
```diff
@@ -179,7 +179,9 @@ export const PuckRichText = Extension.create<PuckRichTextOptions>({
       extensions.push(Blockquote.configure(options.blockquote));
     }
 
-    if (options.bulletList !== false) {
+    // When we want bullet lists, we also need list items
+    // https://tiptap.dev/docs/editor/extensions/nodes/bullet-list#install
+    if (options.bulletList !== false && options.listItem !== false) {
       extensions.push(BulletList.configure(options.bulletList));
     }
```
FedericoBonel, line 187: I think it would be good if we could check once for options.listItem and then nest the other two checks inside, that way this dependency doesn't get lost when reviewing and if we ever need to add another one we know where to put it to not break anything.

```tsx
if (options.listItem !== false) {
  if (options.bulletList !== false) { ... }
  if (options.orderedList !== false) { ... }
}
```

chrisvxd: Let's merge for now so we can release a patch

PR #1537, opened 2026-01-23
Refactor: improve extension setup for bullet and ordered lists

DamianKocjan:


DamianKocjan:
@FedericoBonel here are improvements suggested in #1534

PR #1541, opened 2026-01-23
fix: allow typing polish `ż` letter

### LINKED ISSUE #1277: Cannot type polish letter "ż" in text fields
## Description

When trying to type polish letter `ż` (`Alt Gr + Z`) in text field, it undoes the last action instead (like `Ctrl + Z` hotkey)

## Environment

- Puck version: 0.19.0, 0.20.1
- Browser version: Edge 139 (desktop), Firefox Developer 143 (desktop)
- Additional environment info: Windows with keyboard `Polish (Programmers)`

## Steps to reproduce

1. Set Windows keyboard to `Polish (Programmers)`
2. Create any component with field of type `text`
3. Add the component to preview
4. Type letter `ż` (`Alt Gr + Z` or `Right Alt + Z`)

## What happens

The undo action is triggered and component disappear

## What I expect to happen

Letter `ż` appears in text field

## Additional Media

Logs after adding `document.addEventListener('keydown', console.log)` and pressing just `Alt Gr`:
```
KeyboardEvent {isTrusted: true, key: 'Control', code: 'ControlLeft', location: 1, ctrlKey: true, …}
isTrusted: true
altKey: false
bubbles: true
cancelBubble: false
cancelable: true
charCode: 0
code: "ControlLeft"
composed: true
ctrlKey: true
currentTarget: null
defaultPrevented: false
detail: 0
eventPhase: 0
isComposing: false
key: "Control"
keyCode: 17
location: 1
metaKey: false
repeat: false
returnValue: true
shiftKey: false
sourceCapabilities: InputDeviceCapabilities {firesTouchEvents: false}
srcElement: body
target: body
timeStamp: 16245
type: "keydown"
view: Window {0: Window, window: Window, self: Window, document: document, name: '', location: Location, …}
which: 17
[[Prototype]]: KeyboardEvent
```
```
KeyboardEvent {isTrusted: true, key: 'AltGraph', code: 'AltRight', location: 2, ctrlKey: false, …}
isTrusted: true
altKey: false
bubbles: true
cancelBubble: false
cancelable: true
charCode: 0
code: "AltRight"
composed: true
ctrlKey: false
currentTarget: null
defaultPrevented: false
detail: 0
eventPhase: 0
isComposing: false
key: "AltGraph"
keyCode: 18
location: 2
metaKey: false
repeat: false
returnValue: true
shiftKey: false
sourceCapabilities: InputDeviceCapabilities {firesTouchEvents: false}
srcElement: body
target: body
timeStamp: 16245.5
type: "keydown"
view: Window {0: Window, window: Window, self: Window, document: document, name: '', location: Location, …}
which: 18
[[Prototype]]: KeyboardEvent
```

DamianKocjan:
Closes #1277

### DIFF
--- a/packages/core/lib/use-hotkey.ts
+++ b/packages/core/lib/use-hotkey.ts
@@ -34,6 +34,7 @@ const keys = [
   "z",
   "delete",
   "backspace",
+  "altRight",
 ] as const;
 
 type KeyStrict = (typeof keys)[number];
@@ -76,6 +77,7 @@ const keyCodeMap: KeyCodeMap = {
   KeyZ: "z",
   Delete: "delete",
   Backspace: "backspace",
+  AltRight: "altRight",
 };
 
 const useHotkeyStore = create<{
@@ -98,6 +100,12 @@ const useHotkeyStore = create<{
 
 export const monitorHotkeys = (doc: Document) => {
   const onKeyDown = (e: KeyboardEvent) => {
+    // If altGraphKey (Alt Right) is pressed, register altRight instead of mapping ControlRight to ctrl
+    if (e.getModifierState("AltGraph")) {
+      useHotkeyStore.getState().hold("altRight");
+      return;
+    }
+
     const key = keyCodeMap[e.code];
 
     if (key) {
@@ -115,8 +123,7 @@ export const monitorHotkeys = (doc: Document) => {
           );
 
         // Call hotkey with event; skip preventDefault if callback returns false to allow native input behavior.
-        // Also skip if AltGraph is pressed, as it's used for typing special characters
-        if (conditionMet && !e.getModifierState("AltGraph")) {
+        if (conditionMet) {
           const handled = cb(e);
           if (handled !== false) {
             e.preventDefault();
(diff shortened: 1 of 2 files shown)


packages/core/lib/use-hotkey.ts
```diff
@@ -115,7 +115,8 @@ export const monitorHotkeys = (doc: Document) => {
           );
 
         // Call hotkey with event; skip preventDefault if callback returns false to allow native input behavior.
-        if (conditionMet) {
+        // Also skip if AltGraph is pressed, as it's used for typing special characters
+        if (conditionMet && !e.getModifierState("AltGraph")) {
```
FedericoBonel: I'd actually prefer it if checked for this on the place where we add the subscription to the `meta + z`/ `ctrl + z` hotkey combination [here](https://github.com/puckeditor/puck/blob/31c70f1c551bf99d12f7504eed8f8733bc59a96d/packages/core/store/slices/history.ts#L184C3-L184C44), maybe with something like:

```tsx
  useHotkey({ meta: true, z: true, altRight: false }, back);

  useHotkey({ ctrl: true, z: true, altRight: false }, back);
```

That way we avoid blocking using the Alt graph for any future hotkey combinations we might want to use, and if we run into any other issues like this we can always stack them there, instead of hardcoding them here as additional checks (which might break other behavior as we evolve it), keep in mind you might have to add it to the key map.

PR #1547, opened 2026-01-27
feat: add iframe.srcDoc param

chrisvxd:
Enable a custom srcdoc to be provided to the Puck iframe


FedericoBonel (changes requested):
This looks fine, but I think we should use the `appStore` with a selector. The props context would re-render the iframe whenever any of the props change.

PR #1553, opened 2026-01-29
fix: don't override ui.viewport.current on editor render

### LINKED ISSUE #1544: ui.viewports.current is ignored when custom viewports are provided
## Description

The `<Puck />` component does not respect the initial viewport defined via `ui.viewports.current` when custom `viewports` are provided.

Even when `ui.viewports.current` exactly matches one of the configured viewport options (same `width` and `height`), Puck does not start with that viewport and instead selects a different one based on window size or internal logic.

This makes it impossible to reliably control the initial viewport using the documented `ui` API.

## Environment

Puck version: 0.21<br>Browser version: Chrome 144.0.7559.97 <br>Additional environment info: React 19.2.1,  Next.js 16.0.8

## Steps to reproduce

1. Render with custom viewport options and an explicit initial viewport:

```
const Editor = () => {
  const viewportOptions = [
        { width: 300, height: 300, label: "S" },
        { width: 600, height: 600, label: "M" },
        { width: 900, height: 900, label: "L" },
  ]

  return (
    <Puck
      config={config}
      data={data}
      viewports={viewportOptions}
      ui={{
        viewports: {
          controlsVisible: true,
          current: { width: 300, height: 300 },
          options: viewportOptions
        },
      }}
    />
  );
};
```

2. Start the application and load the editor.

## What happens

* The editor does not start with the `300×300` viewport.
* A different viewport is selected (apparently based on window size).
* This happens even though:<br>-- `ui.viewports.current` exactly matches a viewport option<br>-- no compositional UI (`<Puck.Preview />`) is used <br>-- no iframe or viewport features are disabled

## What I expect to happen

* Puck should start with the viewport defined in `ui.viewports.current`

FedericoBonel:
Closes #1544

## Description

This PR fixes a bug where the default selected viewport provided by the user in `ui.viewports.current` was being overridden by changes introduced in commit 435c32f2071a6a677a4184440a922c2378c1daa2.

## Changes Made

- We now check if the user provided a default viewport ui before selecting the most appropiate one based on screen size.

## How to Test

1. Render the `Puck` component with a default viewport set to the smallest size:

```tsx
const Editor = () => {
  const viewportOptions = [
    { width: 300, height: 300, label: "S" },
    { width: 600, height: 600, label: "M" },
    { width: 900, height: 900, label: "L" },
  ];

  return (
    <Puck
      config={config}
      data={data}
      viewports={viewportOptions}
      ui={{
        viewports: {
          controlsVisible: true,
          current: { width: 300, height: 300 },
          options: viewportOptions,
        },
      }}
    />
  );
};
```

2. Navigate to the editor.
3. Confirm that the specified `current` viewport is selected as expected.

### DIFF
--- a/packages/core/components/Puck/components/Canvas/index.tsx
+++ b/packages/core/components/Puck/components/Canvas/index.tsx
@@ -31,6 +31,7 @@ export const Canvas = () => {
   const {
     _experimentalFullScreenCanvas,
     viewports: viewportOptions = defaultViewports,
+    ui: uiProp,
   } = usePropsContext();
 
   const {
@@ -156,8 +157,12 @@ export const Canvas = () => {
 
   const appStoreApi = useAppStoreApi();
 
+  // Select closest viewport on load
   useEffect(() => {
     if (typeof window === "undefined") return;
+
+    // Don't override if user has set a viewport
+    if (uiProp?.viewports?.current) return;
 
     const viewportWidth = window.innerWidth;
     const frameWidth = frameRef.current?.getBoundingClientRect().width;
@@ -223,7 +228,13 @@ export const Canvas = () => {
 
       appStoreApi.setState({ ...appState, history });
     }
-  }, [viewportOptions, frameRef.current, iframe, appStoreApi]);
+  }, [
+    viewportOptions,
+    frameRef.current,
+    iframe,
+    appStoreApi,
+    uiProp?.viewports?.current,
+  ]);
 
   return (
     <div


packages/core/components/Puck/components/Canvas/index.tsx
```diff
@@ -223,7 +228,7 @@ export const Canvas = () => {
 
       appStoreApi.setState({ ...appState, history });
     }
-  }, [viewportOptions, frameRef.current, iframe, appStoreApi]);
+  }, [viewportOptions, frameRef.current, iframe, appStoreApi, uiProp]);
```
chrisvxd: Better to be more specific here

```suggestion
  }, [viewportOptions, frameRef.current, iframe, appStoreApi, uiProp?.viewports]);
```

FedericoBonel: Yes! Good catch.

FedericoBonel: Changed it! Added a dependency to `uiProp?.viewports?.current` a bit more specific

PR #1554, opened 2026-01-29
fix(core): prevent slot array corruption during dnd

jderboven:
Description:                                                                                                             
  ## Problem                                                                                                               
  Intermittent `TypeError: arr.some is not a function` errors when dragging components with slots.                         
                                                                                                                           
  ## Root Cause                                                                                                            
  Two related bugs:                                                                                                        
  1. **defaultSlots**: Wrong spread order `{ [fieldName]: [], ...acc }` was overwriting existing slot arrays               
  2. **walkField**: `value || []` doesn't protect against non-array truthy values                                          
                                                                                                                           
  ## Solution                                                                                                              
  1. Fix spread order and add `Array.isArray()` check in `defaultSlots`                                                    
  2. Use `Array.isArray(value) ? value : []` in `walkField`                                                                
                                                                                                                           
  ## Testing                                                                                                               
  - Existing tests pass                                                                                                    
  - Manually tested drag & drop with nested slot components

### DIFF
--- a/packages/core/lib/data/default-slots.ts
+++ b/packages/core/lib/data/default-slots.ts
@@ -3,6 +3,9 @@ import { Fields } from "../../types";
 export const defaultSlots = (value: object, fields: Fields) =>
   Object.keys(fields).reduce(
     (acc, fieldName) =>
-      fields[fieldName].type === "slot" ? { [fieldName]: [], ...acc } : acc,
+      fields[fieldName].type === "slot"
+        ? { ...acc, [fieldName]: Array.isArray((acc as Record<string, unknown>)[fieldName]) ? (acc as Record<string,
+            unknown>)[fieldName] : [] }
+        : acc,
     value
   );
--- a/packages/core/lib/data/map-fields.ts
+++ b/packages/core/lib/data/map-fields.ts
@@ -67,7 +67,7 @@ export const walkField = ({
   const map = mappers[fieldType];
 
   if (map && fieldType === "slot") {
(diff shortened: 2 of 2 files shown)


packages/core/lib/data/default-slots.ts
```diff
@@ -3,6 +3,9 @@ import { Fields } from "../../types";
 export const defaultSlots = (value: object, fields: Fields) =>
   Object.keys(fields).reduce(
     (acc, fieldName) =>
-      fields[fieldName].type === "slot" ? { [fieldName]: [], ...acc } : acc,
+      fields[fieldName].type === "slot"
+        ? { ...acc, [fieldName]: Array.isArray((acc as Record<string, unknown>)[fieldName]) ? (acc as Record<string,
+            unknown>)[fieldName] : [] }
+        : acc,
```
chrisvxd, line 9: Can you break this down a bit? I found it a bit hard to see what's going on.

jderboven: Working on it to validate the whole flow, might be a data issue so want to make sure of it !

jderboven: @chrisvxd i found the issue, we were using some "reserved words" in the slots, like "breadcrumb", puck did not really like that, I'm closing the PR ;-)

FedericoBonel: Aha interesting, we should maybe document this, was that the name for the slot fields?

PR #1572, opened 2026-02-17
Type safe allow and disallow

### LINKED ISSUE #1099: Ability to make allow and disallow type safe
## Description

While using `DropZone`, it was possible to override the component interface to make both `allow` and `disallow` type safe.  In practice, we created a thin wrapper around `DropZone`, this is how it looks like in our case:

```tsx
import { DropZone as PuckDropZone } from '@measured/puck';
import { styled } from '@mui/material';

import type { WidgetKeyValue } from '../registry';

interface DropZoneProps extends React.ComponentPropsWithRef<typeof PuckDropZone> {
  /**
   * Only allows the specified widgets to be added.
   *
   * By default, any widgets are allowed.
   */
  allow?: WidgetKeyValue[];

  /**
   * Prevents the specified widgets to be added.
   *
   * By default, no widgets are disallowed
   */
  disallow?: WidgetKeyValue[];
}

// Wrapper to make `allow` and `disallow` type safe and add sxProps support as well
export const DropZone = styled(PuckDropZone)<DropZoneProps>();
```

It prevented engineers from providing wrong keys and saved time during development and debugging.

Ideally, we should be able to define the same behavior with the new slot API.

## Considerations

A possible workaround alternative is to rely on a constant like `Widget.KeyOfMyComponent` but it doesn't guarantee type safety during development and can fails at runtime, which we want to avoid.

### Proposal 

Add  a new generic parameter  is added to  the `PuckComponent`  interface. The second generic parameter would allow to specify the supported keys for both `allow` and `disallow`.  It would also require to update the interface for `SlotComponent` and `DropZoneProps` in order to use the new generic parameter.

```ts
type DropZoneProps<T extends string = string> = {
    zone: string;
    allow?: T[];
    disallow?: T[];
    style?: CSSProperties;
    minEmptyHeight?: number;
    className?: string;
    collisionAxis?: DragAxis;
};

type SlotComponent<ComponentKeys = string> = (props?: Omit<DropZoneProps<ComponentKeys>, "zone">) => ReactNode;

type PuckComponent<Props, ComponentKeys = string> = (props: WithId<WithPuckProps<{
    [K in keyof Props]: WithDeepSlots<Props[K], SlotComponent<ComponentKeys> >;
}>>) => JSX.Element;


// usage

interface Props {
  Slot: Slot;
}

const MyComponent: PuckComponent<Props, "foo" | "bar"> = ({ Slot }) => {
  return <Slot 
               allow={["foo"]}  // OK + autocomplete
               disallow={["xxx"]} // won't compile
             />
}
```

DamianKocjan:
Closes puckeditor/puck#1099

## Description

This PR introduces type safety for the `allow` and `disallow` props within Puck's `Slot`/`DropZone`.

## Changes made

Updated `DropZoneProps` to use the generic component type for `allow` and `disallow` arrays.

## How to test

Open a component configuration in a TypeScript environment (e.g., `apps/demo/config/blocks/Flex/index.tsx`).

Ensure the component is typed using the new generic: `ComponentConfig<Props, keyof Components>`.

Inside the render function, find a `Slot`.

Try to add an allow prop:

```tsx
<Slot allow={["NonExistentComponent"]} /> 
```

Verify: TypeScript should throw an error because `"NonExistentComponent"` is not a key in your `Components` type.

Verify: Triggering autocomplete within the `allow` array should suggest valid component names from your config.

## Summary by CodeRabbit

* **New Features**
  * Component and slot configurations can explicitly define available components.
  * Drop zones and nested fields support typed allow/disallow component restrictions.
  * Shared field configuration supports reusable custom user-field definitions.
* **Refactor**
  * Standardized component and root configuration typing across the demo experience.
  * No runtime behavior changes are expected.

## Summary by CodeRabbit

* **Type Safety**
  * Component and slot configurations now restrict selectable components to those available in the current configuration.
  * Drop zones and nested fields provide stronger validation for allowed and disallowed components.
  * Shared field definitions improve consistency across configuration options.

* **Maintenance**
  * Updated component configurations and test fixtures to use the enhanced configuration typing.
  * Removed an obsolete Rich Text Editor export.


packages/core/components/RichTextEditor/index.ts
```diff
@@ -1 +0,0 @@
-export * from "./Editor";
```
chrisvxd, line 1: Accidental?

apps/demo/config/blocks/Template/client.tsx
```diff
@@ -27,7 +27,10 @@ async function createComponent<T extends keyof Components>(
 
 type TemplateData = Record<string, { label: string; data: Slot }>;
 
-export const TemplateInternal: ComponentConfig<TemplateProps> = {
+export const TemplateInternal: ComponentConfig<
+  TemplateProps,
+  keyof Components
+> = {
```
chrisvxd: I'm a little hesitant of this API because the config types are already on the edge as it is.

I think a cleaner API would be using our new format:

```
ComponentConfig<{
  props: TemplateProps,
  availableComponents: keyof Components
}>
```

Worth adding, this will be a `feat` so we'll hold off merge until closer to `0.22`

chrisvxd (changes requested):
Thanks @DamianKocjan! Some thoughts in the comments.

packages/core/components/RichTextEditor/index.ts
DamianKocjan: There's no `Editor` file or folder

apps/demo/config/blocks/Template/client.tsx
DamianKocjan: I'll refactor later this week to use new format ;)

DamianKocjan: @chrisvxd Took me more time than I anticipated, but it's done now. Sorry for waiting

packages/core/components/DropZone/types.ts
```diff
@@ -1,10 +1,10 @@
 import { CSSProperties, ElementType, Ref } from "react";
 import { DragAxis } from "../../types";
 
-export type DropZoneProps = {
+export type DropZoneProps<T = string> = {
```
DamianKocjan: I'm not sure if `T`should be as it is. Maybe something more descriptive like `Components` or `ComponentKeys`

packages/core/types/Config.tsx
```diff
@@ -14,12 +14,14 @@ import {
   WithDeepSlots,
 } from "./Internal";
 
-export type SlotComponent = (props?: Omit<DropZoneProps, "zone">) => ReactNode;
+export type SlotComponent<T = string> = (
```
DamianKocjan: Same goes here

packages/core/types/Config.tsx
```diff
@@ -14,12 +14,14 @@ import {
...
+  props?: Omit<DropZoneProps<T>, "zone">
+) => ReactNode;
 
-export type PuckComponent<Props> = (
+export type PuckComponent<Props, T = string> = (
```
DamianKocjan: And here...

DamianKocjan:
Hi @chrisvxd, any update on this?

chrisvxd:
Apologies been stacked. Asking @FedericoBonel to take over.

DamianKocjan:
No problem;)

packages/core/components/DropZone/types.ts
FedericoBonel: Yes, please lets use "available components" to be consistent with our entry prop

packages/core/types/Config.tsx
FedericoBonel: Same here

packages/core/types/Config.tsx
FedericoBonel: And here

packages/core/types/Config.tsx
```diff
@@ -252,8 +273,10 @@ export type ConfigParams<
 
 export type ComponentConfigParams<
   Props extends DefaultComponentProps = DefaultComponentProps,
+  Components extends string = string,
   UserFields extends FieldsExtension = never
```
FedericoBonel: This is a breaking change, people passing user fields as the second parameter would now suddenly get a type mismatch and unexpected behaviors.

Please pass it as the third parameter.


```suggestion
export type ComponentConfigParams<
  Props extends DefaultComponentProps = DefaultComponentProps,
  UserFields extends FieldsExtension = never,
  Components extends string = string
```

FedericoBonel (changes requested):
Hey @DamianKocjan! Thank you very much for the contribution and for the work. Sorry for the delay on getting back to you.

I'd say is looking good, but I have a few comments, and also this is not working with definitions at the config level (which are the current recommended way of applying this API):

```tsx
fields: {
  mySlot: {
    type: "slot",
    allow: ["Random string"] // passes
  }
}
```

We need to also add these generics to the field types so that people and us can pass them through.


One last thing, we also need to merge/rebase from main since I think there're some conflicts.

DamianKocjan:
@FedericoBonel I'm not sure how we should handle those type errors in test file. Should I cast all to `Config` type?

FedericoBonel:
Hey @DamianKocjan! The CI is failing because of formatting, you need to run `yarn format` from the root.

Regarding your question, do you mean the ones you added a `@ts-expect-error` to?

DamianKocjan:
@FedericoBonel now you can check error in ci. There're more of them in test files (ci doesn't run `tsc`, so it doesn't show).

PR #1575, opened 2026-02-20
fix: remove the behavior of de-select on second click

### LINKED ISSUE #1567: Remove "de-select component on second click" behavior
## Description

Updating puck to `v0.21.0`, you've changed the behavior of clicking a component; toggling selection. https://github.com/puckeditor/puck/commit/ccff0d8e33f5424556fff398667178a0d8bb79f7. Clicking an already-selected component toggles selection off (`itemSelector` → `null`). Usually, in web builders like figma, `click` usually means "select/confirm this for editing," not "deselect." The same click doing different things (select vs. deselect) is unpredictable and can cause unintended context switches.

**Expected outcome:** Click always selects the component. Deselect only via a dedicated action (e.g. click canvas background), not by re-clicking the selected component.

## Considerations

- Implemented in `DraggableComponent/index.tsx` `onClick`: `isSelected` branches to `itemSelector: null` vs. set current component.
https://github.com/puckeditor/puck/blob/e16ff6b5f246ebb2e8807dcfbab4318d335dcbd5/packages/core/components/DraggableComponent/index.tsx#L362-L376

- Other builders (e.g. Figma) typically do not toggle selection on re-click.
- Design feedback: click = "confirm/edit this section"; re-click deselecting breaks that flow and causes confusion.
- Users who rely on "click to deselect" would need another way to clear selection (e.g. click-outside).

### Proposal 1: Always select on click (remove toggle)

**Behavior:** In `DraggableComponent`, when the user clicks a component, always dispatch a `setUi` action that sets `itemSelector` to the clicked component’s `{ index, zone }`. Remove the branch that sets `itemSelector: null` when `isSelected` is true.

**Pros:**

- Consistent mental model: "click = select this component."
- Aligns with common web builder / design-tool patterns.
- Reduces risk of accidental deselection and context switching.
- No new UI surface; only behavior change.

**Cons:**

- Users who currently "click to deselect" lose that shortcut and must use another method (e.g. click on canvas/background) if one exists.

---

### Alternative: Make toggle behavior configurable

If the project wants to support both behaviors, an optional prop or app-level config (e.g. `deselectOnReClick?: boolean`) could be added, defaulting to `false` (always select on click) for the reasons above.

```tsx
<Puck.Preview deselectOnClick={false} />
```

ojj1123:
close #1567

### DIFF
--- a/packages/core/components/DraggableComponent/index.tsx
+++ b/packages/core/components/DraggableComponent/index.tsx
@@ -114,9 +114,7 @@ export const DraggableComponent = ({
   userDragAxis?: DragAxis;
   inDroppableZone: boolean;
 }) => {
-  const {
-    _experimentalFullScreenCanvas
-  } = usePropsContext();
+  const { _experimentalFullScreenCanvas } = usePropsContext();
 
   const zoom = useAppStore((s) =>
     s.selectedItem?.props.id === id ? s.zoomConfig.zoom : 1


FedericoBonel:
Hey @ojj1123! 

Thanks for taking a look at this, would you mind actually only adding the deselecting behavior if the [`_experimentalFullScreenCanvas`](https://puckeditor.com/docs/api-reference/components/puck#_experimentalfullscreencanvas) is enabled? 

This new mode was the main reason this behavior was introduced, so I think it makes sense to only enable it if that is `true` instead of fully getting rid of it.

ojj1123:
> would you mind actually only adding the deselecting behavior if the [_experimentalFullScreenCanvas](https://puckeditor.com/docs/api-reference/components/puck#_experimentalfullscreencanvas) is enabled?

@FedericoBonel I changed it. Could you please review this PR?

packages/core/components/DraggableComponent/index.tsx
```diff
@@ -113,6 +114,10 @@ export const DraggableComponent = ({
   userDragAxis?: DragAxis;
   inDroppableZone: boolean;
 }) => {
+  const {
+    _experimentalFullScreenCanvas
+  } = usePropsContext();
```
ojj1123: `DraggableComponent` is rendered as a child element of the `PropsProvider` and so I used that context.

Drewbiee123:
Thanks for the update, @ojj123! Gating that deselecting behavior behind _experimentalFullScreenCanvas is the right move for maintaining stability. Ill review the logic in commit 0f1a253 now to ensure it aligns with our dispatch patterns


Sent from Yahoo Mail for iPhone


On Sunday, February 22, 2026, 11:47 AM, Jeongjin Oh ***@***.***> wrote:





@ojj1123 commented on this pull request.

In packages/core/components/DraggableComponent/index.tsx:
> +  const {
+    _experimentalFullScreenCanvas
+  } = usePropsContext();

DraggableComponent is rendered as a child element of the PropsProvider and so I used that context.

ojj1123:
@chrisvxd I've missed it. I fixed it.

FedericoBonel:
I'm going to close this PR since #1601 adds the fix with the store.

chrisvxd:
Reopening as #1601 has outstanding comments and I'm keen to get this merged into main

ojj1123:
huge thanks for your work!

PR #1577, opened 2026-02-20
fix: richtext inline focus problem with sync-editor

### LINKED ISSUE #1557: Inline text formatting in RichText field doesn't call `onChange`
## Description

## Environment

- Puck version: 0.21.1

## Steps to reproduce

## What happens

## What I expect to happen

## Additional Media

Xtron33:
Closes #1557 

## Description

This PR remove overengineering  logic from Editor for checking focus. 

## Changes made

- Delete focusName variable
- Pass `name` to `useSyncedEditor`

Old logic make problem with first `onChange` call when app think what focus on not-inline editor, but call useSyncedEditor from inline editor cause onChange not work.

Using field-name for both work well. 

## How to test

- Use inline richtext editor on `Hero` component.

https://github.com/user-attachments/assets/bfb8dedb-6c1e-4e24-bebf-cc58a1824803

### DIFF
--- a/packages/core/components/RichTextEditor/components/Editor.tsx
+++ b/packages/core/components/RichTextEditor/components/Editor.tsx
@@ -29,14 +29,12 @@ export const Editor = memo((props: EditorProps) => {
 
   const appStoreApi = useAppStoreApi();
 
-  const focusName = `${name}${inline ? "::inline" : ""}`;
-
   const editor = useSyncedEditor({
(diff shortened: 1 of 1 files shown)


chrisvxd:
Thanks @Xtron33, but I'm closing this in favour of the fix in #1587.

This fix creates an unwanted history entry when the user first clicks on the inline field.

PR #1583, opened 2026-02-28
fix: render richtext fields in ServerRender dropzones

### LINKED ISSUE #1580: RichText in DropZones renders escaped HTML (in RSC mode)
## Description

When using `Render` from `@puckeditor/core/rsc`, RichText components inside DropZones display escaped HTML (`&lt;p&gt;Hello&lt;/p&gt;`) instead of rendered markup.

**Suspected Cause(?):** `DropZoneRender` in the RSC bundle doesn't call `useRichtextProps()` to process richtext field data before passing props to components. The `Item` component does call it, but zone content bypasses this.

**Workaround:** Use slot fields instead of DropZones. Slots are processed through `useRichtextProps` correctly:

## Environment

- Puck version: [0.21.1]
- Browser version: NA
- Additional environment info: [Deno]

## Steps to reproduce

Steps, or see minimal reproduction below.

1. Create a Section component with a DropZone
2. Add a RichText component (using `type: "richtext"` field) inside the zone
3. Render server-side with `Render` from `@puckeditor/core/rsc`
4. Output shows raw HTML tags instead of rendered content

```ts
/** @jsxImportSource react */
// repro.tsx
import { Render } from "@puckeditor/core/rsc";
import { Config, Data } from "@puckeditor/core";
import { renderToString } from "react-dom/server";


const config: Config = {
  components: {
    Section: {
      fields: {},
      render: ({ puck }) => (
        <div>{puck.renderDropZone({ zone: "content" })}</div>
      ),
    },
    RichText: {
      fields: {
        content: { type: "richtext" },
      },
      render: ({ content }) => <div>{content}</div>,
    },
  },
};

const data: Data = {
  root: { props: {} },
  content: [
    { type: "Section", props: { id: "section-1" } }
  ],
  zones: {
    "section-1:content": [
      {
        type: "RichText",
        props: {
          id: "richtext-1",
          content: "<p>Hello world</p>",
        },
      },
    ],
  },
};

const html = renderToString(<Render config={config} data={data} />);
console.log(html);
// Expected: <div><div><p>Hello world</p></div></div>
// Actual:   <div><div>&lt;p&gt;Hello world&lt;/p&gt;</div></div>
```

## What happens

Html is rendered with entities encoded:

`<div><div>&lt;p&gt;Hello world&lt;/p&gt;</div></div>`

## What I expect to happen

HTMLl is rendered without entities encoded:

` <div><div><p>Hello world</p></div></div>`

fancyboi999:
## Summary

Fixes `Render` from `@puckeditor/core/rsc` so RichText fields inside DropZones are transformed before rendering.

- add a dedicated `DropZoneRenderItem` component in `ServerRender`
- apply `useRichtextProps` to dropzone item props (matching client/slot render behavior)
- add a regression test covering server rendering of richtext inside a dropzone

Closes #1580.

## Verification

- `cd packages/core && yarn test ServerRender --runInBand`
- `yarn -s eslint components/ServerRender/index.tsx components/ServerRender/__tests__/index.spec.tsx`
- `yarn -s prettier --check packages/core/components/ServerRender/index.tsx packages/core/components/ServerRender/__tests__/index.spec.tsx`

### DIFF
--- a/apps/demo/config/blocks/Hero/render-fields.ts
+++ b/apps/demo/config/blocks/Hero/render-fields.ts
@@ -1,18 +1,16 @@
-import { ComponentConfig } from "@/core/types";
+import { ObjectField, RichtextField, SlotField } from "@/core/types";
 import { HeroProps } from "./Hero";
 
-export const heroRenderFields: NonNullable<
-  ComponentConfig<HeroProps>["fields"]
-> = {
+export const heroRenderFields = {
   description: {
     type: "richtext",
-  },
+  } satisfies RichtextField,
   image: {
     type: "object",
     objectFields: {
       content: {
         type: "slot",
-      },
+      } satisfies SlotField,
     },
-  },
+  } satisfies ObjectField<HeroProps["image"]>,
 };
--- a/apps/demo/config/blocks/Hero/server.tsx
+++ b/apps/demo/config/blocks/Hero/server.tsx
@@ -4,6 +4,6 @@ import HeroComponent, { HeroProps } from "./Hero";
 import { heroRenderFields } from "./render-fields";
 
 export const Hero: ComponentConfig<HeroProps> = {
-  fields: heroRenderFields,
+  fields: heroRenderFields as ComponentConfig<HeroProps>["fields"],
   render: HeroComponent,
 };
--- a/apps/demo/config/blocks/Template/render-fields.ts
+++ b/apps/demo/config/blocks/Template/render-fields.ts
@@ -1,9 +1,9 @@
-import { ComponentConfig } from "@/core/types";
+import { SlotField } from "@/core/types";
 import { TemplateProps } from "./Template";
 
-export const templateRenderFields: NonNullable<
-  ComponentConfig<TemplateProps>["fields"]
-> = {
+export const templateRenderFields: {
+  children: SlotField;
+} = {
   children: {
     type: "slot",
   },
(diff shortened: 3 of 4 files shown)


fancyboi999:
Thanks, I dug into this locally and was able to reproduce what you were seeing on `/rsc`.

The issue turned out to be broader than the original DropZone path. My previous change fixed one `ServerRender` branch, but the real `/rsc` demo was still using a server config that removed field definitions entirely for some components. That meant the RSC render path had no way to know which props needed render-time transforms like `richtext` and `slot`, so values like `Hero.description` were still rendered as escaped strings.

I pushed a follow-up commit that keeps the minimal render-relevant field metadata in the server config, while still avoiding editor-only config on the server. I also added a regression test covering the case where RSC rendering depends on those fields being present.

I re-ran the local checks and verified `/rsc` no longer renders raw `<p>` tags as text.

fancyboi999:
I pushed one more follow-up commit for CI.

The failing `demo#build` check was a TypeScript issue from the shared render-field refactor, not another RSC rendering regression. The shared `description` field was getting widened during inference, so the demo build stopped recognizing it as a `richtext` field in `Hero/client.tsx`.

I narrowed those shared render-field types and re-ran `apps/demo` production build locally. It now completes successfully.

chrisvxd:
Wonderful thanks @fancyboi999

earthlingdavey:
Thanks @fancyboi999 !

Cheers @FedericoBonel @chrisvxd !

PR #1601, opened 2026-03-20
Move `_experimentalFullScreenCanvas` to zustand store

### LINKED ISSUE #1567: Remove "de-select component on second click" behavior

### LINKED ISSUE #1599: Puck props changes re-render all draggable components in the page
## Description

#1575 enabled second-click deselection only when `_experimentalFullScreenCanvas` is set to `true`, but the implementation introduced a dependency on the `Puck` props in every draggable component. This means that any changes to the props will trigger a re-render in every component in the preview. This is unnecessary and may affect performance, depending on the Puck integration.

To avoid this, we should add a slice to the store for `_experimentalFullScreenCanvas` and use that instead.

ojj1123:
Closes #1567
Closes #1599

- This PR is stacked on https://github.com/puckeditor/puck/pull/1575

### DIFF
--- a/packages/core/components/DraggableComponent/index.tsx
+++ b/packages/core/components/DraggableComponent/index.tsx
@@ -34,7 +34,6 @@ import { accumulateTransform } from "../../lib/accumulate-transform";
 import { useContextStore } from "../../lib/use-context-store";
 import { useOnDragFinished } from "../../lib/dnd/use-on-drag-finished";
 import { LoadedRichTextMenu } from "../RichTextMenu";
-import { usePropsContext } from "../Puck";
 
 const getClassName = getClassNameFactory("DraggableComponent", styles);
 
@@ -114,7 +113,9 @@ export const DraggableComponent = ({
   userDragAxis?: DragAxis;
   inDroppableZone: boolean;
 }) => {
-  const { _experimentalFullScreenCanvas } = usePropsContext();
+  const _experimentalFullScreenCanvas = useAppStore(
+    (s) => s._experimentalFullScreenCanvas
+  );
 
   const zoom = useAppStore((s) =>
     s.selectedItem?.props.id === id ? s.zoomConfig.zoom : 1
--- a/packages/core/components/Puck/index.tsx
+++ b/packages/core/components/Puck/index.tsx
@@ -124,6 +124,7 @@ function PuckProvider<
     metadata,
     onAction,
     fieldTransforms,
+    _experimentalFullScreenCanvas = false,
   } = usePropsContext();
 
   const iframe: IframeConfig = useMemo(
@@ -265,6 +266,7 @@ function PuckProvider<
         onAction,
         metadata,
         fieldTransforms: loadedFieldTransforms,
+        _experimentalFullScreenCanvas,
       };
     },
     [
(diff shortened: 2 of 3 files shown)


ojj1123:
NOTE : I've wanted to stack this PR on https://github.com/puckeditor/puck/pull/1601 but I was able to set base branch to only `main`.

packages/core/components/DraggableComponent/index.tsx
```diff
@@ -359,11 +363,11 @@ export const DraggableComponent = ({
         e.stopPropagation();
       }
 
-      if (isSelected) {
+      if (_experimentalFullScreenCanvas) {
         dispatch({
           type: "setUi",
           ui: {
-            itemSelector: null,
+            itemSelector: isSelected ? null : { index, zone: zoneCompound },
```
FedericoBonel, line 370:
```suggestion
      // Deselect component on second click when fullscreen is enabled
      if (isSelected && _experimentalFullScreenCanvas) {
dispatch({
type: "setUi",
ui: {
            itemSelector: null,
```

I think adding a little comment and checking for is selected here is a bit clearer on intent and avoids repeating the reselection we have underneath

chrisvxd:
Closing in favour of #1616 (fixed the tests)

PR #1621, opened 2026-03-31
Virtualization & performance improvements

### LINKED ISSUE #1492: Performance degradation with large nested documents
# Performance degradation with large nested documents (~18 seconds blocking time for mutations)

-- This report was written with help from an AI, that said I am a human 🙃 --

## Description

We're experiencing severe performance issues when performing mutations (delete, duplicate, paste) on documents with many pages and nested components. A single delete operation blocks the main thread for **~18 seconds**, making the editor unusable for large documents.

## Environment

- **Puck version:** 0.20.2
- **React version:** 19.0.0
- **Node version:** 22.16.0
- **Browser:** Chrome (latest)
- **OS:** macOS

## Document Structure

Our documents are PDF-like with:

- **23 pages** (using a custom `Page` component with `type: 'slot'`)
- Each page contains **10+ nested components** (tables, text blocks, images)
- **~230 total components** across all pages

## Performance Analysis

We instrumented the codebase extensively to isolate the bottleneck:

| Metric                              | Value                         | % of Total      |
| ----------------------------------- | ----------------------------- | --------------- |
| **Total blocking time**       | 17,862ms                      | 100%            |
| **Our component render time** | 42ms (98 renders @ 0.4ms avg) | **0.2%**  |
| **Puck internal processing**  | ~17,820ms                     | **99.8%** |

### Timing Breakdown

```
Action START:        T+0ms
First component renders complete: T+100ms
                     ↓
         ══════════ 12 SECOND GAP ══════════  ← Puck internals
                     ↓
Second render batch: T+12,100ms
                     ↓
         ══════════ 5 SECOND GAP ══════════   ← Puck internals
                     ↓
Main thread free:    T+17,862ms
```

## Key Finding

**99.8% of the time is spent BETWEEN component render batches**, not in our component code. Our components render in <1ms each. The blocking time appears to be in Puck's internal state management, likely:

1. **History/undo system** - Deep cloning the entire state tree
2. **Zone reconciliation** - Processing deeply nested slot structures
3. **Multiple render passes** - For selection, overlays, fields, etc.

## Reproduction

1. Create a document with 20+ pages using slot-based `Page` components
2. Add 10+ components to each page (tables, text, etc.)
3. Perform a delete/duplicate/paste operation
4. Observe the UI freeze for 15-20 seconds

## Expected Behavior

Mutations should complete in <500ms for a responsive editing experience.

## Actual Behavior

Mutations block the main thread for 15-20 seconds, making the editor unusable for production documents.

## Questions

1. Is there a way to disable or debounce history recording for better performance?
2. Are there recommended patterns for optimizing deeply nested slot structures?
3. Is there a performance mode or lazy rendering option we're missing?

## Workarounds Attempted

- ✅ Memoized all custom components with `React.memo` - No significant improvement
- ✅ Verified our component render times are <1ms each

We're happy to provide more details or help test potential fixes. This is a critical blocker for our production use case.

chrisvxd:
Closes #1492 

## Description

This PR adds virtualization to both the main canvas and outline to dramatically improve performance. 

Virtualization is enabled by default, with no opt-out. If we want opt-out, this will become a feature release with new APIs.

## Benchmarks

Test page with 400 components. We're only reporting initial load, but similar performance improvements are seen across the board:

* Control: 137s
* Improvements, virtualization off: 2.09s (-98.5% reduction)
* Improvements, virtualization on: 0.71s (-99.5% reduction)

## Changes made

Please review commit history.

## How to test

1. Run demo app
2. Visit /performance route, which loads 100 deep items onto the page
3. 🚀

### DIFF
--- a/apps/demo/config/blocks/Hero/client.tsx
+++ b/apps/demo/config/blocks/Hero/client.tsx
@@ -17,7 +17,6 @@ export const Hero: ComponentConfig<{
   };
 }> = {
   fields: {
-    ...heroRenderFields,
     quote: {
       type: "external",
       placeholder: "Select a quote",


packages/core/store/slices/nodes.ts
```diff
@@ -1,73 +1,57 @@
...
-        sync: () => null,
-        hideOverlay: () => null,
-        showOverlay: () => null,
-      },
-      element: null,
-    };
-
-    const existingNode: PuckNodeInstance | undefined = s.nodes[id];
+  _set: (newState: Partial<AppStore>) => void,
+  _get: () => AppStore
```
FedericoBonel, line 19: Any reasons we leave these even if we're not using them?

packages/core/components/DraggableComponent/index.tsx
```diff
@@ -363,23 +366,31 @@ export const DraggableComponent = ({
...
+    sync: () => null,
+    hideOverlay: () => null,
+    showOverlay: () => null,
+  });
+
+  const registeredNodeHandleRef = useRef<NodeHandle>({
+    sync: () => nodeHandleRef.current.sync(),
+    hideOverlay: () => nodeHandleRef.current.hideOverlay(),
+    showOverlay: () => nodeHandleRef.current.showOverlay(),
+  });
```
FedericoBonel: I assume there's a reason, but why we're using two refs here? 

What's the purpose of `registeredNodeHandleRef`?

packages/core/store/index.ts
```diff
@@ -305,7 +305,11 @@ export const createAppStore = (initialAppStore?: Partial<AppStore>) =>
           state,
           config,
           (content) => content,
-          (childItem) => {
+          (childItem, path) => {
+            // Skip nested items as these get processed by resolveData
```
FedericoBonel:
```suggestion
            // Skip nested items as these get processed by the parent's resolveComponentData
```

packages/core/components/DropZone/VirtualizedDropZone.tsx
```diff
@@ -0,0 +1,265 @@
+import {
```
FedericoBonel, line 1: I imagine you already looked into it and it might have had limitations for our use case, but any reasons why we decided to do dynamic measuring manually instead of using their solution?

https://tanstack.com/virtual/latest/docs/framework/react/examples/dynamic?panel=code

FedericoBonel:
Looks Amazing, left a couple of small comments, but approving in case you need to merge before me looking at it again.

packages/core/store/slices/nodes.ts
chrisvxd: Not used but standard slice pattern

packages/core/components/DraggableComponent/index.tsx
chrisvxd: booboo, fixed

PR #1626, opened 2026-04-03
fix: stabilize plugin override component references to prevent remounts

### LINKED ISSUE #1625: Plugin override components remount on every overrides prop change
## Description

When a plugin provides a component-level override (e.g. overrides.puck), any change to the prop-level overrides object passed to `<Puck>` causes the plugin's override subtree to fully unmount and remount - even if the plugin itself hasn't changed.

`loadOverrides` (packages/core/lib/load-overrides.ts:39-44) wraps every plugin override in a new inline arrow function on each invocation:

  ```
const childNode = collected[overridesType];
const Comp = (props: any) =>                    // ← new function identity every call
    plugin.overrides![overridesType]!({
      ...props,
      children: childNode ? childNode(props) : props.children,
    });
collected[overridesType] = Comp;
```
`useLoadedOverrides` re-runs loadOverrides whenever overrides change:

```
return useMemo(() => loadOverrides({ overrides, plugins }), [plugins, overrides]);
```

The result is then used as a React component type in Layout/index.tsx:138-141,272:

``` 
const CustomPuck = useMemo(() => overrides.puck || DefaultOverride, [overrides]);
  // ...
<CustomPuck>...</CustomPuck>
```

Because Comp is a new function reference, React sees a different component type and unmounts the old tree before mounting a new one.

Any consumer that passes an override prop that changes over time (e.g. headerActions displaying save state, unsaved-changes indicators, etc.) will trigger a full remount of all plugin override subtrees on every change.

## Environment

- Puck version: 0.21.2

## Steps to reproduce

```
import { useEffect, useState } from "react";
  import { Puck, type Config, type Plugin } from "@puckeditor/core";

  const config: Config = {
    components: {
      Text: {
        render: () => <p>Hello</p>,
      },
    },
  };

  const data = { content: [], root: { props: {} } };

  const myPlugin: Plugin = {
    overrides: {
      puck: ({ children }) => {
        useEffect(() => {
          console.log("puck override MOUNTED");
          return () => console.log("puck override UNMOUNTED");
        }, []);
        return <>{children}</>;
      },
    },
  };

    const [count, setCount] = useState(0);

    return (
      <>
        <button onClick={() => setCount((c) => c + 1)}>
          Re-render ({count})
        </button>
        <Puck
          config={config}
          data={data}
          plugins={[myPlugin]}
          overrides={{
            headerActions: ({ children }) => (
              <>
                {children}
                <span>Count: {count}</span>
              </>
            ),
          }}
        />
      </>
    );
  }
```

  1. Open the console
  2. Click the "Re-render" button

## What happens

Every click logs "puck override UNMOUNTED" followed by "puck override MOUNTED". The entire subtree wrapped by the plugin's puck override is destroyed and recreated, and the editor blinks.

## What I expect to happen

The plugin override stays mounted because the plugin hasn't changed.

imironyak:
Closes #1625 

## Description

This PR introduces a WeakMap-based cache in loadOverrides that returns the same wrapper component for a given (plugin.overrides, overrideType/fieldType) pair, as long as the upstream childNode in the currying chain hasn't changed.

## Changes made

A `WeakMap<object, Map<string, CacheEntry>>` keyed by the plugin's overrides object (or fieldTypes sub-object) stores the last wrapper component created for each override type. On cache hit - same childNode reference - the existing component is returned, keeping React's reconciliation stable.

### DIFF
--- a/packages/core/lib/load-overrides.ts
+++ b/packages/core/lib/load-overrides.ts
@@ -1,4 +1,42 @@
+import type { FC } from "react";
 import { Overrides, Plugin } from "../types";
+
+type CacheEntry = {
+  comp: FC<any>;
+  childNode: ((props: any) => any) | undefined;
+};
+
+const wrapperCache = new WeakMap<object, Map<string, CacheEntry>>();
+
+function getOrCreateWrapper(
+  cacheKey: object,
+  name: string,
+  overrideFn: (props: any) => any,
+  childNode: ((props: any) => any) | undefined
+): FC<any> {
+  let typeMap = wrapperCache.get(cacheKey);
+  if (!typeMap) {
+    typeMap = new Map();
+    wrapperCache.set(cacheKey, typeMap);
+  }
+
+  const entry = typeMap.get(name);
+
+  if (entry && entry.childNode === childNode) {
+    return entry.comp;
+  }
+
+  const comp: FC<any> = (props: any) =>
+    overrideFn({
+      ...props,
+      children: childNode ? childNode(props) : props.children,
+    });
+
+  comp.displayName = `PluginOverride(${name})`;
+  typeMap.set(name, { comp, childNode });
+
+  return comp;
+}
 
 export const loadOverrides = ({
   overrides,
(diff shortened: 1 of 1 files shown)


FedericoBonel:
Thanks for the contribution! 
I'll be closing this as the upstream issue got closed. 

Interesting approach though, I'll keep it tabbed somewhere for future reference.

PR #1647, opened 2026-04-17
fix: preserve caret position in controlled custom fields

### LINKED ISSUE #1642: Cursor jumps to end in controlled custom input fields
## Description

When using custom fields with a controlled `<input>` element (i.e. passing a `value` prop), the cursor jumps to the end of the input on every keystroke. This makes editing text inside custom fields quite frustrating, as it's only really possible to append characters at the end.

## Environment

Puck version: 0.21.2
Browser version: Chrome 135 (desktop)
Additional environment info: Next.js 16, React 19, Turbopack

## Steps to reproduce

1. Create a custom field with a controlled `<input>`:

```
const MyTextField = ({ id, onChange, value }: CustomFieldProps<string>) => {
  return (
    <input
      id={id}
      value={value}
    />
  )
}
```

2. Register it as a custom field in the Puck config:

```
const config = {
  components: {
    MyComponent: {
      fields: {
        title: {
          type: "custom",
          render: (props) => <MyTextField {...props} />,
        },
      },
    },
  },
}
```

Open the editor, select `MyComponent`, and try typing somewhere in the middle of an existing value in the `title` field.

## What happens

The cursor jumps to the end of the input after each keystroke, making it impossible to edit text in place.

## What I expect to happen

The cursor should remain at the position where the character was inserted, consistent with standard browser behavior for text inputs.

## Additional Media

https://github.com/user-attachments/assets/8fb4ac6a-35f8-46a8-acca-13ad89187c70

rinormhi:
Closes #1642

## Description

This PR fixes a bug in Puck where controlled custom input fields would move the caret to the end on every keystroke while editing existing text.

The root cause was that `custom` fields read from the local field store, but their value was only reflected back after the async `resolveComponentData -> replace` roundtrip completed. For controlled inputs, that delayed sync was enough to break normal browser caret preservation.

This change keeps the existing update flow intact, but synchronizes the local field store immediately for `custom` and override-backed field types so controlled inputs can be edited in place as expected.

## Changes made

- Updated `AutoField` to synchronously mirror changes for `custom` field types into the local field store before forwarding `onChange`.
- Applied the same local sync behavior to override-backed field types, since they use the same value path.
- Added a regression test covering controlled custom field updates.

## How to test

- Create a custom field that renders a controlled `<input>`:

```tsx
const MyTextField = ({ id, value, onChange }) => {
  return (
    <input
      id={id}
      value={value ?? ""}
    />
  );
};

### DIFF
--- a/packages/core/components/AutoField/index.spec.tsx
+++ b/packages/core/components/AutoField/index.spec.tsx
@@ -1,0 +1,61 @@
+import { fireEvent, render, screen } from "@testing-library/react";
+import { appStoreContext, createAppStore } from "../../store";
+import { fieldContextStore } from "./store";
+import { AutoFieldPrivate } from ".";
+
+jest.mock("../Sortable", () => ({
+  Sortable: ({
+    children,
+  }: {
+    children: (props: {
+      isDragging: boolean;
+      ref: null;
+      handleRef: null;
+    }) => JSX.Element;
+  }) => children({ isDragging: false, ref: null, handleRef: null }),
+  SortableProvider: ({ children }: { children: JSX.Element }) => children,
+}));
+
+jest.mock("../../bundle", () => {
+  const { setDeep } = jest.requireActual("../../lib/data/set-deep");
+
+  return { setDeep };
+});
+
+describe("AutoField", () => {
+  it("updates controlled custom fields immediately", () => {
+    const appStore = createAppStore();
+    const onChange = jest.fn();
+
+    render(
+      <appStoreContext.Provider value={appStore}>
+        <fieldContextStore.Provider value={{ title: "Hello world" }}>
+          <AutoFieldPrivate
+            field={{
+              type: "custom",
+              render: ({ id, value, onChange }) => (
+                <input
+                  id={id}
+                  value={value ?? ""}
+                  onChange={(e) => onChange(e.currentTarget.value)}
+                />
+              ),
+            }}
+            id="title"
+            name="title"
+            onChange={onChange}
+          />
+        </fieldContextStore.Provider>
+      </appStoreContext.Provider>
+    );
+
+    fireEvent.change(screen.getByRole("textbox"), {
+      target: { value: "Hello brave world" },
+    });
+
+    expect(onChange).toHaveBeenCalledWith("Hello brave world", undefined);
+    expect((screen.getByRole("textbox") as HTMLInputElement).value).toBe(
+      "Hello brave world"
+    );
+  });
+});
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by Daniil-Padiryakov** [comment]
Independently hit this same bug (controlled `custom` field, caret jumping to the end since 0.21.0). Applied this PR and can confirm it fixes it — both in the `@puckeditor/core` test suite and in my actual project on a patched `0.21.3`. Thanks @rinormhi 🙏 Would be great to get this reviewed.

**Comment by rinormhi** [comment]
@FedericoBonel review pls

File: packages/core/components/AutoField/index.tsx [hunk 1]
```diff
@@ -106,6 +108,22 @@ function AutoFieldInternal<
...
+        fieldStore.setState(
+          setDeep(fieldStore.getState(), props.name ?? resolvedId, value)
+        );
+      }
+
+      props.onChange(value, uiState);
+    },
+    [fieldStore, props.name, props.onChange, resolvedId, shouldSyncLocalValue]
+  );
+
```
FedericoBonel, line 126: This is okay, but I think it looks really similar to the public AutoField path we got above in [`AutoFieldPublicInternal`](https://github.com/puckeditor/puck/pull/1647/changes#diff-318914a232e6c10c884d269a07d4d3fc4c10ee1d2ae72b97b276d3e5a09f0877R264-R273). It also makes it so that, if you use `AutoField` you write to the store two times per change.

I wonder if there's a way we can consolidate that with this here to avoid repetition. What do you think?

packages/core/components/AutoField/index.spec.tsx
```diff
@@ -0,0 +1,61 @@
...
+    }) => JSX.Element;
+  }) => children({ isDragging: false, ref: null, handleRef: null }),
+  SortableProvider: ({ children }: { children: JSX.Element }) => children,
+}));
+
+jest.mock("../../bundle", () => {
+  const { setDeep } = jest.requireActual("../../lib/data/set-deep");
+
+  return { setDeep };
+});
```
FedericoBonel, line 23: Any reasons we're mocking the `/../../bundle` module here? We don't seem to be using it in `AutoField`

packages/core/components/AutoField/index.tsx
```diff
@@ -106,6 +108,22 @@ function AutoFieldInternal<
     }
   });
 
+  const shouldSyncLocalValue =
+    field.type === "custom" || !!overrides.fieldTypes?.[field.type];
```
FedericoBonel, line 112: This is the same check as above (line 105), this one checks when to write, the one above when to read. Could we consolidate it in a variable?

FedericoBonel (changes requested):
Awesome work, thanks for the contribution and I'm really sorry for the delay.

chrisvxd:
@FedericoBonel we have a duplicate implementation here #1724. I'll let you decide which implementation is cleaner.

PR #1652, opened 2026-04-24
Dynamic css injection

### LINKED ISSUE #590: Dynamic CSS injection
Currently, the user is responsible for importing the CSS. This means that Puck CSS commonly ends up in production CSS bundles for rendered pages, not just for pages containing Puck.

Instead, we should inject the CSS dynamically into the head when rendering the `<Puck>` component . This will prevent Puck CSS from ending up in the production CSS bundle.

## Considerations

* The CSS will likely have to be bundled before importing, which may require a custom build pipeline and dev mode

chrisvxd:
Closes #590

## Description

Make server CSS optional, falling back to dynamic CSS injection.

Introduces the `iframe.syncHostStyles` API to prevent the iframe from syncing styles from the host. This is possible as we now dynamically insert the Puck iframe styles.

## How to test

- In the Next.js recipe (not the demo app), remove the `puck.css` import and run a build of Puck. Note that styles still load.
- Disable style sync with `iframe.syncHostStyles: false`. Note that no styles are synced into the iframe. Puck's iframe-only styles are still injected.


apps/docs/pages/docs/api-reference/components/puck.mdx
```diff
@@ -225,6 +227,42 @@ Disabling iframes will also disable [viewports](#viewports).
...
+}
+```
+
+#### uiStyles params
+
+| Param               | Example                 | Type                         | Status |
+| ------------------- | ----------------------- | ---------------------------- | ------ |
+| [`bundle`](#bundle) | `bundle: "no-external"` | `"default" \| "no-external"` | -      |
+
+Pass `false` to disable runtime style injection entirely and load the CSS yourself.
```
FedericoBonel: `uiStyles` has been dropped, so we should remove this section

apps/docs/pages/docs/api-reference/components/puck.mdx
```diff
@@ -225,6 +227,42 @@ Disabling iframes will also disable [viewports](#viewports).
 
 Defer rendering of the Puck preview until the iframe styles have loaded, showing a spinner. Defaults to `true`.
 
+##### `syncHostStyles`
+
+Mirror host `style` and stylesheet tags into the preview iframe. Defaults to `true`.
+
+Set `syncHostStyles: false` to fully isolate the iframe from host styles and host `html` / `body` attributes while still keeping Puck's iframe interaction styles active.
```
FedericoBonel, line 233: This might be a bit unclear and too detailed for some people. What about something like:

```suggestion
Copy host `style` and stylesheet tags into the preview iframe. Defaults to `true`.

Set `syncHostStyles: false` to fully isolate the iframe from host styles and host `html` / `body` attributes.
```

apps/docs/pages/docs/api-reference/components/puck.mdx
```diff
@@ -225,6 +227,42 @@ Disabling iframes will also disable [viewports](#viewports).
...
+
+Pass `false` to disable runtime style injection entirely and load the CSS yourself.
+
+##### `bundle`
+
+Choose which built-in editor UI bundle to inject. Defaults to `"default"`.
+
+- `"default"` loads the standard editor bundle, including the Inter font import.
+- `"no-external"` omits the external font import so you can supply your own font.
+
```
FedericoBonel: I can't see this prop, so we should remove this section

apps/demo/app/runtime-css-debug/client.tsx
```diff
@@ -0,0 +1,108 @@
+"use client";
```
FedericoBonel, line 1: We probably want to get rid of this route as it's hard to test from here. Adding an additional application might be better

apps/demo/package.json
```diff
@@ -3,6 +3,8 @@
   "version": "1.0.0",
   "private": true,
   "scripts": {
+    "prebuild": "yarn workspace @puckeditor/core build",
+    "predev": "yarn workspace @puckeditor/core build",
```
FedericoBonel: We shouldn't do this as it will break the demo.

apps/docs/pages/docs/api-reference/components/puck.mdx
```diff
@@ -45,6 +45,7 @@ export function Editor() {
 | [`permissions`](#permissions)                                     | `permissions: {}`                                  | [Permissions\[\]](/docs/api-reference/permissions#supported-permissions) | -            |
 | [`plugins`](#plugins)                                             | `plugins: [myPlugin]`                              | [Plugin\[\]](/docs/api-reference/plugin)                                 | Experimental |
 | [`ui`](#ui)                                                       | `ui: {leftSideBarVisible: false}`                  | [AppState.ui](/docs/api-reference/data-model/app-state#ui)               | -            |
+| [`uiStyles`](#uistyles)                                           | `uiStyles: false`                                  | `false \| { bundle?: "default" \| "no-external" }`                       | -            |
```
FedericoBonel: `uiStyles` has been dropped, so we should remove this section

apps/docs/pages/docs/api-reference/theming.mdx
```diff
@@ -20,10 +20,11 @@ CSS properties for theming the default Puck user interface.
 
 The font family used for the Puck interface. Must be used with the `no-external` bundle that stops Puck from loading the default font.
 
-```css
-/* Load bundle without existing font */
-@import "@puckeditor/core/no-external.css";
+```tsx
+<Puck uiStyles={{ bundle: "no-external" }} /* ... */ />
+```
```
FedericoBonel: This has been dropped, so we should remove this

apps/docs/pages/docs/api-reference/theming.mdx
```diff
@@ -33,10 +34,11 @@ The font family used for the Puck interface. Must be used with the `no-external`
 
 The font family used for monospaced elements of the Puck interface.
 
-```css
-/* Monospaced fonts don't use external files, so the default bundle is safe */
-@import "@puckeditor/core/puck.css";
+```tsx
+<Puck /* ... */ />
+```
```
FedericoBonel: This is not right with the current APIs, I think it's a left over of `uiStyles`

apps/docs/pages/docs/extending-puck/theming/fonts.mdx
```diff
@@ -6,21 +6,21 @@ It's possible to change the font, or provide your own version of Inter if you ne
...
 ## Load your own font file
 
-To load your own font file, use the `no-external.css` bundle instead of the primary one.
+To load your own font file, switch Puck to the `no-external` runtime bundle:
 
-```css
-/* @import "@puckeditor/core/puck.css"; */
-@import "@puckeditor/core/no-external.css";
+```tsx
+<Puck uiStyles={{ bundle: "no-external" }} /* ... */ />
```
FedericoBonel: This has been dropped so we should rollback

apps/docs/pages/docs/extending-puck/theming/fonts.mdx
```diff
@@ -6,21 +6,21 @@ It's possible to change the font, or provide your own version of Inter if you ne
...
 ## Changing the font family
 
 To change the font family Puck uses, change the [`--puck-font-family`](/docs/api-reference/theming#--puck-font-family) CSS property:
 
-```css
-/* @import "@puckeditor/core/puck.css"; */
-@import "@puckeditor/core/no-external.css";
+```tsx
+<Puck uiStyles={{ bundle: "no-external" }} /* ... */ />
+```
```
FedericoBonel: This has been dropped so we should rollback

apps/docs/pages/docs/getting-started.mdx
```diff
@@ -48,6 +47,8 @@ export function Editor() {
 }
 ```
 
+Puck injects the editor UI styles at runtime by default. Set `uiStyles={false}` if you prefer to load them manually.
+
```
FedericoBonel: I don't think we need this here, and it's also using the dropped API

```suggestion
```

packages/core/bundle/iframe-styles.ts
```diff
@@ -0,0 +1,13 @@
+import "../styles/color.css";
```
FedericoBonel: This includes the whole editor styles, so maybe we can name it `editor-styles` instead of `iframe-styles`

FedericoBonel (changes requested):

FedericoBonel:
Left an initial round of reviews, not super familiar with the style syncing code so I skimmed through and tested the functionality with inline styles, iframe overrides, tailwind and mui to test both CSS and JS based styling, code might need a clean up since I found some leftover lines here and there that we're not using.

On top of what I mentioned above, I also realized we don't have a way to stop Inter from being imported if using CSS injection, that might be fine, but we need to document it.

packages/core/bundle/iframe-styles.ts
chrisvxd: As discussed, this only includes iframe styles so we'll keep the name as-is.

apps/demo/app/runtime-css-debug/client.tsx
chrisvxd: It now works after running `yarn build` in the core package. I updated the TS paths so this no longer breaks the main demo.

FedericoBonel (approved):

PR #1658, opened 2026-05-05
perf: improve AutoFrame style mirroring performance and fix deduplication

### LINKED ISSUE #1659: AutoFrame style mirroring causes main thread freeze when many style nodes are present
## Description

When using the Puck editor with `iframe` enabled (the default), dragging a component into the canvas causes a significant main thread freeze (30–45 seconds in severe cases) before the component appears. The freeze scales with the number and size of `<style>` tags present in the parent document.

The root cause is in `AutoFrame`'s `CopyHostStyles` component, which uses `object-hash` to hash the full `outerHTML` of every style node for deduplication. This is O(n) on the CSS string length and is called once per node on initial load, and again on every `MutationObserver` callback. With a large number of sizeable style nodes, this blocks the main thread entirely.

A secondary issue is that the `MutationObserver` fires one independent deferred task per added/removed node, so a burst of style injections queues many separate main-thread tasks instead of being handled in a single pass.

### Our specific case

In our use case, we are loading Webpack federated modules (micro-frontends) as Puck blocks. The fed. mods. are styled via PandaCSS.

Puck is using an iframe to render the drop zones/slots in the editor by default. Since iframes can't inherit any styles from the parent window Puck uses a MutationObserver to mirror every style tag from the parent into the iframe. This in combination with all the duplicated styles created by Panda CSS from our fedmods results in a ~30-45s long task blocking the main thread.

## Environment

- Puck version: 0.21.2
- Browser version: Any
- Additional environment info: Reproducible in any setup where a large number of `<style>` tags are present in the parent document, e.g. when using atomic CSS frameworks or CSS-in-JS libraries

## Steps to reproduce

1. Set up a Puck editor with `iframe` enabled (default):

```tsx
<Puck config={config} data={data} />
```

2. Ensure a large number of `<style>` tags are present in the parent document (e.g. by using an atomic CSS framework or a CSS-in-JS library that injects many style nodes).
3. Open the editor and drag any component from the component list into the canvas.

## What happens

The browser main thread is blocked for an extended period (seconds to tens of seconds depending on the number and size of style nodes) before the component appears in the canvas.

## What I expect to happen

Dragging a component into the canvas should be responsive regardless of the number or size of style nodes present in the parent document.

richardszegh:
Closes #1659

## Disclaimer

Since I am pretty new to Puck's source code, I used a fair bit of AI help with the code. Any constructive feedback is welcome.

## Our problem

In our use case, we are loading Webpack federated modules (micro-frontends) as Puck blocks. The fed. mods. are styled via PandaCSS.

Puck is using an iframe to render the drop zones/slots in the editor by default. Since iframes can't inherit any styles from the parent window Puck uses a MutationObserver to mirror every style tag from the parent into the iframe. This in combination with all the duplicated styles created by Panda CSS from our fedmods results in a ~30-45s long task blocking the main thread.

With these changes, we are no longer experiencing the issue. Drag'n'dropping blocks is now responsive, and does not block the main thread.

## Description

`AutoFrame`'s `CopyHostStyles` component was using `object-hash` to hash the full `outerHTML` of each style node for deduplication. This is O(n) on the CSS string length and was causing significant main thread blocking when a large number of style nodes were present in the parent document.

Additionally, the `MutationObserver` callback was firing one independent deferred task per added/removed node, meaning a burst of style injections would queue many separate tasks rather than being handled in a single pass. There was also a bug where `removeEl` deleted the hash entry but never spliced the `elements` array, causing stale entries to accumulate indefinitely.

## Changes made

- Replaced `object-hash` with an inline djb2 hash function. `<link>` elements are keyed by `href`, `<style>` elements by their `innerHTML` hash plus serialised attributes. This avoids full node serialisation while maintaining correctness for cases where two `<style>` tags share identical CSS but differ in attributes 
- Deduplication now happens before `mirrorEl()` is called, so duplicate nodes are rejected without any DOM cloning.
- `MutationObserver` callbacks are batched: added/removed nodes are collected into pending arrays and flushed in a single deferred pass, preventing bursts of style injections from queuing one main-thread task per node.
- Style nodes are also deduplicated during initial collection, not just during subsequent mutations.
- Fixed `removeEl` to splice the `elements` array after removal, preventing stale entries that could cause `lookupEl` to return incorrect results for re-added nodes and unbounded array growth over the lifetime of the observer.
- Removed `object-hash` as a dependency.

## How to test

- Open the editor with `iframe` enabled (default).
- Add a component that injects a large number of `<style>` tags into the parent document (e.g. via a CSS-in-JS library or atomic CSS framework).
- Drag the component into the canvas and confirm it is added without any noticeable delay or main thread freeze.
- Confirm styles are correctly mirrored into the iframe and the component renders as expected.

## Summary by CodeRabbit

* **Performance**
  * Improved style mirroring and deduplication for faster, more efficient rendering.
  * Reduced unnecessary processing when handling dynamic page updates.
* **Bug Fixes**
  * Improved cleanup of outdated mirrored elements.
  * Prevented pending update operations from running after cleanup.

### DIFF
--- a/packages/core/components/AutoFrame/index.tsx
+++ b/packages/core/components/AutoFrame/index.tsx
@@ -93,7 +93,9 @@ const defer = (fn: () => void) => setTimeout(fn, 0);
 /**
  * Compute a cheap hash key for a style element.
  * - For <link> elements we key on the href (a short string).
- * - For <style> elements we key on the CSS text content.
+ * - For <style> elements we key on the CSS text content plus any attributes
+ *   (e.g. nonce, data-*) so that two tags with identical CSS but different
+ *   attributes are not incorrectly treated as duplicates.
  * We deliberately avoid hashing outerHTML because it includes the full
  * serialised CSS blob and was the primary source of the 30-45 s freeze.
  */
@@ -101,7 +103,10 @@ const styleElHash = (el: HTMLElement): string => {
   if (el.nodeName === "LINK") {
     return `link:${(el as HTMLLinkElement).href}`;
   }
-  return `style:${fastHash(el.innerHTML)}`;
+  const attrs = Array.from(el.attributes)
+    .map((a) => `${a.name}=${a.value}`)
+    .join(",");
+  return `style:${fastHash(el.innerHTML)}:${attrs}`;
 };
 
 const CopyHostStyles = ({
--- a/packages/core/components/AutoFrame/index.tsx
+++ b/packages/core/components/AutoFrame/index.tsx
@@ -246,6 +246,9 @@ const CopyHostStyles = ({
       elements[index]?.mirror?.remove();
       delete hashes[elHash];
 
+      // Must splice so lookupEl doesn't return stale entries for nodes that
+      // were removed and later re-added (e.g. on remount), and to avoid
+      // unbounded growth of the array over the lifetime of the observer.
       elements.splice(index, 1);
 
       if (debug) console.log(`Removed style node ${el.outerHTML}`);
--- a/packages/core/components/AutoFrame/index.tsx
+++ b/packages/core/components/AutoFrame/index.tsx
@@ -31,7 +31,7 @@ export const shouldMirrorStyleElement = (style: HTMLElement) => {
  * Fast, non-cryptographic djb2 hash over a string.
  * Replaces `object-hash` which serialises the full outerHTML via JSON
  * and is O(n) on the CSS string length — catastrophically slow when
- * PandaCSS / federated modules inject hundreds of KB of atomic CSS.
+ * many large style nodes are present.
  */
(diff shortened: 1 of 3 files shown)


FedericoBonel:
Hey @richardszegh! Thanks for your contribution and sorry for the delay in getting back to you.

Since you opened this we actually added some changes to the CSS loading. Would you mind rebasing and fixing any conflicts before I give this a thorough code review?

Thank you once again! 🙏

FedericoBonel (changes requested):
Please check the comment above

richardszegh:
@FedericoBonel I don't know if the bug still happens with the latest Puck version given that CSS processing has been updated since. I also don't have much time to look into it now and update the PR. I'll close this PR for now.

PR #1664, opened 2026-05-13
fix: InlineTextField renders empty for nullish values

### LINKED ISSUE #1639: InlineTextField renders literal "null" when value is null/undefined
### Version
`@puckeditor/core` 0.21.2

### Summary
Any `text` or `textarea` field with `contentEditable: true` renders the string literal `"null"` in the editor canvas when its value is `null` (or any nullish non-Node value).

### Reproduction
Minimal config:

```tsx
const config = {
  components: {
    MyBlock: {
      fields: {
        subtitle: {
          type: 'textarea',
          contentEditable: true,
        },
      },
      defaultProps: { subtitle: null },
      render: ({ subtitle }) => <p>{subtitle}</p>,
    },
  },
};
```

Insert `MyBlock` in the editor. The canvas shows the text `null` where the subtitle would be. The rendered output from `render()` is fine; the bug is only in the inline editable overlay.

### Expected
Empty text (just like an empty string).

### Actual
Literal `"null"` displayed in the inline editor. On focus, the caret is placed after the four characters `n-u-l-l` and typing appends to them, so the user-visible bug is persistent until the user manually selects-all and deletes.

### Root cause
`packages/core/components/InlineTextField/index.tsx` (compiled location: `dist/chunk-PK2F2YZX.mjs:4342`):

```js
if (ref.current) {
  if (value !== ref.current.innerText) {
    ref.current.replaceChildren(value);   // <-- bug
  }
```

`Node.replaceChildren(...nodes)` converts each non-Node argument to a text node via `String(...)`. `String(null) === "null"` and `String(undefined) === "undefined"`, so a nullish `value` is materialized as a literal text node.

### Suggested fix
```js
const safeValue = value == null ? "" : value;
if (safeValue !== ref.current.innerText) {
  ref.current.replaceChildren(safeValue);
}
```

### Why this matters
- Every `defaultProps` that sets an optional text/textarea field to `null` (a reasonable "no value" choice for `string | null` types) triggers this.
- Any server-returned data that stores `null` for an unfilled optional field shows "null" on load.
- The `richtext` field type (`InlineEditorWrapper` → TipTap) is unaffected because TipTap handles nullish content. The bug is specific to `InlineTextField`.

Same three lines need the fix across the three build outputs (`dist/chunk-PK2F2YZX.mjs`, `dist/index.js`, `dist/no-external.js`). Applied locally as a `pnpm patch` while waiting for upstream.

shiminshen:
Closes #1639

## Description

`InlineTextField` was rendering the literal text `"null"` (or `"undefined"`) in the contentEditable canvas whenever a `text` or `textarea` field with `contentEditable: true` received a nullish value. The cause is `Node.replaceChildren(...nodes)` coercing each non-Node argument via `String(...)` — `String(null) === "null"` and `String(undefined) === "undefined"` — so a defaultProp of `subtitle: null`, or a server response returning `null` for an unfilled optional field, persistently surfaced as `null` text in the editor until the user manually selected-all and deleted.

The `richtext` field type is unaffected because TipTap handles nullish content; this is specific to `InlineTextField` only.

## Changes made

- Coerce `value ?? ""` before calling `replaceChildren`, in the same effect that syncs the contentEditable span's text.
- Widen the `value` prop type from `string` to `string | null | undefined` so the type matches what the `text` / `textarea` field transforms actually pass in (the data model allows nullish props).
- Add jest tests covering `value={null}`, `value={undefined}`, and `value="Hello world"`.

## How to test

Minimal repro from the issue:

```tsx
const config = {
  components: {
    MyBlock: {
      fields: {
        subtitle: { type: 'textarea', contentEditable: true },
      },
      defaultProps: { subtitle: null },
      render: ({ subtitle }) => <p>{subtitle}</p>,
    },
  },
};
```

Insert `MyBlock` in the editor. Before: the canvas shows the text `null` where the subtitle would be. After: empty, like an empty string.

Or run `yarn test --testPathPattern=InlineTextField` from `packages/core` — 3 new tests cover the null / undefined / string cases.

### DIFF
--- a/packages/core/components/InlineTextField/__tests__/index.spec.tsx
+++ b/packages/core/components/InlineTextField/__tests__/index.spec.tsx
@@ -1,0 +1,76 @@
+import { render } from "@testing-library/react";
+import "@testing-library/jest-dom";
+import { InlineTextField } from "../index";
+
+// useAppStoreApi reads from a Zustand-style API; we only need a getState() that
+// returns the slice InlineTextField touches in its effect.
+const makeAppStore = () => ({
+  getState: () => ({
+    state: {
+      indexes: {
+        nodes: {
+          "comp-1": {
+            data: { type: "MyBlock", props: { id: "comp-1" } },
+          },
+        },
+      },
+    },
+    getComponentConfig: () => ({ fields: {} }),
+  }),
+});
+
+jest.mock("../../../store", () => ({
+  useAppStoreApi: () => makeAppStore(),
+}));
+
+jest.mock("../../../lib/overlay-portal", () => ({
+  registerOverlayPortal: () => () => undefined,
+}));
+
+jest.mock("../styles.module.css");
+
+describe("InlineTextField", () => {
+  it("renders empty when value is null instead of the literal string 'null'", () => {
+    const { container } = render(
+      <InlineTextField
+        propPath="subtitle"
+        componentId="comp-1"
+        value={null}
+        isReadOnly={false}
+      />
+    );
+
+    const span = container.querySelector("span");
+    expect(span).not.toBeNull();
+    expect(span?.innerText ?? span?.textContent ?? "").toBe("");
+  });
+
+  it("renders empty when value is undefined instead of the literal string 'undefined'", () => {
+    const { container } = render(
+      <InlineTextField
+        propPath="subtitle"
+        componentId="comp-1"
+        value={undefined}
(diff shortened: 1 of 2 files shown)


FedericoBonel (approved):

FedericoBonel:
Thanks for the contribution!

chrisvxd (approved):

PR #1670, opened 2026-05-18
Fix depandabot vulnerabilities

FedericoBonel:
## Description

This PR consolidates and resolves all open Dependabot security alerts by upgrading direct and transitive dependencies across the monorepo. It also fixes a CSS module resolution issue introduced by the dependency upgrades.

After this PR, `yarn audit` reports **0 actionable vulnerabilities** (the only 2 remaining are `postcss@8.4.31` pinned inside Next.js; fix is merged in Next.js canary but not yet in a stable release).

## Changes made

- Upgraded dependencies that had vulnerabilities (both direct and transitive)
- Lifted and added UUID v4 generation to the repo to mitigate an upstream vulnerability
- Upgraded eslint and migrated to flat config


apps/demo/tsconfig.json
```diff
@@ -4,9 +4,9 @@
     "strict": true,
     "plugins": [{ "name": "next" }],
     "paths": {
-      "@/core": ["../../packages/core"],
+      "@/core": ["../../packages/core/index.ts"],
       "@/core/*": ["../../packages/core/*"],
-      "@puckeditor/core": ["../../packages/core"],
+      "@puckeditor/core": ["../../packages/core/index.ts"],
```
chrisvxd, line 9: This will conflict with #590, so we just need to align the ordering

apps/docs/tsconfig.json
```diff
@@ -3,9 +3,10 @@
   "compilerOptions": {
     "plugins": [{ "name": "next" }],
     "paths": {
-      "@/puck": ["../../packages/core"],
-      "@/core": ["../../packages/core"],
+      "@/puck": ["../../packages/core/index.ts"],
+      "@/core": ["../../packages/core/index.ts"],
       "@/core/*": ["../../packages/core/*"],
+      "@puckeditor/core": ["../../packages/core/index.ts"],
```
chrisvxd: It's possible you might be better off just setting `@puckeditor/core` as a dependency for docs

eslint.config.mjs
```diff
@@ -0,0 +1,19 @@
+import custom from "eslint-config-custom";
+
+const config = [
+  // eslint-config-next causes warning on Remix's default remix.config.js
+  { ignores: ["recipes/remix/remix.config.js"] },
```
chrisvxd: Is this still necessary?

chrisvxd (changes requested):

eslint.config.mjs
FedericoBonel: No! Good catch did this before opening the other PR

chrisvxd (approved):

PR #1672, opened 2026-05-19
fix: IconButton exposes accessible name via aria-label

### LINKED ISSUE #1669: a11y: Viewport-switcher buttons rely on visually-hidden span for accessible name
# a11y: Viewport-switcher buttons rely on visually-hidden span for accessible name

## Summary

The viewport switcher inside `<Puck />`'s preview chrome (Mobile / Tablet / Desktop buttons, plus Zoom in / Zoom out) currently provides its accessible name via a child `<span>` (presumably visually hidden via `sr-only` CSS) rather than an explicit `aria-label` on the `<button>`. The buttons ARE reachable by screen readers, but the pattern is fragile and inconsistent with the rest of Puck's icon-only buttons (which use `aria-label` directly, e.g. the chevrons inside drawer-item).

## Steps to reproduce

1. Mount `<Puck />` with any config.
2. Inspect the viewport-switcher buttons in the lower-left chrome.
3. Note the button structure:

```html
<button class="…">
  <span>Switch to Mobile viewport</span>  <!-- presumably sr-only -->
  <svg>…</svg>
</button>
```

vs. the more common Puck pattern elsewhere:

```html
<button aria-label="Drag" class="…">
  <svg>…</svg>
</button>
```

## Expected

The icon-only `<button>` should expose its label via `aria-label` for consistency and resilience to CSS regressions (a one-character change to the `sr-only` rule risks visually exposing what should be hidden).

## Actual

```js
document.querySelectorAll('button').forEach(b => {
  if (b.textContent?.includes('viewport')) {
    console.log(b.textContent.trim(), '| aria-label=', b.getAttribute('aria-label'));
  }
});
// → "Switch to Mobile viewport | aria-label= null"
// → "Switch to Tablet viewport | aria-label= null"
// → "Switch to Desktop viewport | aria-label= null"
// → "Zoom viewport out | aria-label= null"
// → "Zoom viewport in | aria-label= null"
```

## Why this matters

Two reasons:

1. **Consistency.** Puck uses `aria-label` for other icon-only buttons. Mixing patterns makes the codebase harder to maintain and audit.
2. **Robustness.** `sr-only`-based names break silently if the CSS class is renamed, removed, or overridden by host styles. `aria-label` is a single source of truth that can't be visually broken.

This is a polish issue, not a critical a11y bug — current users with screen readers DO get the labels. But the buttons are user-facing chrome where this kind of inconsistency is easy to spot and easy to fix.

## Suggested fix

In the viewport-switcher render, replace the sr-only `<span>` with `aria-label` on the button:

```diff
- <button className="…">
-   <span className="sr-only">Switch to Mobile viewport</span>
-   <svg>…</svg>
- </button>
+ <button className="…" aria-label="Switch to Mobile viewport">
+   <svg>…</svg>
+ </button>
```

## Environment

- `@puckeditor/core@0.21.2`
- React 19.2 / Next.js 16
- Confirmed in Chrome 132 on macOS

## Related

A sibling issue covers `aria-expanded` on the drawer category-header buttons (#1668).

shiminshen:
Closes #1669

## Description

The icon-only buttons rendered by `IconButton` (used in `ViewportControls` and `ExternalInput`) currently expose their accessible name via a child `<span>` that's CSS-clipped (`.IconButton-title`) to act as sr-only. The `<button>` itself has no `aria-label`.

As #1669 describes, this pattern is fragile — a regression to the `.IconButton-title` CSS would visually expose what should be hidden — and inconsistent with the `aria-label` pattern used elsewhere in puck.

This PR sets `aria-label` on the button using the existing required `title` prop, drops the now-redundant clipped child span, and removes the `.IconButton-title` CSS rule.

## Changes made

- `IconButton.tsx` — set `aria-label={title}` on the rendered `<button>` / `<a>`; remove the child `<span className="IconButton-title">{title}</span>`.
- `IconButton.module.css` — remove the now-unused `.IconButton-title` rule.
- `IconButton/__tests__/index.spec.tsx` — new test file asserting the rendered button has `aria-label` set and that icon children render unchanged.

Both call sites (`ViewportControls`, `ExternalInput`) keep their existing `title="..."` prop — no API change.

## How to test

Automated:
- \`yarn workspace @puckeditor/core test --testPathPattern='IconButton'\` — both new tests pass.
- The full core suite shows the same pre-existing failure as \`main\` (\`components/Puck/__tests__/index.tsx\` — fallout from the recent dnd-kit@0.4.0 upgrade, unrelated to this change).

Manual:
1. Run the dev server with any Puck config.
2. Inspect the viewport-switcher buttons (Mobile / Tablet / Desktop, Zoom in / out).
3. Each \`<button>\` should now have an \`aria-label\` matching its \`title\`:

\`\`\`js
document.querySelectorAll('button').forEach(b => {
  const t = b.getAttribute('title');
  if (t?.includes('viewport')) {
    console.log(t, '| aria-label =', b.getAttribute('aria-label'));
  }
});
\`\`\`


packages/core/components/IconButton/__tests__/index.spec.tsx
```diff
@@ -0,0 +1,32 @@
+import { render } from "@testing-library/react";
```
FedericoBonel: I'm not opposed to this, but if we add a test for this, we should then add tests for every prop, so I think we can leave this test out for this one.

I love tests, but I think testing for the presence of an HTML attributes can quickly get out of hand.

FedericoBonel (changes requested):
Please check the comment above. 

I think we can safely remove the test for this.

packages/core/components/IconButton/__tests__/index.spec.tsx
shiminshen: Good point — dropped the test file in 11fd2a74. Agreed that asserting the presence of individual HTML attributes would invite doing the same for every prop, so I've left it out and kept this PR scoped to the accessibility fix. Thanks for the review!

shiminshen:
@FedericoBonel removed the test 👍 ready for another look whenever you have a sec

FedericoBonel (approved):

PR #1675, opened 2026-05-26
fix: don't crash richtext field when string is empty

chrisvxd:
## Description

Fixes a crash in RichTextRender when a richtext field contains an empty string, impacting AI generation.

ProseMirror rejects empty text nodes, so programmatically setting a richtext field to `""` (or initializing it as empty) would throw `RangeError: Empty text nodes are not allowed`, preventing the component from rendering and blocking subsequent updates.

## Changes made

- Added an early return in `RichTextRender` for empty strings, returning an empty doc (`{ type: "doc", content: [] }`) instead of an invalid text node.
- Added focused test coverage for programmatic richtext field updates via `dispatch`, including the empty string edge case and plain-text transition.
- Removed the broader multi-field dispatch test based on review feedback to keep this suite scoped to richtext behavior.

## How to test

Create a component with a richtext field initialized to `""`, then programmatically update it via dispatch. The updated content should render correctly:

```ts
// Initial data with empty richtext
const data = {
  content: [{ type: "Block", props: { id: "1", body: "" } }],
};

// Later, dispatch a replace to set the value
dispatch({
  type: "replace",
  data: { type: "Block", props: { id: "1", body: "<p>Hello</p>" } },
  destinationIndex: 0,
  destinationZone: "default-zone",
});
```

### DIFF
--- a/packages/core/components/Puck/__tests__/richtext.spec.tsx
+++ b/packages/core/components/Puck/__tests__/richtext.spec.tsx
@@ -185,66 +185,6 @@ describe("Puck - richtext programmatic updates", () => {
     expect(screen.queryByText("Hello world")).not.toBeInTheDocument();
   });
 
-  it("updates richtext when only the richtext field changes on a component with multiple fields", async () => {
-    const config: Config = {
-      components: {
-        Card: {
-          fields: {
-            title: { type: "text" },
-            body: { type: "richtext", contentEditable: false },
-          },
-          render: ({ title, body }) => (
-            <div>
-              <h2>{title}</h2>
-              <div data-testid="body">{body}</div>
-            </div>
-          ),
-        },
-      },
-    };
-
-    const data: Data = {
-      root: { props: {} },
-      content: [
-        {
-          type: "Card",
-          props: {
-            id: "card-1",
-            title: "My Card",
-            body: "<p>Original body</p>",
-          },
-        },
-      ],
-    };
-
-    render(<Puck config={config} data={data} iframe={{ enabled: false }} />);
-    await flush();
-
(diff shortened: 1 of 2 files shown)


packages/core/components/Puck/__tests__/richtext.spec.tsx
```diff
@@ -0,0 +1,331 @@
...
+    });
+    await flush();
+
+    await waitFor(() => {
+      expect(screen.getByText("Updated content")).toBeInTheDocument();
+    });
+    expect(screen.queryByText("Hello world")).not.toBeInTheDocument();
+  });
+
+  it("updates richtext when only the richtext field changes on a component with multiple fields", async () => {
```
FedericoBonel: This seems more of a dispatch test than a rich text one, also is this one needed?

packages/core/components/Puck/__tests__/richtext.spec.tsx
```diff
@@ -0,0 +1,331 @@
...
+
+type PuckInternal = {
+  appStore: AppStoreApi;
+};
+
+const getInternal = () => {
+  return (window as any).__PUCK_INTERNAL_DO_NOT_USE as PuckInternal;
+};
+
+const flush = () => act(async () => {});
```
FedericoBonel, line 72: Not now, but we should potentially pull these setups into a shared test utils

FedericoBonel (approved):

packages/core/components/Puck/__tests__/richtext.spec.tsx
chrisvxd: @copilot remove this

PR #1690, opened 2026-06-06
fix(core): track overlay on inner-container scroll (#913)

### LINKED ISSUE #913: DropZones That Overflow in the X Axis Provide Buggy Visual Feedback When Scrolling
## Description  
When a `DropZone` overflows in the horizontal (X) axis, the visual feedback for drag-and-drop becomes buggy when scrolling. The preview of each item’s drag zone remains fixed in its original position instead of updating with the scroll.  

## Steps to Reproduce  

1. Configure a `DropZone` and an item as follows:  

```jsx
   Row: {
     render: () => {
       return (
         <DropZone zone="row" style={{ display: "flex", overflow: "auto" }} />
       );
     },
   },
   Item: {
     inline: true,
     render: ({ puck }) => {
       return (
         <div
           style={{
             padding: "16px",
             width: 500,
             height: 200,
             border: "1px solid #ccc",
             flexShrink: 0,
           }}
           ref={puck.dragRef}
         >
           Some item
         </div>
       );
     },
   },
```

2. Drag and drop a Row component into the editor.
3. Add multiple Item components inside until they overflow horizontally.
4. Scroll within the overflown row (Shift + Scroll Down).

## Expected Behavior
The drag zone preview should update its position correctly when scrolling, staying aligned with the items.

## Actual Behavior
The preview of each item’s drag zone is misplaced—it remains fixed in the position the item was in before scrolling.

## Video

https://github.com/user-attachments/assets/9be5384b-e644-47b1-89c9-0791973e4f07

aalimovs:
Not gonna lie, thank you Claude.

However, this was extensively manually reviewed and tested, for both iframe and non-iframe setup.

Happy to clean up comments if this is approved.

Closes #913.

## Description

Fixes issue with horizontal scrolling when a component is selected or hovered, exactly like on video in #913.

The overlay is portaled outside the scrolling container, but `getStyle()` was compensating with `getDeepScrollPosition(el)`, which summed the scroll of every ancestor, including the inner `DropZone`. That extra amount cancelled out the item's movement, so the overlay froze.

Now it only accounts for the scroll of the element the overlay is actually attached to:
  - **non-iframe:** the `[data-puck-preview]` container's own scroll
  - **iframe / body portal:** the document's own scroll

## Changes made

  - `getStyle()` compensates for only the portal target's own scroll instead of summing every ancestor.
  - Hovered components (not just selected/dragging) now re-sync on scroll, so hover overlays track too — without starting the per-frame rAF loop on hover.
  - Removed the now-unused `get-deep-scroll-position` helper (internal-only).

## How to test

See #913 video.


chrisvxd:
@shannonhochkins would you mind checking this against your Home Assistant use-case?

PR #1700, opened 2026-06-15
chore: add aria-expanded to category buttons

### LINKED ISSUE #1668: a11y: Add Section drawer category-disclosure buttons lack `aria-expanded`
# a11y: Add Section drawer category-disclosure buttons lack `aria-expanded`

## Summary

The "Add Section" drawer in `<Puck />` renders each `categories[].title` as a clickable disclosure button that expands/collapses its component list. These buttons do not expose `aria-expanded`, so screen readers can't tell whether a category is currently open or closed.

## Steps to reproduce

1. Mount `<Puck />` with a config containing 2+ `categories`, e.g.:

```tsx
<Puck
  config={{
    categories: {
      navigation: { title: "Navigation", components: ["SiteHeader"] },
      content: { title: "Content", components: ["Hero", "About"], defaultExpanded: true },
    },
    components: { /* … */ },
    root: { /* … */ },
  }}
  data={{ content: [], root: {} }}
  onPublish={() => {}}
/>
```

2. In DevTools, inspect the "Navigation" or "Content" header button in the left drawer.
3. Click the button to collapse/expand. The visual chevron rotates; the DOM `aria-expanded` attribute does not exist before or after the click.

## Expected

The category-header `<button>` should expose `aria-expanded="true"` when its component list is rendered and `aria-expanded="false"` when collapsed.

## Actual

The button has no `aria-expanded`. JS query:

```js
document.querySelectorAll('button').forEach(b => {
  const t = b.textContent?.trim();
  if (["Navigation","Content","Site Structure"].includes(t)) {
    console.log(t, 'aria-expanded=', b.getAttribute('aria-expanded'));
  }
});
// → "Navigation aria-expanded= null"
```

## Why this matters

* **Screen-reader users** rely on `aria-expanded` to know whether activating the button will expand or collapse a region. Without it, the only feedback is the visual chevron rotation, which is invisible to AT users.
* Disclosure widgets are one of the most common WAI-ARIA patterns and `aria-expanded` is the load-bearing attribute (see [ARIA Authoring Practices — Disclosure](<https://www.w3.org/WAI/ARIA/apg/patterns/disclosure/>)).
* The buttons already act semantically as disclosures (clicking opens/closes the list); adding the attribute is a one-line change in Puck's drawer-category render.

## Suggested fix

Add `aria-expanded={isExpanded}` (and likely `aria-controls={…}` pointing at the list region) to the `<button>` element that renders each category header inside Puck's `Drawer` / `SidebarSection` component.

## Environment

* `@puckeditor/core@0.21.2`
* React 19.2 / Next.js 16
* Confirmed in Chrome 132 on macOS

## Related

A sibling issue covers `aria-label` on the viewport-switcher buttons (puckeditor/puck#1669).

FedericoBonel:
Closes #1668

This PR adds aria-expanded to category buttons for accesibility.

## Changes made

- The Component list now sends down the expanded flag for each category in the `aria-expanded` attribute.

### DIFF
--- a/packages/core/components/ComponentList/index.tsx
+++ b/packages/core/components/ComponentList/index.tsx
@@ -54,12 +54,16 @@ const ComponentList = ({
 
   const { expanded = true } = componentList[id] || {};
 
+  const contentId = `puck-drawer-category-${id}`;
+
   return (
     <div className={getClassName({ isExpanded: expanded })}>
       {title && (
         <button
           type="button"
           className={getClassName("title")}
+          aria-expanded={expanded}
+          aria-controls={contentId}
           onClick={() =>
             setUi({
               componentList: {
@@ -83,7 +87,7 @@ const ComponentList = ({
           </div>
         </button>
       )}
-      <div className={getClassName("content")}>
+      <div className={getClassName("content")} id={contentId}>
         <Drawer>
           {children ||
             Object.keys(config.components).map((componentKey) => {


chrisvxd (commented):
Any reason you didn't want to implement the second part of the ticket suggestion?

> and likely aria-controls={…} pointing at the list region

FedericoBonel:
Went ahead and added it!

xaviemirmon (approved):
LGTM

PR #1704, opened 2026-06-17
fix(core): guard document.body in NestedDroppablePlugin teardown

leegunwoo98:
## Summary

`NestedDroppablePlugin` adds a `pointermove` listener to `document.body` and removes it in its effect cleanup. When the `<Puck>` tree unmounts, `document.body` can be `null` at teardown, so the cleanup throws:

```
Uncaught TypeError: Cannot read properties of null (reading 'removeEventListener')
    at cleanup (NestedDroppablePlugin)
    at NestedDroppablePlugin.destroy
    at PluginRegistry.destroy
    at DragDropRegistry.destroy
```

Because `manager.destroy()` runs **synchronously inside `DragDropProvider`'s `useEffect` cleanup**, the throw escapes React's unmount commit — so it doesn't just log, it can leave the editor broken on the next mount.

## When it happens

Any path that unmounts the Puck tree, e.g.:

- **React StrictMode** (dev) double-invokes effects (mount → unmount → mount), so it fires on the very first editor mount.
- A **keyed remount** of the editor (`<Puck key={...} />`), e.g. switching pages/designs.
- **Route navigation** away from the editor.

## Fix

Guard the listener add/remove with optional chaining so the cleanup tolerates a torn-down DOM:

```diff
-document.body.addEventListener("pointermove", handlePointerMove, { capture: true });
+document.body?.addEventListener("pointermove", handlePointerMove, { capture: true });
 ...
-document.body.removeEventListener("pointermove", handlePointerMove, { capture: true });
+document.body?.removeEventListener("pointermove", handlePointerMove, { capture: true });
```

No behavior change when `document.body` is present; it only avoids throwing during teardown. Effect cleanups can't assume the DOM is still intact.

## Notes

- Reproducible on `main` and the current `@puckeditor/core` release (`0.21.2`).
- Happy to add a regression test (mount/unmount the plugin with `document.body` nulled at teardown) if you'd like one.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


FedericoBonel:
Hey @leegunwoo98! Thanks for the contribution and for your work.

I don't think this is a risky change, but I would really appreciate it if you could first provide a minimal reproduction through a [bug report](https://github.com/puckeditor/puck/issues/new?template=bug-report.md).

That way, we can determine whether this is indeed a problem with Puck, something in your integration that is causing the issue, or both. Minimal reproductions are extremely useful for us, but they can also help you identify exactly when the issue occurs and make it easier to debug.

Before I can approve this, could you provide a minimal reproduction of the issue? If you're unsure how to format it, check out the [GitHub bug report template](https://github.com/puckeditor/puck/issues/new?template=bug-report.md). It includes instructions, but essentially, a simple Puck setup snippet without external dependencies, along with a series of steps to reproduce the issue, is enough.

Thank you again!

FedericoBonel (changes requested):
Please check the above comment

PR #1678, opened 2026-06-01
Theming (WIP)

chrisvxd:
Closes #139

## Description

This PR adds a CSS theming API.

## TODO

- [x] Initial theming pass
- [x] Component themes
- [ ] Tokens are prefixed based on intent (i.e.`-color-`)
- [x] Audit items addressed
- [x] Add documentation

## Acceptance criteria

1. The developer can update the core theme and impact the entire UI
2. The developer can update the theme of a particular component
3. The developer can update the theme of interactive components for different states
4. When updating the theme, the developer can update color, spacing, typography, and border radius

## Claude audit of current progress

Here's the full audit of inconsistencies in the token system:

#### `--puck-color-text-strong` is weaker than `--puck-color-text`

The text scale orders like: `text` (black) > `text-strong` (grey-04) > `text-muted` (grey-05) > `text-subtle` (grey-07). "Strong" implies MORE emphasis than the base, but grey-04 is lighter than black. This is the secondary body text color — a name like `text-secondary` would be clearer. Every other modifier in the system (`-muted`, `-subtle`) correctly indicates a step DOWN from the base.

#### `--puck-color-interactive-soft` is stronger than `--puck-color-interactive-subtle`

`-soft` (azure-10, darker) sits between `-subtle` (azure-11) and the core selection colors — it's the MORE visible one. But "soft" in English implies gentler/less visible than "subtle." The naming is inverted. A natural reading would put the scale as: `interactive > soft > subtle > subtle-hover`, with soft being lighter. Consider swapping the names or using something like `-subtle-active`.

#### Surface and text modifier directions are inverted

For text: `muted` (grey-05) is MORE visible than `subtle` (grey-07). Good — muted > subtle.
For surfaces: `muted` (grey-12 / #fafafa) is LIGHTER than `subtle` (grey-11 / #f5f5f5). So `muted` is LESS visible than `subtle`. The same modifier words mean opposite relative intensities depending on whether you're looking at text or surfaces.

---

#### `--puck-color-text-muted` used as a border color (5 places)

This means overriding `--puck-color-text-muted` to change help text color also changes input focus/hover borders — a surprising side effect.

| File | Line | Usage |
|---|---|---|
| `AutoField/styles.module.css` | 66, 72, 78 | Input hover/focus `border-color` |
| `RichTextEditor/styles.module.css` | 102 | Editor focus `border-color` |
| `ExternalInput/styles.module.css` | 258, 265 | Search focus/hover `border-color` |

[...]

PR #1705, opened 2026-06-18
Theming

FedericoBonel:
## Description

This PR builds on top of #1678 and addresses its remaining TODO, the audit comments, and the acceptance criteria for the theming API.

#1678 introduced the initial CSS theming pass; this PR finishes the token-naming work, fixes the semantic inconsistencies raised in the audit, extends theming down to the individual-component level, and documents the whole API.

## Changes made

Most of the changes are styles and tokenization, but at a higher level:

**Token naming**
- Every design token now carries the `--puck-` namespace plus an intent segment (`-color-`, `-space-`, `-radius-`, …), so a token's name tells you what it controls.
- Descriptive and consistency renames: `--puck-space-px` → `--puck-space-chrome-gutter`, `--puck-side-bar-width` → `--puck-sidebar-width`.

**Fixed semantic naming contradictions**
- `--puck-color-text-strong` → `--puck-color-text-secondary` — it sits *below* the base text color, so "secondary" is accurate where "strong" implied more emphasis.
- Swapped the `interactive-subtle` / `interactive-soft` pair (and their `*-hover`) so "subtle" maps to the quieter shade, matching the rest of the scale.
- Swapped the `surface-muted` / `surface-subtle` values so the muted/subtle direction is consistent with the text tokens.
- Motion easings `ease-standard` / `ease-exit` → `ease-exit` / `ease-entrance` (ease-in = leaving, ease-out = arriving), matching motion convention.

**Fixed tokens used for the wrong semantic role**
- Added a dedicated `--puck-color-border-hover` so input hover/focus borders no longer reuse `text-muted`.
- Added `--puck-color-interactive-neutral-hover` for icon-button / select hover backgrounds instead of borrowing `border-muted`.

**Hygiene**
- Removed the unused `--puck-color-danger` token.
- Stopped polluting the consumer's global CSS namespace: every unprefixed `:root` variable (`--line-height-*`, `--fallback-font-stack`, `--space-m-unitless`, `--min-empty-height`, the `*-side-bar-width` vars) is now `--puck-`-prefixed, with all internal references updated.
- Removed the redundant component-level `--puck-space-px` override that was silently blocking user customization.

**Removed the hardcoded `15px` field padding**
- Input horizontal padding is now `calc(var(--puck-space-4) - var(--puck-border-width-regular))` (exposed as `--puck-field-space-x`), applied to text/select inputs, rich text content, and external-input search; so it stays on the spacing scale and visually matches other components.

**Per-component theming**
- Component-scoped tokens added so individual editor parts can be themed independently, following a consistent `--puck<component>-*` convention: **Action bar, Button, Canvas, Drawer, Fields, Header, Icon button, Outline, Plugin bar, Sidebar, and Slot**.

**Supporting refactors**
- `Action` and `IconButton` converted to `forwardRef` components that spread DOM props, so `Select` can reuse them as its Radix Popover trigger (`Action` in the action bar, `IconButton` elsewhere) instead of duplicating button styling; this also let the bespoke `.Select-button` styles be deleted.
- `SelectField` now renders a real `<ChevronDown>` element instead of an inline SVG background image, so the dropdown chevron follows theme tokens.

**Visual consistency fixes**
- DraggableComponent's selected border now uses the shared selection-border color instead of a one-off azure shade.
- RichTextEditor gained a hover border to match other text inputs.
- Mobile panels now animate in with the new `--puck-ease-entrance` easing.
- Viewport zoom `<select>` inherits `currentColor` rather than the browser default.

**Documentation**
- Reworked the global theming API reference (color swatches + a "Used for" column per token).
- 11 new per-component theming pages, each with a token table and a preview if possible.
- Added a computed-values reference page documenting the default resolved value for each component-level theming token.
- Expanded the theming overview (applying a theme, global vs. component tokens, going further) and added further-reading links from the fonts page.

## How to test

**Global theme (dark mode)**
- Navigate to the demo
- Open dev tools
- Select "emulate dark mode" (Rendering tab → *Emulate CSS prefers-color-scheme: dark*)
- Observe the entire editor UI switch to the dark theme.

**Component theme (optional):**
- Open dev tools.
- Select the HTML element at the top of the document.
- Open the styles tab.
- Override a component-scoped token on an ancestor element (e.g. `--puck-header-color-bg`) and confirm only that component changes, not the whole UI.

**Computed values docs**
- Open `/docs/api-reference/theming/computed-values`.
- Verify each themable component section lists default computed values for its component-level tokens.


packages/core/components/IconButton/IconButton.tsx
```diff
@@ -1,73 +1,84 @@
...
-  disabled?: boolean;
-  tabIndex?: number;
-  newTab?: boolean;
-  fullWidth?: boolean;
-  title: string;
-  suppressHydrationWarning?: boolean;
-}) => {
-  const [loading, setLoading] = useState(false);
+export const IconButton = forwardRef<
+  HTMLButtonElement & HTMLAnchorElement,
```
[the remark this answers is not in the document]
FedericoBonel: Doing that will create typescript issues when spreading props

apps/docs/components/Color/styles.module.css
```diff
@@ -0,0 +1,13 @@
...
+  gap: 6px;
+}
+
+.Color-swatch {
+  display: inline-block;
+  height: 1.2em;
+  aspect-ratio: 1 / 1;
+  border: 1px solid color-mix(in srgb, currentColor 30%, transparent);
+  border-radius: 3px;
+}
```
chrisvxd, line 13: We need to consider how this will migrate to the actual site

apps/docs/pages/docs/extending-puck/theming.mdx
```diff
@@ -1,20 +1,67 @@
-import { Callout } from "nextra/components";
-
 # Theming
```
chrisvxd, line 1: Having this entire section for 2 sub pages seems a bit overkill now. It might be worth collapsing these down into a single file

apps/docs/pages/docs/api-reference/theming/global.mdx
```diff
@@ -1,43 +1,137 @@
-import { Callout } from "nextra/components";
+import { Color } from "@/docs/components/Color";
```
chrisvxd, line 1: We generally don't use these root node docs (which is only an issue because you have child pages) because they get lost. You'd be better off having a "Global" section at the top of the Theming directory and getting rid of this file.

chrisvxd:
@copilot Can you document every change to computed value for each component based on the theming API?

packages/core/components/DropZone/styles.module.css
```diff
@@ -13,35 +11,25 @@
...
+  transition: min-height var(--puck-duration-medium) var(--puck-ease-exit);
 }
 
 .DropZone--isAreaSelected,
 .DropZone--hoveringOverArea:not(.DropZone--isRootZone) {
-  background: color-mix(in srgb, var(--puck-color-azure-09) 30%, transparent);
-  outline: 2px dashed var(--puck-color-azure-08);
+  background: var(--puck-slot-color-bg, var(--puck-color-selection-bg));
+  outline: var(--puck-border-width-focus) dashed
+    var(--puck-slot-color-border, var(--puck-color-selection-border));
```
chrisvxd: I don't mind it defaulting to them, but I would probably expect this outline to be customisable separately from the border styles, including the `dashed`

apps/docs/pages/docs/api-reference/theming/computed-values.mdx
```diff
@@ -0,0 +1,190 @@
...
+| `--puck-outline-radius`                | `4px`                  |
+| `--puck-outline-font-size`             | `0.875rem` (`14px`)    |
+| `--puck-outline-zone-color-text`       | `#767676`              |
+| `--puck-outline-zone-font-size`        | `0.75rem` (`12px`)     |
+
+## Plugin bar
+
+| Token                                  | Computed default value |
+| -------------------------------------- | ---------------------- |
+| `--puck-pluginbar-color-bg`            | `#fafafa`              |
```
chrisvxd: There's an inconsistency in the naming for some of these. `puck-pluginbar` vs `puck-icon-button`. Should be consistent here.

chrisvxd (dismissed):
Some initial comments to keep you busy! Will do another pass later.

apps/docs/components/Color/styles.module.css
FedericoBonel: Yep! I saw that we are duplicating these components, so that should be a follow up after this gets merged, moving it there.

apps/docs/pages/docs/extending-puck/theming.mdx
FedericoBonel: Yeah good idea!

packages/core/components/DropZone/styles.module.css
FedericoBonel: Ah okay nice! I initially did a bunch of tokens, but rolled them back slowly to not blow it up. Sounds good though I agree.

apps/docs/pages/docs/api-reference/theming/global.mdx
FedericoBonel: Ah okay, I did it like this because I saw we did this for the `Fields` API reference.

apps/docs/pages/docs/extending-puck/theming.mdx
FedericoBonel: Done!

apps/docs/pages/docs/api-reference/theming/global.mdx
FedericoBonel: Done! I moved all the global tokens to a dedicated page and moved the list of all tokens to the extending puck page. That way, we still get an index of sorts for people to explore. 

I also added some links from each docs page to the global vars docs so people can cross-check the defaults (`vars(...)`) if they want.

I do think it would be cool to have some way of checking the default value without navigating away. I could make a little component for it and define an object somewhere that represents the cascading graph so you can check it while hovering on them, but it might be a bit of a hassle to maintain. Happy to explore it if you want.

FedericoBonel: Also not enterily sure as to what to do with the computed values page copilot added, the global docs kind of do the same, so I'm a bit unsure about it.

packages/core/components/IconButton/IconButton.tsx
```diff
@@ -1,73 +1,84 @@
...
-          &nbsp;&nbsp;
-          <Loader size={14} />
-        </>
-      )}
-    </ElementType>
-  );
+          setLoading(true);
+          Promise.resolve(onClick(e)).then(() => {
+            setLoading(false);
+          });
```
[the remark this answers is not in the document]
FedericoBonel: This was already a problem before this PR, happy to fix it but seems a bit out of scope and can make the commit history complicated to manage

apps/docs/pages/docs/api-reference/theming/computed-values.mdx
FedericoBonel: Done!

packages/core/components/DropZone/styles.module.css
FedericoBonel: Done! 

I added two properties, one for the width (`--puck-slot-border-width`), another for the style (`--puck-slot-border-style`). 

Also normalized the enabled state border to use the same color. The border used to get a bit darker when targetting the dropzone, but I realized that the biggest feedback was the component preview instead of the border. Happy to roll the border back for the enabled state though!

FedericoBonel:
Addressed the first wave of comments!

xaviemirmon (commented):
### 1. `ViewportControls` tokens are off-convention on every axis
`packages/core/components/ViewportControls/styles.module.css:2-6`
- No `-color-` intent segment — `--puck-viewports-control-toggle-bg` vs the norm `...-color-bg`
- State in the middle — `...-toggle-hover-text` vs the norm `...-color-text-hover`
- Hyphenated, plural prefix `viewports-control-toggle` vs the non-hyphenated compounds elsewhere (`actionbar`, `iconbutton`, `pluginbar`)
- Declared with raw-palette defaults (`grey-02`, `azure-07`, …) instead of semantic tokens
- Public-looking (no `--_puck-` prefix) yet undocumented — appears on no theming page

Tokens: `--puck-viewports-control-toggle-bg | -border | -text | -hover-text | -expanded-bg`

### 2. State-modifier position is unpredictable across components
The same state lands in a different position depending on the component (and even flips per-state):

| State | role→state | state→role | role-less |
|---|---|---|---|
| hover | `field-color-bg-hover`, `pluginbar-color-text-hover`, `iconbutton-color-bg-hover` | `viewports-control-toggle-hover-text` | `actionbar-color-hover` |
| active | `field-color-text-active`, `iconbutton-color-bg-active` | — | `actionbar-color-active` |
| selected | `outline-color-border-selected` (`LayerTree/styles.module.css:79`) | `slot-component-color-selected-border` (`DraggableComponent/styles.module.css:76`) | — |
| dragging | `field-color-border-dragging` (`AutoField/fields/ArrayField/styles.module.css:128`) | `slot-component-color-dragging-border` (`DraggableComponent/styles.css:48`) | — |
| disabled | `iconbutton-color-icon-disabled` (`IconButton/IconButton.module.css:62`) | `field-color-disabled-bg` (`AutoField/styles.module.css:129`), `button-primary-color-disabled-bg` (`Button/Button.module.css:126`) | `actionbar-color-disabled` |

Note `selected`/`dragging` are `state→role` on slot but `disabled` is `state→role` on field/button and `role→state` on icon-button — there's no single rule.

### 3. Sidebar: color uses `sidebar→side`, width uses `side→sidebar`
| | Token | Location |
|---|---|---|
| color | `--puck-sidebar-left-color-bg` | `Puck/components/Sidebar/styles.module.css:16` |
| width | `--puck-left-sidebar-width` | `Puck/components/Layout/styles.module.css:43` |

### 4. The same left rail has two names
| | Token | Location |
|---|---|---|
| color | `--puck-pluginbar-color-bg` | `Puck/components/Layout/styles.module.css:199` |
| width | `--puck-side-nav-width` | `Puck/components/Layout/styles.module.css:41` |

### 5. `--puck-outline-list-connector` — no intent segment + raw-palette default
`packages/core/components/OutlineList/styles.module.css:2` → `var(--puck-color-grey-08)`. Should carry an intent segment and default to a semantic token, or be made private/documented.

### 6. opacity tokens prefixed `color-`
`ActionBar/styles.module.css:21` `--puck-actionbar-color-text-opacity` and `:83` `--puck-actionbar-color-disabled-opacity` label an opacity as a color.

### 7. Sortable drag placeholder is untouched 
`Sortable/styles.css still hardcodes` `--puck-color-azure-09 !important` + `#00000000`, while DraggableComponent’s drag visuals were tokenized. Drag theming is half-done.
### 8. #00000000 instead of transparent 
(`DraggableComponent/styles.css:30`, `Sortable/styles.css:8`)

### 9. `--puck-field-color-bg` vs `--puck-field-color-surface`
two background tokens in one namespace with identical defaults and overlapping meaning; likely to confuse. Document the split or merge.

### 10. Coverage asymmetry 
Header exposes 2 tokens (bg/text), Canvas 3, while Fields exposes ~35. Worth a deliberate call on minimum themable surface per componen

### Suggested resolution
Pick one ordering rule and apply it everywhere; give every token an intent segment; default to semantic tokens (not raw palette); make `ViewportControls` + `outline-list-connector` either `--_puck-*` private or documented-and-renamed; and reconcile the inherited Layout vars (`side-nav-width`, `*-sidebar-width`) into the scheme.

xaviemirmon (approved):

PR #1718, opened 2026-07-02
Add dictionary

### LINKED ISSUE #1134: Add localization
## Description

Support localization for all hardcoded strings.

## Considerations

- We need to identify full list of strings to be translated
- This might be used for use-cases outside of localization

### Proposal 1

Add a dictionary 

```tsx
const dictionary: PuckDictionary = {
  "page-breadcrumb": "Hello, world",
  "publish-button": "Publish me"
}

<Puck dictionary={dictionary} /> 
```

FedericoBonel:
Closes #1134

## Description

This PR adds the `dictionary` mentioned in the "Localization" issue. It allows for developers to override any of the strings surfaced to the user in the browser. These include visible and non visible strings.

## Changes made

- Introduced a new `dictionary` prop to the Puck component that expects a partial object with the strings that you wish to override. This prop is reactive and will update with any changes. The keys of the dictionary are prefixed by the section where they're used, and allow interpolation using `{value}`. Also allows for hiding strings by passing `""`.
- Introduced a `useMessage` hook that reads the values from the dictionary by key and interpolates any values on it.

## How to test

- Render the Puck component and override any of the keys specified in the [`dictionary`](https://github.com/puckeditor/puck/blob/f118fbe68509bb451fc1ad1152bce712c5efe474/packages/core/lib/dictionary.ts) component.
- Navigate to the editor
- Observe the strings changing

## TODO

- [x] Add documentation
- [x] Expose the dictionary object in the `usePuck` hook
- [x] Allow for users to provide additional custom keys to the `dictionary` object so they can decide if they need more strings.

### DIFF
--- a/apps/docs/pages/docs/api-reference/dictionary.mdx
+++ b/apps/docs/pages/docs/api-reference/dictionary.mdx
@@ -1,0 +1,172 @@
+---
+title: Dictionary
+---
+
+# Dictionary
+
+The dictionary defines the text strings used throughout Puck's UI.
+
+Use this to implement [localization](/docs/integrating-puck/localization) or to override default messages:
+
+```tsx showLineNumbers copy
+const dictionary = {
+  "header-publish": "Save",
+};
+
+<Puck dictionary={dictionary} />;
+```
+
+When a default string contains a template variable (e.g. `{count}`), you can use that to position the variable in your string:
+
+```tsx showLineNumbers copy
+const dictionary = {
+  "field-external-result-singular": "Total: {count}",
+  "field-external-result-plural": "Total: {count}",
+};
+```
+
+The dictionary also accepts your own keys, which you can read from the [internal Puck API](/docs/api-reference/puck-api#dictionary). See [Custom strings](/docs/integrating-puck/localization#custom-strings).
+
+## Tokens
+
+Tokens follow a `{area}-{name}-{variant}` format.
+
+Many tokens set an element's HTML `title` attribute, a tooltip that only appears on hover, rather than always-visible text. These are marked _(tooltip)_ below.
+
+### Header
+
+| Token                        | Default                  | Used for                         |
+| ---------------------------- | ------------------------ | -------------------------------- |
+| `header-publish`             | `"Publish"`              | Publish button                   |
+| `header-undo`                | `"undo"`                 | Undo button (tooltip)            |
+| `header-redo`                | `"redo"`                 | Redo button (tooltip)            |
+| `header-toggle-leftsidebar`  | `"Toggle left sidebar"`  | Sidebar toggle (tooltip)         |
+| `header-toggle-rightsidebar` | `"Toggle right sidebar"` | Sidebar toggle (tooltip)         |
+| `header-toggle-menubar`      | `"Toggle menu bar"`      | Mobile menu bar toggle (tooltip) |
+
+### Actions
+
+| Token                 | Default           | Used for                       |
+| --------------------- | ----------------- | ------------------------------ |
+| `action-selectparent` | `"Select parent"` | Component action bar (tooltip) |
+| `action-duplicate`    | `"Duplicate"`     | Component action bar (tooltip) |
+| `action-delete`       | `"Delete"`        | Component action bar (tooltip) |
+
+### Labels
+
+| Token             | Default       | Used for                                                |
+| ----------------- | ------------- | ------------------------------------------------------- |
+| `label-page`      | `"Page"`      | Breadcrumbs, fields panel title, header title fallback  |
+| `label-component` | `"Component"` | Outline and breadcrumbs, for unresolved component types |
+
+### Outline
+
+| Token              | Default      | Used for               |
+| ------------------ | ------------ | ---------------------- |
+| `outline-empty`    | `"No items"` | Empty zone             |
+| `outline-collapse` | `"Collapse"` | Layer toggle (tooltip) |
+| `outline-expand`   | `"Expand"`   | Layer toggle (tooltip) |
+
+### Drawer
+
+| Token                      | Default              | Used for                                      |
+| -------------------------- | -------------------- | --------------------------------------------- |
+| `drawer-category-collapse` | `"Collapse {title}"` | Category toggle (tooltip)                     |
+| `drawer-category-expand`   | `"Expand {title}"`   | Category toggle (tooltip)                     |
+| `drawer-category-other`    | `"Other"`            | Default category for uncategorized components |
+
+### Canvas
+
+| Token             | Default                         | Used for                                           |
+| ----------------- | ------------------------------- | -------------------------------------------------- |
+| `canvas-noconfig` | `"No configuration for {type}"` | Placeholder for components missing from the config |
+
+### Fields
+
+| Token            | Default       | Used for                            |
+| ---------------- | ------------- | ----------------------------------- |
+| `field-readonly` | `"Read-only"` | Read-only field lock icon (tooltip) |
+
+#### Array field
+
+| Token                       | Default           | Used for              |
+| --------------------------- | ----------------- | --------------------- |
+| `field-arrayitem-summary`   | `"Item #{index}"` | Item default label    |
+| `field-arrayitem-duplicate` | `"Duplicate"`     | Item action (tooltip) |
+| `field-arrayitem-delete`    | `"Delete"`        | Item action (tooltip) |
+
+#### External field
+
+| Token                            | Default             | Used for                         |
+| -------------------------------- | ------------------- | -------------------------------- |
+| `field-external-selectdata`      | `"Select data"`     | Trigger button and modal heading |
+| `field-external-search`          | `"Search"`          | Modal search                     |
+| `field-external-togglefilters`   | `"Toggle filters"`  | Modal filter toggle (tooltip)    |
+| `field-external-item`            | `"External item"`   | Selected item default label      |
+| `field-external-result-singular` | `"{count} result"`  | Modal footer                     |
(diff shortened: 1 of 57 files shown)


packages/core/components/Puck/components/Layout/index.tsx
```diff
@@ -192,11 +194,19 @@ export const Layout = ({ children }: { children?: ReactNode }) => {
     [plugins]
   );
 
+  // Localized labels for the built-in plugin tabs, passed into their factories.
+  const blocksLabel = useMessage("plugin-blocks");
+  const outlineLabel = useMessage("plugin-outline");
+  const fieldsLabel = useMessage("plugin-fields");
+
```
chrisvxd, line 201: I think this is fine but it's also achievable by loading the built in plugins explicitly

packages/core/lib/dictionary.ts
```diff
@@ -0,0 +1,113 @@
+/**
+ * Default messages for every hard-coded string shown in the Puck editor UI.
+ *
+ * Keys are area-prefixed kebab-case.
+ */
+export const defaultDictionary = {
+  // Header
+  "action-publish": "Publish",
```
chrisvxd: Not 100% on the action- prefix for this and undo/redo as it could be conflated with action bar. What do you think?

chrisvxd (commented):
Great stuff! Couple of comments. Lots to document.

packages/core/components/Puck/components/Layout/index.tsx
FedericoBonel: Yeah I also thought about that, since I added the new props for them you can really add whatever you want, but I thought it might be good to also add them here since this would work without having to pass them explicitly

packages/core/lib/use-component-list.tsx
```diff
@@ -31,7 +33,13 @@ export const useComponentList = () => {
             <ComponentList
               id={categoryKey}
               key={categoryKey}
-              title={category.title || categoryKey}
+              // Prefer the config title (reactive) over the ui snapshot, so
+              // updating `config.categories[x].title` shows new title immediately.
```
FedericoBonel, line 37: Thinking about this, we might also want to update the uiComponentList whenever the config changes instead

FedericoBonel:
Added some docs, also refactored some keys since I realized they were a bit inconsistent and we still had a lot of rich text messages that were not covered. Also added a little fix to actually react to changes in the field configs since I realized some were not changing.

packages/core/lib/dictionary.ts
FedericoBonel: Changed it!

apps/docs/pages/docs/api-reference/dictionary.mdx
```diff
@@ -0,0 +1,159 @@
...
+| `field-external-togglefilters`   | `Toggle filters`  | Modal filter toggle              |
+| `field-external-item`            | `External item`   | Selected item default label      |
+| `field-external-result-singular` | `{count} result`  | Modal footer                     |
+| `field-external-result-plural`   | `{count} results` | Modal footer                     |
+
+#### Rich text field
+
+| Token                                | Default           | Used for                |
+| ------------------------------------ | ----------------- | ----------------------- |
+| `field-richtext-bold`                | `Bold`            | Formatting button       |
```
chrisvxd: There are a lot of keys that override `title` attributes. I think it might be worth pointing that out in the docs, because the user might not know where it is shown.

apps/docs/pages/docs/api-reference/dictionary.mdx
```diff
@@ -0,0 +1,159 @@
...
+
+| Token             | Default                       | Used for                                           |
+| ----------------- | ----------------------------- | -------------------------------------------------- |
+| `canvas-noconfig` | `No configuration for {type}` | Placeholder for components missing from the config |
+
+### Fields
+
+| Token            | Default     | Used for                  |
+| ---------------- | ----------- | ------------------------- |
+| `field-readonly` | `Read-only` | Read-only field lock icon |
```
chrisvxd: The default string values should use double quotations for consistency with other docs, I.e. `"Read-only"`

apps/docs/pages/docs/api-reference/plugins/outline-plugin.mdx
```diff
@@ -20,4 +22,17 @@ export function Editor() {
...
+| [`icon`](#icon)   | `icon: <Layers />` | ReactNode | -      |
+| [`label`](#label) | `label: "Outline"` | String    | -      |
+
+## Optional params
+
+### `icon`
+
+React node displayed alongside `label` in the plugin rail. Defaults to a layers icon.
+
+### `label`
```
chrisvxd, line 47: Okay on second thoughts, I think we should still support dictionary for this so the user can do it all in one place for the built in plugins, which makes using react-i18n possible.

I think the added label and icon are also fine.

apps/docs/pages/docs/api-reference/plugins/outline-plugin.mdx
```diff
@@ -9,7 +9,9 @@ Show an outline of the components currently on the canvas.
 ```tsx copy
 import { Puck, outlinePlugin } from "@puckeditor/core";
 
-const outline = outlinePlugin();
+const outline = outlinePlugin({
+  label: "Outline",
```
chrisvxd: We don't include optional params in the leading example

apps/docs/pages/docs/integrating-puck/localization.mdx
```diff
@@ -0,0 +1,201 @@
+# Localization
+
+Puck's default UI uses English. Use the [`dictionary`](/docs/api-reference/dictionary) prop to translate or reword Puck-owned strings.
```
chrisvxd: "US English"

apps/docs/pages/docs/integrating-puck/localization.mdx
```diff
@@ -0,0 +1,201 @@
...
+            title: {
+              ...config.components.HeadingBlock.fields.title,
+              label: labels.fields.title,
+            },
+          },
+        },
+      },
+    }),
+    [labels]
+  );
```
chrisvxd, line 160: Can you simplify this? A bit hard to read

chrisvxd (changes requested):
Great! Some comments on docs

apps/docs/pages/docs/api-reference/dictionary.mdx
FedericoBonel: Done!

apps/docs/pages/docs/api-reference/dictionary.mdx
FedericoBonel: Done!

apps/docs/pages/docs/api-reference/plugins/outline-plugin.mdx
FedericoBonel: Fixed it!

apps/docs/pages/docs/integrating-puck/localization.mdx
FedericoBonel: Good call! Fixed it.

apps/docs/pages/docs/integrating-puck/localization.mdx
FedericoBonel: Simplified it a bit, let me know what you think!

apps/docs/pages/docs/api-reference/plugins/outline-plugin.mdx
FedericoBonel: Rolled them back!

FedericoBonel:
Addressed your comments and added support for custom strings 🏓

packages/core/store/slices/fields.ts
```diff
@@ -107,9 +115,23 @@ export const useRegisterFieldsSlice = (
...
+    // until another component is selected or data happens to change.
+    const unsubscribeConfig = appStore.subscribe(
+      (s) => s.config,
+      () => resolveFields(true)
+    );
+
+    return () => {
+      unsubscribeData();
+      unsubscribeConfig();
+    };
```
chrisvxd, line 135: I'm not sure I understand the message in this commit. Is this when _changing_ the config? If so, the commit should specify that.

apps/docs/pages/docs/integrating-puck/localization.mdx
```diff
@@ -0,0 +1,228 @@
...
+      "field-richtext-bold": t("field-richtext-bold"),
+    }),
+    [t]
+  );
+
+  return <Puck dictionary={dictionary} config={config} data={data} />;
+}
+```
+
+## Translating user-defined labels
```
chrisvxd, line 77: I find the title on this a bit confusing, because I can't help but feel there is a missing section to this page: adding localization to your Puck app.

It makes sense that I would come to this page looking for how to provide multi-language support to my puck fields.

How do you think we should handle this?

chrisvxd (changes requested):
Great stuff! Clocked a couple more things on another review.

packages/core/store/slices/fields.ts
FedericoBonel: Yes it's while changing the config, good call though I'll add that to the message

apps/docs/pages/docs/integrating-puck/localization.mdx
FedericoBonel: I see what you mean. 

Do you think I should maybe add a library like the example above (`useTranslation`) so that's clearer?

chrisvxd: As discussed, will do as a follow-on

chrisvxd (approved):

Issue #1397, opened 2025-10-24
Race conditions on resolver's caches


FedericoBonel:
## Description

The current implementation of resolvers relies on caches for Puck to determine when they should execute. This generally works well, but because resolvers are asynchronous, the cache state is shared across concurrent promises.

This creates a race condition when a component's data changes rapidly. Since the cache isn't updated until the resolver finishes, multiple updates can trigger multiple executions that all read an outdated or empty cache. This can result in unexpected executions.

`resolveFields` appears less affected by this because its resolvers only run for the currently selected component.

Effectively, we have a critical section around resolver execution that is unprotected.

We should introduce a locking mechanism for resolver executions on each component, so each execution waits for the previous one to complete before accessing the cache. A semaphore or similar mechanism could be a good fit.

Alternatively, we could store resolver parameters in the cache *before* executing the resolver. If two calls use the same parameters, we could treat the result as cached. However, this risks breaking cases where the resolver has side effects independent of its parameters (e.g., closures), and may still run into the same race condition.

## Environment

- **Puck version**: 0.20.2

## Steps to Reproduce

Due to the nature of race conditions, it can be hard to consistently reproduce. However, the issue can be observed with the following test:

**For `resolvePermissions`:**

1. Comment out this [`waitFor`](https://github.com/puckeditor/puck/blob/main/packages/core/store/slices/__tests__/permissions.spec.tsx#L677) in the test that checks `resolvePermissions` does not run when neither data nor the parent have changed after a reordering in the same slot:
   ```tsx
   // await waitFor(
   //   () => Object.keys(appStore.getState().permissions.cache).length > 0
   // );
   ```
2. Add `.only` to the relevant `it` block [here](https://github.com/puckeditor/puck/blob/main/packages/core/store/slices/__tests__/permissions.spec.tsx#L615):
   ```tsx
   it.only("doesn't update if moving on the same parent", async () => {
   ```
3. Add a `console.log` of the cache in the [permissions slice resolver](https://github.com/puckeditor/puck/blob/049cfbbcf3602c32a2beeb85298e0ca8728215bf/packages/core/store/slices/permissions.ts#L83):
   ```tsx
   console.log("CURRENT CACHE: ", cache);
   const parentChanged = itemCache?.lastParentId !== parentId;
   ```
4. Run the test:
   ```bash
   npx turbo run test -- ./packages/core/store/slices/__tests__/permissions.spec.tsx
   ```
[...]

PR #1721, opened 2026-07-03
fix(core): prevent unbounded cache growth in resolveAllData

### LINKED ISSUE #1459: Memory leak in resolveComponentData cache - cache.lastChange grows unbounded with unique component IDs
## Description

The `resolveComponentData` function in `lib/resolve-component-data.ts` maintains a module-level cache (`cache.lastChange`) that stores resolved data per component ID. This cache never gets cleared, causing memory to accumulate when processing many pages with unique component IDs.

When `resolveAllData` is called multiple times (e.g., when processing multiple pages in a batch renderer), each unique component ID creates a new entry in `cache.lastChange`. These entries persist across all invocations, leading to unbounded memory growth.

## Environment

- Puck version: 0.20.2
- Browser version: N/A (occurs in Node.js/Deno runtime)
- Additional environment info: 
  - we generate (call Render) on thousands of pages in a Deno context that is shared for 5 minutes
  - Each page has unique component IDs

## Steps to reproduce

1. Call `resolveAllData` multiple times with different component IDs
2. (optional) Repeat this pattern many times (e.g., processing 1000+ pages in a batch renderer)
3. Observe memory usage - it will grow continuously as each unique component ID adds a new cache entry

## What happens

Memory grows unbounded during the 5-minute window, eventually crashing the Deno process due to exceeding the memory limit.

## What I expect to happen
1. Don't use a cache at all for `<Render/>`

or one/all of the following:

1. Cache should not grow unbounded
2. Memory usage should remain stable when processing many pages
3. Option to clear cache between page processing cycles

chrisvxd:
Closes #1459

## Description

`resolveComponentData` writes resolved entries to a module-level cache keyed by component ID. `resolveAllData` always resolves with the `"force"` trigger, so the cache is never read to skip work there — it only accumulates one entry per unique component ID and never frees them. When resolving many pages with unique IDs in a long-lived shared runtime (e.g. batch rendering thousands of pages), this grows memory unbounded.

The fix makes the cache injectable and gives `resolveAllData` its own local cache, so its entries are garbage collected once the call returns. The editor's shared-cache behaviour is unchanged.

## Changes made

- `resolveComponentData` accepts an optional `cacheStore` param (defaults to the shared module-level `cache`), and a `ResolveDataCache` type is exported.
- `resolveAllData` passes a fresh local cache, so it no longer writes to the shared cache.
- Added a test asserting `resolveAllData` does not pollute the shared cache.

## How to test

Call `resolveAllData` repeatedly with distinct component IDs and confirm the module-level `cache.lastChange` stays empty (covered by the new test), and that resolved output is unchanged.

### DIFF
--- a/packages/core/lib/__tests__/resolve-all-data.spec.tsx
+++ b/packages/core/lib/__tests__/resolve-all-data.spec.tsx
@@ -1,5 +1,6 @@
 import { ComponentData, Config, Data, RootData } from "../../types";
 import { resolveAllData } from "../resolve-all-data";
+import { cache } from "../resolve-component-data";
 
 const item4 = {
   type: "ComponentWithResolveProps",
@@ -363,5 +364,15 @@ describe("resolve-data", () => {
       item1_2.props.id
     );
     expect(receivedParentById[item1_2_1.props.id]?.props.prop).toBe("Resolved");
+  });
+
+  it("should not write to the shared module-level cache", async () => {
+    cache.lastChange = {};
+
+    await resolveAllData(data, config);
+
+    // Entries should live in a local cache that is garbage collected once
+    // resolveAllData returns, not accumulate in the shared cache.
+    expect(cache.lastChange).toEqual({});
   });
 });
--- a/packages/core/lib/resolve-all-data.ts
+++ b/packages/core/lib/resolve-all-data.ts
@@ -8,7 +8,10 @@ import {
   Metadata,
   RootData,
 } from "../types";
-import { resolveComponentData } from "./resolve-component-data";
+import {
+  ResolveDataCache,
+  resolveComponentData,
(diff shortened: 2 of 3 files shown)


FedericoBonel (approved):
Great! I prefer this approach, it feels similar to memoization in DP.

I have to say, however, that this doesn't fix the issue where, if you add a bunch of components to a page and then remove them, the cache keeps growing bigger and bigger (since we never clear from it). I think that's something we mention in #1397 though, and it involves a bigger refactor, so I'm approving this.

That said, I'm wondering if we might be able to avoid this altogether once we implement the caching solution we need for it and that gets wired into `resolveAllData`.

chrisvxd:
@FedericoBonel merged under assumption can be revisited in #1397

PR #1722, opened 2026-07-03
fix(core): allow overriding title type in root props

### LINKED ISSUE #1263: Overriding title in RootProps makes Puck complain about data
## Description

```tsx
import { type Config, type Data, Puck } from "@measured/puck";
import type { UserGenerics } from "@measured/puck/dist/no-external";

type FooRootProps = {
  title: number;
};
type FooComponents = {
  Box: { size: number };
};

const testData = 0 as unknown as Data<FooComponents, FooRootProps>;
const testConfig = 0 as unknown as Config<{
  components: FooComponents;
  root: FooRootProps;
}>;

// this is ok
const y: UserGenerics<{ components: FooComponents; root: FooRootProps }> =
  testData;

function TestComponent() {
  // data has an Typescript error
  return <Puck config={testConfig} data={testData} />;
}
```

The error is
> Type `Data<FooComponents, FooRootProps>` is not assignable to type`Partial<Data<FooComponents, FooRootProps & DefaultRootFieldProps>>`, since title is `number` and cannot be assigned to `number & string` which is `never`. 

The issue is the [following](https://github.com/puckeditor/puck/blob/main/packages/core/types/Utils.tsx),
```ts
export type UserGenerics<
  // ...
> = {
   // ...
   UserRootProps: UserParams["rootProps"] & DefaultRootFieldProps;
   // ...
}
```

## Environment

- Puck version: 0.20.0, 0.20.1. No issue in 0.19.3
- Typescript version: 5.9.2, 7.0.0 (tsgo)

chrisvxd:
Closes #1263

## Description

Overriding `title` in root props with a non-string type (e.g. `title: number`) made `Puck` reject the `data` prop with a type error. The default root field props (`{ title?: string }`) were merged into user root props with a plain intersection, so `number & string` collapsed to `never`.

This introduces a `WithDefaultRootFieldProps` helper that omits any keys the user has already defined before merging, so user-provided prop types take precedence over the defaults.

## Changes made

- Added `WithDefaultRootFieldProps<RootProps>` in `types/Props.tsx`.
- Used it in `ExtractConfigParams` (`types/Config.tsx`) and `UserGenerics.UserRootProps` (`types/Utils.tsx`) in place of the plain `& DefaultRootFieldProps` intersection.

## How to test

```tsx
type FooRootProps = { title: number };
type FooComponents = { Box: { size: number } };

const testData = 0 as unknown as Data<FooComponents, FooRootProps>;
const testConfig = 0 as unknown as Config<{
  components: FooComponents;
  root: FooRootProps;
}>;

// no longer errors
<Puck config={testConfig} data={testData} />;
```

### DIFF
--- a/packages/core/types/Props.tsx
+++ b/packages/core/types/Props.tsx
@@ -13,6 +13,14 @@ export type DefaultRootFieldProps = {
   title?: string;
 };
 
+/**
+ * Merge the default root field props into user-provided root props, letting
+ * the user's props take precedence. Using a plain intersection would turn an
+ * overridden `title` (e.g. `number`) into `never` (`number & string`).
+ */
+export type WithDefaultRootFieldProps<RootProps> = RootProps &
+  Omit<DefaultRootFieldProps, keyof RootProps>;
+
 export type DefaultRootRenderProps<
   Props extends DefaultComponentProps = DefaultRootFieldProps
 > = WithPuckProps<WithChildren<Props>>;
--- a/packages/core/types/Config.tsx
+++ b/packages/core/types/Config.tsx
@@ -4,7 +4,7 @@ import { ComponentData, ComponentMetadata, RootData } from "./Data";
 
 import { AsFieldProps, WithChildren, WithId, WithPuckProps } from "./Utils";
 import { AppState } from "./AppState";
-import { DefaultComponentProps, DefaultRootFieldProps } from "./Props";
+import { DefaultComponentProps, WithDefaultRootFieldProps } from "./Props";
 import { Permissions } from "./API";
 import { DropZoneProps } from "../components/DropZone/types";
 import {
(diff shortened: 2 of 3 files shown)


packages/core/types/Props.tsx
```diff
@@ -13,6 +13,14 @@ export type DefaultRootFieldProps = {
   title?: string;
 };
 
+/**
+ * Merge the default root field props into user-provided root props, letting
+ * the user's props take precedence. Using a plain intersection would turn an
+ * overridden `title` (e.g. `number`) into `never` (`number & string`).
+ */
+export type WithDefaultRootFieldProps<RootProps> = RootProps &
```
chrisvxd, line 21: Little unsure if we should be exporting this publicly or putting it in our internal types. Presume the latter.

FedericoBonel: I could see people wanting to use this if we added more default props, at the moment is only one though. I would leave it in case we do end up adding additional default props (i.g. when we migrate to a full recursive tree we might need to add a `content` slot).

FedericoBonel (approved):

PR #1723, opened 2026-07-03
feat(core): forward additional props to slot `as` component

### LINKED ISSUE #1441: Include additional props when using the `as` prop in Slots
## Description

Currently, using the `as` slot render prop to provide your own components or elements for slots does not support passing down additional props. While that might be fine for some cases, there are many situations where we want to pass props to the component we use for the slot.

## Proposal

Similar to how MUI handles it, we could allow any additional props to be passed to the component provided:

```tsx
render: ({ SlotComponent, height }) => {
  // Example: Box accepts an sx prop for styling with your own theming solution
  return <SlotComponent as={Box} sx={{ height: (theme) => theme.units * height }} />
}
```

For implementation, the slot render function could capture all remaining props via spreading, and then apply them to the component when it renders.

chrisvxd:
Closes puckeditor/puck#1441

## Description

Slots (and `puck.renderDropZone`) accept an `as` prop to render a custom element or component, but previously any other props were dropped. This forwards those additional props onto the `as` element/component, so you can pass e.g. styling props through:

```tsx
render: ({ SlotComponent, height }) => (
  <SlotComponent as={Box} sx={{ height: (theme) => theme.units * height }} />
);
```

Puck-internal props (`zone`, `allow`, `disallow`, `collisionAxis`, `minEmptyHeight`, `content`) are stripped so they don't leak onto the DOM.

## Changes made

* `DropZoneProps` is now polymorphic over `as` (defaulting to `"div"`), so additional props are typed against the chosen element/component. `SlotComponent` is typed accordingly.
* `SlotRender`, `DropZoneRender`, and `DropZoneEdit` spread the remaining props onto the rendered element, after removing the Puck-internal ones.
* Documented the behaviour in the slot field reference.

## How to test

Add extra props alongside `as` in a slot render function and confirm they reach the element and that internal props don't leak:

```tsx
  <Content as="section" id="my-slot" data-custom="foo" />
);
```

Covered by a new test in `packages/core/components/ServerRender/__tests__/index.spec.tsx`.

### DIFF
--- a/apps/docs/pages/docs/api-reference/fields/slot.mdx
+++ b/apps/docs/pages/docs/api-reference/fields/slot.mdx
@@ -225,6 +225,20 @@ const config = {
 };
 ```
 
+Any additional props are forwarded to the element or component provided via `as`, typed against it.
+
+```tsx copy {5}
+const config = {
+  components: {
+    Example: {
+      render: ({ content: Content }) => {
+        return <Content as={Box} sx={{ display: "flex" }} />;
+      },
+    },
+  },
+};
+```
+
 ### `className`
 
 Provide a className to the rendered element. The default styles will still be applied.
--- a/apps/docs/pages/docs/api-reference/fields/slot.mdx
+++ b/apps/docs/pages/docs/api-reference/fields/slot.mdx
@@ -239,6 +239,17 @@ const config = {
 };
 ```
 
+The following Puck-internal props are **not** forwarded, as they are consumed by Puck itself:
+
+| Prop | Reason |
+| ---- | ------ |
+| `zone` | Identifies the drop zone internally |
+| `allow` | Controls which component types are allowed |
+| `disallow` | Controls which component types are disallowed |
+| `collisionAxis` | Configures drag collision detection |
+| `minEmptyHeight` | Sets the minimum height of an empty zone |
+| `content` | Injected by Puck with the slot's component data |
+
 ### `className`
 
 Provide a className to the rendered element. The default styles will still be applied.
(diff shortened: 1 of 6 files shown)


apps/docs/pages/docs/api-reference/fields/slot.mdx
```diff
@@ -225,6 +225,20 @@ const config = {
 };
 ```
 
+Any additional props are forwarded to the element or component provided via `as`, typed against it.
```
FedericoBonel, line 228: I think we should mention which props wouldn't get passed down because they collide with Puck and should be avoided (e.g. `content`). I know we enforce it with typescript, but there's also some we accept and people/other libraries might be using.

FedericoBonel (changes requested):
I think this looks good! Tested it with MUI and it works like a charm, the only thing I'd say is that we need to add some docs about which props are not gonna get passed because they collide with Puck.

packages/core/components/DropZone/types.ts
```diff
@@ -1,14 +1,31 @@
...
+ *
+ * Any additional props not consumed by Puck are forwarded to the element or
+ * component provided via `as` (defaulting to a `div`), typed against it.
+ */
+export type DropZoneProps<T extends ElementType = "div"> = DropZoneOwnProps & {
+  as?: T;
+} & Omit<
+    ComponentPropsWithoutRef<T>,
+    keyof DropZoneOwnProps | "as" | "content"
+  >;
```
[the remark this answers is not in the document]
FedericoBonel: This one might be good to add, we could also just document it as part of the docs I mentioned in the other comment

apps/docs/pages/docs/api-reference/fields/slot.mdx
chrisvxd: @copilot capture this

PR #1726, opened 2026-07-03
fix: run field transforms for fields without default props

### LINKED ISSUE #1342: Field transforms not called if no default props are set
## Description

Field transforms aren't being called if default props aren't set for the field leaving the component selection in a broken state. 

## Environment

- Puck version: 0.20.x
- Browser version: Chrome 135 (desktop)
- Additional environment info: MacOS

## Steps to reproduce

1. Take a component that is reliant on fieldTransforms e.g. Text and don't define default props

```tsx
import React from "react";
import { ALargeSmall, AlignLeft } from "lucide-react";

import { ComponentConfig } from "@/core/types";
import { Section } from "../../components/Section";
import { WithLayout, withLayout } from "../../components/Layout";

export type TextProps = WithLayout<{
  align: "left" | "center" | "right";
  text?: string;
  padding?: string;
  size?: "s" | "m";
  color: "default" | "muted";
  maxWidth?: string;
}>;

const TextInner: ComponentConfig<TextProps> = {
  fields: {
    text: {
      type: "textarea",
      contentEditable: true,
    },
    size: {
      type: "select",
      labelIcon: <ALargeSmall size={16} />,
      options: [
        { label: "S", value: "s" },
        { label: "M", value: "m" },
      ],
    },
    align: {
      type: "radio",
      labelIcon: <AlignLeft size={16} />,
      options: [
        { label: "Left", value: "left" },
        { label: "Center", value: "center" },
        { label: "Right", value: "right" },
      ],
    },
    color: {
      type: "radio",
      options: [
        { label: "Default", value: "default" },
        { label: "Muted", value: "muted" },
      ],
    },
    maxWidth: { type: "text" },
  },
  defaultProps: {
    align: "left",
    size: "m",
    color: "default",
  },
  render: ({ align, color, text, size, maxWidth }) => {
    return (
      <Section maxWidth={maxWidth}>
        <span
          style={{
            color:
              color === "default" ? "inherit" : "var(--puck-color-grey-05)",
            display: "flex",
            textAlign: align,
            width: "100%",
            fontSize: size === "m" ? "20px" : "16px",
            fontWeight: 300,
            maxWidth,
            justifyContent:
              align === "center"
                ? "center"
                : align === "right"
                ? "flex-end"
                : "flex-start",
          }}
        >
          {text}
        </span>
      </Section>
    );
  },
};

export const Text = withLayout(TextInner);

```
_notice the omission of text: "Text"_
2. Run the application in development mode and drag "Text" to the canvas

## What happens

Component doesn't render inner text transforms. 

```html
<div
    class="styles_Layout__vbar2"
    style="
        grid-area: span 1 / span 1;
        padding-top: 0px;
        padding-bottom: 0px;
        position: relative;
    "
    data-puck-component="Text-6ae5d633-7413-4446-a42b-117a1f632d21"
    data-puck-dnd="Text-6ae5d633-7413-4446-a42b-117a1f632d21"
    tabindex="0"
    role="button"
    aria-roledescription="draggable"
    aria-describedby="dnd-kit-description-2"
    aria-pressed="false"
    aria-grabbed="false"
    aria-disabled="false"
>
    <div class="styles_Section__4tQXe">
        <div class="styles_Section-inner__8sPWU" style="max-width: 1280px">
            <span
                style="
                    color: inherit;
                    display: flex;
                    text-align: left;
                    width: 100%;
                    font-size: 20px;
                    font-weight: 300;
                    justify-content: flex-start;
                "
            ></span>
        </div>
    </div>
</div>
```
## What I expect to happen

Component should render inline controls

```html

https://github.com/user-attachments/assets/23791d2f-53e4-4458-9c07-cc11030af4e7

<div
    class="styles_Layout__vbar2"
    style="
        grid-area: span 1 / span 1;
        padding-top: 0px;
        padding-bottom: 0px;
        position: relative;
    "
    data-puck-component="Text-a4012272-18cc-40fa-9fbc-f6f9c5ac87c1"
    data-puck-dnd="Text-a4012272-18cc-40fa-9fbc-f6f9c5ac87c1"
    tabindex="0"
    role="button"
    aria-roledescription="draggable"
    aria-describedby="dnd-kit-description-2"
    aria-pressed="false"
    aria-grabbed="false"
    aria-disabled="false"
>
    <div class="styles_Section__4tQXe">
        <div class="styles_Section-inner__8sPWU" style="max-width: 1280px">
            <span
                style="
                    color: inherit;
                    display: flex;
                    text-align: left;
                    width: 100%;
                    font-size: 20px;
                    font-weight: 300;
                    justify-content: flex-start;
                "
                ><span
                    class="styles_InlineTextField__3gFuB"
                    contenteditable="false"
                    data-puck-overlay-portal="true"
                    ></span
                ></span
            >
        </div>
    </div>
</div>
```

## Additional Media

https://github.com/user-attachments/assets/a56765c4-66d7-41ef-95d0-369782d0bc41

chrisvxd:
Closes #1342

## Description

Field transforms only ran for props that had a value. `walkObject` iterated over the props object (`Object.entries(value)`) rather than the component's configured fields, so a field defined in `fields` but absent from props (no default) never reached its transform. In practice a `contentEditable` text/textarea field without a default prop was never transformed, so inline editing controls weren't rendered and the component was left in a broken selection state.

## Changes made

- `walkObject` now also visits fields that have a registered mapper but no value, running their transform with an `undefined` value. Slots are excluded since they're already defaulted separately via `defaultSlots`. Gated behind an `ownedFields` flag so it only applies where the field set actually describes the object (top-level props and `object` field recursion), never the fallback object-walk path.
- Added `map-fields` unit tests covering missing top-level fields, missing object subfields, fields that still have values, and field types with no mapper.

## How to test

- Add a component with a `contentEditable` text/textarea field and omit that field from `defaultProps` (e.g. the demo `Text` component without `text: "Text"`).
- Drag it onto the canvas and confirm the inline editing control (`InlineTextField`) renders and is editable.

### DIFF
--- a/packages/core/lib/data/__tests__/map-fields.spec.tsx
+++ b/packages/core/lib/data/__tests__/map-fields.spec.tsx
@@ -1,0 +1,78 @@
+import { Config } from "../../../types";
+import { mapFields, Mappers } from "../map-fields";
+
+type Props = {
+  Comp: {
+    title?: string;
+    plain?: string;
+    nested?: { title?: string };
+  };
+};
+
+const config: Config<Props> = {
+  components: {
+    Comp: {
+      fields: {
+        title: { type: "text" },
+        plain: { type: "text" },
+        nested: {
+          type: "object",
+          objectFields: {
+            title: { type: "text" },
+          },
+        },
+      },
+      render: () => <div />,
+    },
+  },
+};
+
+const mappers: Mappers = {
+  text: ({ value, propPath }) => `mapped:${propPath}:${value}`,
+};
+
+describe("mapFields", () => {
+  it("runs transforms for fields that have no value set", () => {
+    const result: any = mapFields(
+      { type: "Comp", props: { id: "abc" } },
+      mappers,
+      config
+    );
+
+    // `title` has a mapper but no value in props — it should still be visited
+    expect(result.props.title).toBe("mapped:title:undefined");
+    expect(result.props.plain).toBe("mapped:plain:undefined");
+  });
+
+  it("still runs transforms for fields that do have a value", () => {
+    const result: any = mapFields(
+      { type: "Comp", props: { id: "abc", title: "Hello" } },
+      mappers,
+      config
+    );
+
+    expect(result.props.title).toBe("mapped:title:Hello");
+  });
+
+  it("does not add keys for field types without a mapper", () => {
+    const result: any = mapFields(
+      { type: "Comp", props: { id: "abc" } },
+      // No mapper registered at all
+      {},
+      config
+    );
+
+    expect("title" in result.props).toBe(false);
+    expect("plain" in result.props).toBe(false);
+  });
+
+  it("runs transforms for missing subfields of an object field", () => {
+    const result: any = mapFields(
+      { type: "Comp", props: { id: "abc", nested: {} } },
+      mappers,
+      config
+    );
+
+    expect(result.props.nested.title).toBe("mapped:nested.title:undefined");
+  });
+});
--- a/packages/core/lib/data/map-fields.ts
+++ b/packages/core/lib/data/map-fields.ts
@@ -43,6 +43,7 @@ type WalkObjectOpts = {
   getPropPath: (str: string) => string;
   config: Config;
   recurseSlots?: boolean;
+  ownedFields?: boolean;
 };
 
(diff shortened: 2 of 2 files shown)


FedericoBonel (approved):
LGTM, unrelated but the map-fields function is getting a bit out of hand might be worth doing a proper refactor at some point.

PR #1734, opened 2026-07-10
Add draggable outline

### LINKED ISSUE #730: Improve Outline UI

FedericoBonel:
Closes puckeditor/puck#730 and puckeditor/puck#1048

## TODO

- [X] Add prop to disable drag and drop in the outline
- [X] Add docs explaining that allows and disallows only works with slot configs
- [X] Refactor LayerTree styles (they're correct, but a bit of a mess)

## Description

This PR improves the outline:

* Makes components draggable
* Supports multiple components in the outline to be expanded at once
* Supports component duplication and deletion from the outline
* Improves styles

## Changes made

* Refactored the LayerTree component:
  * Broke it down to individual files for each part
* 

## How to test

*


packages/core/components/LayerTree/styles.module.css
```diff
@@ -1,5 +1,59 @@
...
+  );
+  --_puck-outline-border-width: var(
+    --puck-outline-border-width,
+    var(--puck-border-width-regular)
+  );
+  --_puck-outline-radius: var(--puck-outline-radius, var(--puck-radius-m));
+  --_puck-outline-caret-size: var(
+    --puck-outline-action-size,
+    var(--puck-icon-size-xs)
+  );
```
chrisvxd, line 18: I get this is inheriting the action size from our theming API, but it feels like it's too small a hit area 🤔

packages/core/components/LayerTree/components/layer-actions/index.tsx
```diff
@@ -0,0 +1,118 @@
...
+    },
+    [dispatch, outlineStore, node.index, node.zoneCompound]
+  );
+
+  if (!permissions.delete && !permissions.duplicate) {
+    return null;
+  }
+
+  return (
+    <div className={getClassName({ visible })}>
```
chrisvxd, line 90: These are too close to the right border 

Image

packages/core/components/LayerTree/styles.module.css
```diff
@@ -8,48 +62,87 @@
...
+}
+
+.LayerActions--visible {
+  visibility: visible;
+}
+
+.Layer-caret svg,
+.LayerActions svg {
+  height: var(--_puck-outline-caret-size);
+  width: var(--_puck-outline-caret-size);
```
chrisvxd: Interesting we're reusing the size between these too. Either way, it's too small for the actions - the hit box is too small, making them awkward to click. We should either increase the size, or just the hit box.

packages/core/components/LayerTree/styles.module.css
```diff
@@ -1,5 +1,59 @@
+.LayerTreeRoot {
```
chrisvxd, line 1: I might do a contribution to this PR since I have a few ideas on the style here.

Tightening up to get things closer to Figma:

Image

packages/core/components/LayerTree/components/layer-tree-zone/index.tsx
```diff
@@ -0,0 +1,77 @@
...
+      </div>
+      {label}
+    </div>
+  );
+};
+
+/**
+ * Renders one zone of the tree: its label followed by its rows.
+ */
+export const LayerTreeZone = ({
```
chrisvxd, line 48: Did you consider making these collapsible too?

packages/core/components/Puck/components/Outline/collapse-all.tsx
```diff
@@ -0,0 +1,42 @@
...
+import { useAppStore } from "../../../../store";
+import { IconButton } from "../../../IconButton";
+
+import styles from "./styles.module.css";
+
+export interface CollapseAllProps {
+  className?: string;
+}
+
+const getClassName = getClassNameFactory("CollapseAll", styles);
```
chrisvxd: As useful as it is, I don't think this looks quite right. I wonder if we should add reintroduce a header to be closer to Figma and only show the collapse button when items are expanded

Image

apps/docs/pages/docs/api-reference/theming/icon-button.mdx
```diff
@@ -40,12 +40,28 @@ Defaults shown as `var(--puck-...)` reference [global tokens](/docs/api-referenc
...
+| `--puck-iconbutton-primary-radius`   | `var(--puck-radius-m)` | Border radius |
+
+**Secondary variant (`--puck-iconbutton-secondary-*`)**
+
+| Token                                      | Default                            | Description   |
+| ------------------------------------------ | ---------------------------------- | ------------- |
+| `--puck-iconbutton-secondary-color-bg`     | `var(--puck-color-surface)`        | Background    |
+| `--puck-iconbutton-secondary-color-border` | `var(--puck-color-border)`         | Border color  |
+| `--puck-iconbutton-secondary-border-width` | `var(--puck-border-width-regular)` | Border width  |
+| `--puck-iconbutton-secondary-radius`       | `var(--puck-radius-round)`         | Border radius |
```
chrisvxd: Would love to sanity check this variant is necessary

chrisvxd (changes requested):
Works brilliantly. Just some thoughts on UI. Happy to collab on getting some of those bits through.

packages/core/components/LayerTree/styles.module.css
FedericoBonel: I actually added a new size for making it smaller, the action bar size made it feel too big for such a small space, but happy to swap it back to the action bar size

packages/core/components/LayerTree/components/layer-actions/index.tsx
FedericoBonel: Ah will change them! Good catch

packages/core/components/LayerTree/styles.module.css
FedericoBonel: Interesting, didn't think about the hitbox increase only. Will do that first we can iterate if needed,

packages/core/components/LayerTree/styles.module.css
FedericoBonel: Ah like this, I can do it, I see what you mean.

packages/core/components/LayerTree/components/layer-tree-zone/index.tsx
FedericoBonel: Interesting, I did have a think about it but didn't get around it. Will try it out and make some additional commits, I'll stack them so we can roll back if needed

packages/core/components/Puck/components/Outline/collapse-all.tsx
FedericoBonel: Happy to do that, I just felt it was a bit redundant since we have the tab name, but will add it and only render when there is any expanded item 🫡

apps/docs/pages/docs/api-reference/theming/icon-button.mdx
FedericoBonel: With the "figma" design for the bar is not. So I'll remove it in the next commits.

packages/core/components/LayerTree/styles.module.css
FedericoBonel: Increased it, I played with increasing the hit area but it was weird and it still eat up the space, so just decided to make them bigger.

packages/core/components/LayerTree/components/layer-actions/index.tsx
FedericoBonel: Added some padding:

Screenshot 2026-07-13 at 15 40 08

packages/core/components/LayerTree/styles.module.css
FedericoBonel: Same as above, ended up using the same size as the action bar, played with adding only padding, but it introduces changes on the IconButton API and it also doesn't fully look right

apps/docs/pages/docs/api-reference/theming/icon-button.mdx
FedericoBonel: Removed it as it was unnecessary after redesign

packages/core/components/Puck/components/Outline/collapse-all.tsx
FedericoBonel: Implemented it like this

packages/core/components/LayerTree/styles.module.css
FedericoBonel: Implemented it like this

packages/core/components/LayerTree/styles.module.css
FedericoBonel: Done

packages/core/components/LayerTree/components/layer-actions/index.tsx
FedericoBonel: Done

packages/core/components/LayerTree/styles.module.css
FedericoBonel: Increased it to the action bar size

packages/core/components/LayerTree/styles.module.css
chrisvxd: Still some missing bits. Use `space-1` for all margins, don't use monospaced fonts for the family (miscommunication here - I meant only for the zone title, but let's just leave it off), increase indentation on layer zones to `space-4`

image

packages/core/components/Puck/components/Outline/collapse-all.tsx
chrisvxd: Needs to be knocked back to `var(--puck-color-grey-03)`

packages/core/components/LayerTree/components/drop-line/styles.module.css
```diff
@@ -0,0 +1,26 @@
+/* Private --_puck-outline-* vars resolve on .LayerTreeRoot. */
+.DropLine {
+  background: var(--_puck-outline-color-drop-indicator);
```
chrisvxd, line 3: Suggest this should default to `var(--puck-color-azure-06)` to match the new dnd behaviour

chrisvxd (changes requested):

packages/core/components/Puck/components/Outline/collapse-all.tsx
FedericoBonel: Good catch! 

Updated it to `var(--puck-color-text-secondary)` to be consistent with the theming system (the default is `--puck-color-grey-04` so a little bit lighter).

packages/core/components/LayerTree/styles.module.css
FedericoBonel: Done!

packages/core/components/LayerTree/components/drop-line/styles.module.css
FedericoBonel: Added a global token for it! we can reuse that on your PR so that it's themable from one place

chrisvxd (approved):
LGTM! Just conflicts remaining.

FedericoBonel:
Rebased! Also added the new strings to the dictionary

PR #1735, opened 2026-07-10
Improve drag and drop for granular changes

### LINKED ISSUE #1286: Limit max height of element while dragging
## Description

Something I noticed while playing around and "abusing" the editor with large elements (full width hero images, etc.) is that on smaller screens like laptops, moving around those large objects becomes cumbersome when they reach a certain height that is close to or larger than the iframe viewport.

Once element height exceeds viewport height, things can get frustrating.

This screen recording shows the janky behaviour on a screen that is about laptop-ish size.

## Proposals

- Max height on element while dragging and
- Max height on drag target

Either define a max-height based on viewport dynamically or allow to set it via css className, that restricts an elements height that is being dragged to an amount that is conveniently below the viewport height. Or limit the height of the active drag target to a max height, so that other targets above and below have a better chance to be reached without having to rely on scrolling while dragging.

To make it visually more appealing, this could received a transparency gradient at the bottom to indicate the fact that there's more.

### LINKED ISSUE #1358: Improve drag-and-drop for smaller slots
## Description

Dragging components into small slots can be fiddly, and the end-user can struggle to insert into the correct component scope.

## Considerations

- We must retain CSS Grid / Flexbox support
- We should retain Puck's fluid animations
- Allow/disallow helps
- Some of this might be mitigated by #730 
- Related to #1286

### Proposal 1

Introduce a "line" preview mode, instead of rendering the whole component preview. Similar to [pragmatic-drag-and-drop](https://github.com/atlassian/pragmatic-drag-and-drop).

However, balancing this with full CSS display mode support is likely to be challenging.

### Proposal 2

Introduce a "line" preview mode when dragging between slot "areas" only (i.e. when dragging between parent containers).

This is what Figma does, and may mitigate layout shift when handling deeply nested layouts.

### Proposal 3

Make existing "area changed" debounce configurable, which determines how sticky a parent container is before Puck will try dragging into a smaller item.

chrisvxd:
Closes puckeditor/puck#1358, closes puckeditor/puck#1286

## Description

This PR changes the default drag-and-drop behaviour when dragging between slots to use a static placeholder with a line preview. When dragging within the same slot, the existing behaviour remains.

This is configurable via the `dnd.behavior` API, which can be `auto` (the default, changing based on reparenting), `fluid` (the original) or `static`.

## How to test

* Try dragging items between slots
* Set `?dndBehavior="fluid"` on the demo URL to restore the legacy behavior
* Set `?dndBehavior="static"` on the demo URL to enable static behavior for everything

In code, set the `dnd.behavior` property to change behavior:

```tsx
<Puck dnd={{ behaviour: "static" }} />
```

## Remaining considerations

1. I wasn't totally sure of the parameter name `behavior`, so if you have any suggestions, I'm all ears.

### DIFF
--- a/apps/docs/pages/docs/api-reference/components/puck.mdx
+++ b/apps/docs/pages/docs/api-reference/components/puck.mdx
@@ -123,9 +123,18 @@ Configure drag-and-drop behavior.
 
 #### dnd params
 
-| Param                                     | Example                   | Type    | Status |
-| ----------------------------------------- | ------------------------- | ------- | ------ |
-| [`disableAutoScroll`](#disableautoscroll) | `disableAutoScroll: true` | boolean | -      |
+| Param                                     | Example                   | Type                            | Status |
+| ----------------------------------------- | ------------------------- | ------------------------------- | ------ |
+| [`behavior`](#behavior)                   | `behavior: "static"`      | `"auto" \| "fluid" \| "static"` | -      |
+| [`disableAutoScroll`](#disableautoscroll) | `disableAutoScroll: true` | boolean                         | -      |
+
+##### `behavior`
+
+Control animation behavior while dragging. Defaults to `"auto"`.
+
+- `"auto"`: Use fluid behavior within a slot, then switch to a static when dragging an item between slots.
+- `"fluid"`: Move the item during a drag, animating other items out of the way.
+- `"static"`: Freeze the item in-place during a drag, showing a line placeholder at the drop destination.
 
 ##### `disableAutoScroll`
 
--- a/packages/core/components/DragDropContext/index.tsx
+++ b/packages/core/components/DragDropContext/index.tsx
@@ -30,6 +30,11 @@ import {
   ZoneStoreProvider,
 } from "../DropZone/context";
 import { createNestedDroppablePlugin } from "../../lib/dnd/NestedDroppablePlugin";
+import { prepareCommitFlip } from "../../lib/dnd/flip-commit";
+import {
+  resolveDndMode,
+  resolveOriginPreviewIndex,
+} from "../../lib/dnd/resolve-dnd-mode";
 import { insertComponent } from "../../lib/insert-component";
 import { moveComponent } from "../../lib/move-component";
 import { useDebouncedCallback } from "use-debounce";
@@ -44,6 +49,8 @@ import { useSafeId } from "../../lib/use-safe-id";
 import { getFrame } from "../../lib/get-frame";
 import { effect } from "@dnd-kit/state";
 import { scrollIntoView } from "../../lib/scroll-into-view";
+import type { DndBehavior } from "../../types";
+import { useLinePlaceholder } from "./use-line-placeholder";
 
 const DEBUG = false;
 
@@ -86,6 +93,7 @@ const AREA_CHANGE_DEBOUNCE_MS = 100;
 type DragDropContextProps = {
   children: ReactNode;
   disableAutoScroll?: boolean;
+  behavior?: DndBehavior;
 };
 
 /**
@@ -117,6 +125,7 @@ const useTempDisableFallback = (timeout: number) => {
 const DragDropContextClient = ({
   children,
   disableAutoScroll,
+  behavior = "auto",
 }: DragDropContextProps) => {
   const dispatch = useAppStore((s) => s.dispatch);
   const instanceId = useAppStore((s) => s.instanceId);
@@ -318,6 +327,14 @@ const DragDropContextClient = ({
 
   const initialSelector = useRef<{ zone: string; index: number }>(undefined);
 
+  const {
+    getTargetIndex: getLinePlaceholderTargetIndex,
+    setActive: setLinePlaceholderActive,
+    startScrollTracking: startLinePlaceholderScrollTracking,
+    stopScrollTracking: stopLinePlaceholderScrollTracking,
+    update: updateLinePlaceholder,
+  } = useLinePlaceholder(zoneStore);
+
   const nextContextValue = useMemo<DropZoneContext>(
     () => ({
       mode: "edit",
@@ -338,12 +355,15 @@ const DragDropContextClient = ({
         plugins={plugins}
         sensors={sensors}
         onDragEnd={(event, manager) => {
+          stopLinePlaceholderScrollTracking();
+
           const entryEl = getFrame()?.querySelector("[data-puck-entry]");
           entryEl?.removeAttribute("data-puck-dragging");
 
           const { source, target } = event.operation;
 
           if (!source) {
+            setLinePlaceholderActive(false);
             zoneStore.setState({ draggedItem: null });
 
             return;
@@ -353,12 +373,40 @@ const DragDropContextClient = ({
 
           const { previewIndex = {} } = zoneStore.getState() || {};
 
+          // Look the preview up by id: during line placeholder (cross-zone)
+          // drags the source stays in its original zone, so the preview key
+          // no longer matches the source's zone. Ghost previews only pin the
+          // item visually and never describe the drop position.
           const thisPreview: Preview | null =
-            previewIndex[zone]?.props.id === source.id
-              ? previewIndex[zone]
+            Object.values(previewIndex).find(
+              (preview) => preview?.props.id === source.id && !preview.ghost
+            ) ?? null;
+
+          // Capture sibling positions now, before dnd-kit tears the drag
+          // state down, so the slide animations measure from what's on
+          // screen at the moment of drop
+          const playCommitFlip =
+            !event.canceled &&
+            target?.type !== "void" &&
+            thisPreview?.linePlaceholder
+              ? prepareCommitFlip({
+                  zones: initialSelector.current
+                    ? [initialSelector.current.zone, thisPreview.zone]
+                    : [thisPreview.zone],
+                  itemId:
+                    thisPreview.type === "move"
+                      ? thisPreview.props.id
+                      : undefined,
+                  targetZone: thisPreview.zone,
+                  getExpectedOrder: () =>
+                    appStore.getState().state.indexes.zones[thisPreview.zone]
+                      ?.contentIds ?? [],
+                })
               : null;
 
           const onAnimationEnd = () => {
+            // Keep the ghost faded until the drop animation lands
+            setLinePlaceholderActive(false);
             zoneStore.setState({ draggedItem: null });
 
             // Tidy up cancellation
@@ -380,6 +428,18 @@ const DragDropContextClient = ({
               return;
             }
 
+            // Line placeholder indices count the dragged item at its
+            // original position, so moving to a later gap within the same
+            // zone lands one index lower once the item is removed
+            const commitIndex =
+              thisPreview &&
+              thisPreview.linePlaceholder &&
+              initialSelector.current &&
+              thisPreview.zone === initialSelector.current.zone &&
+              thisPreview.index > initialSelector.current.index
+                ? thisPreview.index - 1
+                : thisPreview?.index ?? index;
+
             // Finalise the drag
             if (thisPreview) {
               zoneStore.setState({ previewIndex: {} });
@@ -395,20 +455,24 @@ const DragDropContextClient = ({
                 moveComponent(
                   thisPreview.props.id,
                   initialSelector.current,
-                  thisPreview,
+                  { ...thisPreview, index: commitIndex },
                   appStore
                 );
               }
+
+              playCommitFlip?.();
             }
 
             const movedToNewPosition =
               initialSelector.current?.zone !== thisPreview?.zone ||
-              initialSelector.current?.index !== thisPreview?.index;
+              initialSelector.current?.index !== commitIndex;
 
             dispatch({
               type: "setUi",
               ui: {
-                itemSelector: { index, zone },
+                itemSelector: thisPreview
+                  ? { index: commitIndex, zone: thisPreview.zone }
+                  : { index, zone },
                 isDragging: false,
               },
               recordHistory: movedToNewPosition,
@@ -427,6 +491,16 @@ const DragDropContextClient = ({
               onAnimationEnd();
               dispose?.();
             }
+          });
+        }}
+        onDragMove={(event, manager) => {
+          // Keep the line in the gap nearest the pointer as it moves within
+          // the target zone: collisions only re-fire when the target
+          // changes, which can leave the line in a stale gap
+          updateLinePlaceholder(manager);
+
+          dragListeners.dragmove?.forEach((fn) => {
+            fn(event, manager);
           });
         }}
         onDragOver={(event, manager) => {
@@ -500,6 +574,17 @@ const DragDropContextClient = ({
           }
 
           if (dragMode.current === "new") {
+            const useLinePlaceholder =
+              resolveDndMode(behavior, { isNewComponent: true }) === "static";
+
+            if (useLinePlaceholder) {
+              targetIndex =
+                getLinePlaceholderTargetIndex(targetZone, manager) ??
+                targetIndex;
+            }
+
+            setLinePlaceholderActive(useLinePlaceholder);
+
             zoneStore.setState({
               previewIndex: {
                 [targetZone]: {
@@ -511,6 +596,7 @@ const DragDropContextClient = ({
                   props: {
                     id: source.id.toString(),
                   },
+                  linePlaceholder: useLinePlaceholder,
                 },
               },
             });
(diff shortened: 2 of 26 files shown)


packages/core/components/DropZone/LinePlaceholder.tsx
```diff
@@ -0,0 +1,139 @@
+import { CSSProperties, RefObject, useLayoutEffect, useState } from "react";
+import { getClassNameFactory } from "../../lib";
+import type { DragAxis } from "../../types";
+import styles from "./styles.module.css";
+
+const getClassName = getClassNameFactory("DropZone", styles);
+const LINE_PLACEHOLDER_SIZE = 4;
```
FedericoBonel, line 7: We need to theme this, in the outline we use `var(--puck-border-width-focus)`, so we should probably decide if we want to use that or another theme property. We might want a slot level override for this as user can easily want to keep this size while overriding the theming property everywhere else.

packages/core/components/DropZone/LinePlaceholder.tsx
```diff
@@ -0,0 +1,139 @@
...
+      // gap or sibling margin; in empty zones, start after the zone padding.
+      const y = next
+        ? prev && prev.rect.bottom <= next.rect.top
+          ? (prev.rect.bottom + next.rect.top) / 2
+          : next.rect.top -
+            Math.max(px(styleOf(next.el)?.marginTop), rowGap) / 2
+        : prev
+        ? prev.rect.bottom +
+          Math.max(px(styleOf(prev.el)?.marginBottom), rowGap) / 2
+        : zoneRect.top + px(zoneStyle?.paddingTop);
```
FedericoBonel, line 76: Can we break these calculations into functions or clarify them a bit with `if` statements? Is a bit hard to read and know what is going on here. 
I do get what's happening (if between two elements, if before an element, if after an element, else in an empty zone) but it takes a while and I can see myself or someone else wasting a lot of time going through these. We also might need to add additional logic in the future if we're missing some edge cases

packages/core/components/DropZone/styles.module.css
```diff
@@ -40,6 +40,17 @@
   position: relative;
 }
 
+.DropZone-linePlaceholder {
+  background: var(
+    --puck-slot-component-color-placeholder,
+    var(--puck-color-azure-06)
```
FedericoBonel, line 46: We now got a new `--puck-color-insertion-indicator` variable we can use here as the default.

packages/core/components/DropZone/styles.module.css
```diff
@@ -40,6 +40,17 @@
...
+  );
+  border-radius: 2px;
```
FedericoBonel, line 48: In the outline I use the var I [mentioned above / 2](https://github.com/puckeditor/puck/blob/main/packages/core/components/LayerTree/components/drop-line/styles.module.css#L4) which matches this if the user increases it.

I think we should do the same here

packages/core/components/Puck/components/Layout/index.tsx
FedericoBonel, line 77: dnd is now also available in the app store to select only what you need from it (little optimization I did for the styling of each row in the outline), you might want to use that if you want.

packages/core/components/Puck/index.tsx
```diff
@@ -85,6 +86,14 @@ type PuckProps<
   iframe?: IframeConfig;
   dnd?: {
     disableAutoScroll?: boolean;
+    /**
+     * - `auto` (default): fluid drags within a slot, switching to a static
+     *   line placeholder when dragging between slots or inserting a new item
+     * - `fluid`: always animate sibling items during a drag
+     * - `static`: always show a line placeholder during a drag, only
+     *   animating sibling items on drop
+     */
```
FedericoBonel, line 95: Side note but since we're incrementally adding these we should eventually have them everywhere

apps/docs/pages/docs/api-reference/components/puck.mdx
```diff
@@ -123,9 +123,18 @@ Configure drag-and-drop behavior.
...
+| [`behavior`](#behavior)                   | `behavior: "static"`      | `"auto" \| "fluid" \| "static"` | -      |
+| [`disableAutoScroll`](#disableautoscroll) | `disableAutoScroll: true` | boolean                         | -      |
+
+##### `behavior`
+
+Control animation behavior while dragging. Defaults to `"auto"`.
+
+- `"auto"`: Use fluid behavior within a slot, then switch to a static when dragging an item between slots.
+- `"fluid"`: Move the item during a drag, animating other items out of the way.
+- `"static"`: Freeze the item in-place during a drag, showing a line placeholder at the drop destination.
```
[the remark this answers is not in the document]
FedericoBonel: This is worth fixing

packages/core/components/DropZone/LinePlaceholder.tsx
```diff
@@ -0,0 +1,139 @@
+import { CSSProperties, RefObject, useLayoutEffect, useState } from "react";
+import { getClassNameFactory } from "../../lib";
+import type { DragAxis } from "../../types";
+import styles from "./styles.module.css";
+
+const getClassName = getClassNameFactory("DropZone", styles);
+const LINE_PLACEHOLDER_SIZE = 4;
+
+const clamp = (value: number, min: number, max: number) =>
```
FedericoBonel, line 9: Do we have other utilities we can use here? If not we would do good to extract this to something like `lib/math` or similar for future reusabilitiy

packages/core/components/DropZone/LinePlaceholder.tsx
```diff
@@ -0,0 +1,139 @@
...
+          ? (prev.rect.bottom + next.rect.top) / 2
+          : next.rect.top -
+            Math.max(px(styleOf(next.el)?.marginTop), rowGap) / 2
+        : prev
+        ? prev.rect.bottom +
+          Math.max(px(styleOf(prev.el)?.marginBottom), rowGap) / 2
+        : zoneRect.top + px(zoneStyle?.paddingTop);
+
+      setStyle({
+        left:
```
FedericoBonel, line 79: Same with the ones below here 👇, I think they need a bit of breaking down or assignment to a more intent clear variable

packages/core/lib/dnd/nearest-gap.ts
```diff
@@ -0,0 +1,159 @@
...
+
+  const candidates: GapCandidate[] = [];
+
+  const edgeCandidate = (
+    index: number,
+    rect: DOMRect,
+    side: "before" | "after"
+  ) => {
+    if (horizontalFlow) {
+      const x = side === "before" ? rect.left : rect.right;
```
FedericoBonel, line 89: `dir=rtl` and `row-reverse` are currently broken, they don't show the line preview correctly, and I believe this is why, but there might be more places where this is actually breaking as well. Basically we can't assume `before` means `rect.left`. 

I actually added a utility to map directions to `before` or `after` based on the direction in the document: https://github.com/puckeditor/puck/blob/main/packages/core/lib/dnd/get-insert-index.ts#L9-L17

I suggest we use that or a similar approach

packages/core/components/DropZone/index.tsx
```diff
@@ -444,7 +445,7 @@ export const DropZoneEdit = forwardRef<HTMLDivElement, DropZoneProps>(
     const isDropEnabled =
       isEnabled &&
       (preview
-        ? contentIdsWithPreview.length === 1
+        ? contentIdsWithPreview.length === (preview.linePlaceholder ? 0 : 1)
         : contentIdsWithPreview.length === 0);
```
FedericoBonel, line 449: I find this check a bit confusing I suggest we use an if or a helper function

packages/core/lib/dnd/resolve-dnd-mode.ts
```diff
@@ -0,0 +1,33 @@
...
+export const resolveOriginPreviewIndex = (
+  initialIndex: number,
+  preview?: { index: number; linePlaceholder?: boolean } | null
+) => {
+  if (!preview || preview.linePlaceholder) {
+    return initialIndex;
+  }
+
+  return preview.index;
+};
```
FedericoBonel, line 33: Not sure this belongs to this file

packages/core/lib/dnd/nearest-gap.ts
```diff
@@ -0,0 +1,159 @@
...
+ * Returns the insertion index whose gap is nearest to the pointer, using the
+ * same geometry as the line placeholder indicator. This is more predictable
+ * than directional midpoint collisions for cross-zone drags, where items
+ * don't move out of the way and the drag shape's size is unrelated to the
+ * target zone's items.
+ *
+ * Rendered items are mapped to their real content indices via `contentIds`:
+ * virtualized zones only render a window of their content, so DOM positions
+ * can't be used as indices directly.
+ */
```
FedericoBonel, line 52: This seems a bit verbose and somewhat AIfied, it doesn't really explain what the function does, but why it created it when it added it.

packages/core/lib/dnd/nearest-gap.ts
```diff
@@ -0,0 +1,159 @@
...
+    zoneEl.querySelectorAll(
+      ":scope > [data-puck-component]:not([data-dnd-dragging]):not([data-dnd-placeholder])"
+    )
+  )
+    .map((el) => ({
+      index: contentIds.indexOf(el.getAttribute("data-puck-component") ?? ""),
+      rect: el.getBoundingClientRect(),
+    }))
+    .filter((item) => item.index !== -1)
+    .sort((a, b) => a.index - b.index);
```
FedericoBonel, line 72: This can be at worst O(n^2) on every drag.

Can we build a map of ids to indexes to do a fast look up before this? That would bring it down to O(n logn) since you don't have to do `indexOf`.

I know it seems like an over optimization, but is a cheap thing that is not really hurting it at all and we have had performance issues before. 

Though the individual element measurement for each item is probably hurting it more.

Once you create a map you can also move the `filter` before the `map` to avoid measuring every single child and instead you measure what you need.

Some of this stuff is also cacheable per drag as things don't swap around in static mode at least, I did do some caching in [my pr for some of the computations](https://github.com/puckeditor/puck/commit/db035ce57a94425da9d0e82d3f1c06ecabd4c350#diff-779d979d8e8c75a0ad8e0b7c40f16bdedac5fe6da79d55c13e65d40cb7579008R31-R46), so wondering if that's useful at all.

packages/core/lib/dnd/nearest-gap.ts
```diff
@@ -0,0 +1,159 @@
...
+
+  if (rendered.length === 0) return 0;
+
+  const style = win.getComputedStyle(zoneEl);
+  const horizontalFlow =
+    style.display === "grid" ||
+    (style.display === "flex" && style.flexDirection.startsWith("row"));
```
FedericoBonel, line 79: I think we should also check for `inline-grid` and `inline-flex`

packages/core/lib/dnd/drop-animation.ts
```diff
@@ -0,0 +1,246 @@
...
+}: {
+  feedbackElement: HTMLElement;
+  itemId?: string;
+  targetZone: string;
+  getExpectedOrder: () => string[];
+}) => {
+  const overlayDoc = feedbackElement.ownerDocument;
+  const targetDoc = getFrame() ?? overlayDoc;
+
+  if (prefersReducedMotion(overlayDoc)) return;
```
FedericoBonel, line 132: Should we also do this on `runFallbackDropAnimation`?

packages/core/lib/dnd/commit-animation.ts
```diff
@@ -0,0 +1,82 @@
...
+    targetZone,
+    getExpectedOrder,
+    initialExpectedOrder = [],
+  }: WaitForCommitOptions,
+  callback: (committed: HTMLElement | null) => void
+) => {
+  let attempts = 0;
+
+  const tick = () => {
+    const zoneEl = doc.querySelector(`[data-puck-dropzone="${targetZone}"]`);
```
FedericoBonel, line 39: I think we do this in multiple places, but I also saw you correctly used `CSS.scape` (which I didn't know about). I think we should extract the creation of this string to a central util and re-use that everywhere for consistency

packages/core/lib/dnd/drop-animation.ts
```diff
@@ -0,0 +1,246 @@
...
+    transition: "none",
+    translate: "none",
+    zIndex: "2147483647",
+  });
+
+  const hideStyle = targetDoc.createElement("style");
+  hideStyle.textContent = `
+    ${
+      itemId
+        ? `[data-puck-component="${itemId}"] { visibility: hidden !important; }`
```
FedericoBonel, line 164: Same thing here for the id injection

packages/core/lib/dnd/commit-animation.ts
```diff
@@ -0,0 +1,82 @@
...
+    const expected = getExpectedOrder();
+    const committedItemId =
+      itemId ?? expected.find((id) => !initialExpectedOrder.includes(id));
```
FedericoBonel, line 42: This and the includes below can be avoided if we drop `initialExpectedOrder` and `rendered` into a `Set` before this

packages/core/lib/dnd/nearest-gap.ts
```diff
@@ -0,0 +1,159 @@
...
+  }
+
+  const rect = frameEl.getBoundingClientRect();
+  const scale = rect.width / (frameEl.contentWindow?.innerWidth || 1);
+
+  return {
+    x: (position.x - rect.left) / scale,
+    y: (position.y - rect.top) / scale,
+  };
+};
```
FedericoBonel, line 40: This is quite similar to the [`GlobalPosition` class](https://github.com/puckeditor/puck/blob/db035ce57a94425da9d0e82d3f1c06ecabd4c350/packages/core/lib/global-position.ts), wonder if maybe there's a shared abstraction we can use.

Also this doesn't seem to belong to this file.

packages/core/components/DragDropContext/use-line-placeholder.ts
```diff
@@ -0,0 +1,148 @@
...
+      const zoneEl = getFrame()?.querySelector(
+        `[data-puck-dropzone="${zone}"]`
+      );
+
+      if (!zoneEl) return null;
+
+      return getTargetIndex(
+        zoneEl,
+        manager,
+        appStore.getState().state.indexes.zones[zone]?.contentIds ?? []
```
FedericoBonel, line 46: We do this access a few times in the PR `state.indexes.zones[zone]?.contentIds ?? [] ` I wonder if a shared `getZoneContentIds` with automatic defaulting would be useful for this and other parts of the codebase.

packages/core/components/Drawer/index.tsx
```diff
@@ -72,11 +84,36 @@ const DrawerItemDraggable = ({
...
+        preview
+          ? {
+              targetZone: preview.zone,
+              getExpectedOrder: () =>
+                appStore.getState().state.indexes.zones[preview.zone]
+                  ?.contentIds ?? [],
+            }
+          : undefined
+      );
+    },
```
FedericoBonel, line 107: This and the `DraggableComponent` one are very similar, maybe worth extracting into a hook or function? But unsure.

packages/core/components/DragDropContext/index.tsx
```diff
@@ -380,6 +428,18 @@ const DragDropContextClient = ({
...
+            // original position, so moving to a later gap within the same
+            // zone lands one index lower once the item is removed
+            const commitIndex =
+              thisPreview &&
+              thisPreview.linePlaceholder &&
+              initialSelector.current &&
+              thisPreview.zone === initialSelector.current.zone &&
+              thisPreview.index > initialSelector.current.index
+                ? thisPreview.index - 1
+                : thisPreview?.index ?? index;
```
FedericoBonel, line 441: There's a new `lib/dnd/get-insert-index.ts` that implements this, so I think we should re-use it here.

packages/core/lib/dnd/nearest-gap.ts
FedericoBonel: I also noticed single column grids and column flex containers show a vertical preview instead of a horizontal one, which might also come from this

packages/core/lib/dnd/nearest-gap.ts
```diff
@@ -0,0 +1,159 @@
...
+    } else if (!prev) {
+      // Before the first rendered item
+      edgeCandidate(next.index, next.rect, "before");
+    } else if (next.index - prev.index > 1) {
+      // The items between these two are virtualized out: expose both edges
+      // as separate insertion points
+      edgeCandidate(prev.index + 1, prev.rect, "after");
+      edgeCandidate(next.index, next.rect, "before");
+    } else if (horizontalFlow) {
+      const wraps = prev.rect.right > next.rect.left;
```
FedericoBonel, line 115: I think this also is related to the rtl problem

packages/core/components/DropZone/LinePlaceholder.tsx
```diff
@@ -0,0 +1,139 @@
...
+    };
+
+    const zoneRect = zoneEl.getBoundingClientRect();
+    const zoneStyle = styleOf(zoneEl);
+    const prev = getItem(index - 1);
+    const next = getItem(index);
+    const closest = next ?? prev;
+    const size = LINE_PLACEHOLDER_SIZE;
+
+    if (dragAxis === "y") {
```
FedericoBonel, line 63: Also I just realized, this draws the preview, but `getNearestGapIndex` calculates it for other usages. However this determines the flow by using `dragAxis`, and `getNearestGapIndex` determines it by checking the styles underneath. 

Since both are different we could potentially hit a case where there's a mistmatch. I suggest we try to unify this, and all of the flow resolutions so that we can easily handle rtl as well

packages/core/components/DragDropContext/index.tsx
```diff
@@ -548,20 +672,29 @@ const DragDropContextClient = ({
           });
         }}
         onDragStart={(event, manager) => {
+          // Scrolling moves the content under a stationary pointer without
+          // firing drag events, which would leave the line placeholder
+          // drifting with the page; recompute it on scroll.
+          startLinePlaceholderScrollTracking(manager);
```
FedericoBonel, line 678: Maybe we call this when behavior is `auto` or `static` since it'll attach a listener to the scroll event for `fluid` as well just to abort when it fires.

FedericoBonel (changes requested):
This is great! I love it. Left some comments and thoughts. Addtionally:

- I think we need the action bar drag handle if possible

chrisvxd:
Closed in favour of #1783

PR #1742, opened 2026-07-15
fix: don't drop transformed fields on updates

### LINKED ISSUE #1740: v0.22.1 regression: inline richtext fields render empty after config/data re-walk (fallout from #1342 field-transform fix)
### Description

After upgrading `@puckeditor/core` from **0.22.0 to 0.22.1**, every inline `richtext` field in our editor renders **empty on the canvas** after the first save in our app. The underlying data is intact (verified in our database and by reloading the page — the content comes back). Downgrading to 0.22.0 with no other change fixes it, so we bisected the regression to 0.22.1.

We believe the culprit is the fix for puckeditor/puck#1342 (`fix: run field transforms for fields without default props`).

### Environment

* `@puckeditor/core` 0.22.1
* React 19.2.7 / React DOM 19.2.7
* Field config on the affected component:

```tsx
fields: {
  bodyRichtextHtml: {
    type: "richtext",
    contentEditable: true,
    options: { heading: false, code: false, horizontalRule: false, blockquote: false },
  },
},
defaultProps: {
  bodyRichtextHtml: "<p>Write your paragraph here.</p>",
},
```

### Reproduction

1. `<Puck config={config} data={data} onChange={...} />` where `config` is rebuilt (new object identity) when our app's "active version" object changes — e.g. after we persist the draft to our server and store the response.
2. Type into an inline contentEditable richtext block. Content renders fine.
3. Save (our app swaps the active-version object → the `config` prop gets a new identity → Puck re-walks the tree).
4. **Actual:** every richtext block on the canvas is now blank. The data emitted by `onChange` and persisted is still correct; a full page reload shows the content again.
5. **Expected:** the canvas keeps rendering the richtext values, as it did on 0.22.0.

### Analysis

Diffing the published dists of 0.22.0 vs 0.22.1, the only editor-behavior change is in the field-transform walker (`walkObject`):

```js
// new in 0.22.1
const keys = Object.keys(value);
if (ownedFields) {
  for (const fieldName in fields) {
    const fieldType = fields[fieldName].type;
    if (fieldType !== "slot" && mappers[fieldType] && !(fieldName in value)) {
      keys.push(fieldName);
    }
  }
}
const newProps = keys.map((k) => ({ ..., value: value[k] })); // value[k] === undefined for the pushed keys
```

Transforms now also run for configured mapper-typed fields that are missing from the walked object, with `value === undefined`. On the re-walk triggered in step 3, the built-in `richtext` transform ends up receiving `undefined` and the inline tiptap editor initializes empty, replacing the previously rendered content. On 0.22.0 this branch didn't exist, so the re-walk left the rendered value alone.

(Analysis is from reading the minified dists, so the exact internal path may differ — but the A/B result between 0.22.0 and 0.22.1 is clean and reproducible on our side.)

### Workaround

Pin `@puckeditor/core` to exactly `0.22.0`.

FedericoBonel:
Closes puckeditor/puck#1740

## Description

This PR fixes an issue where field transforms where being set to undefined after updating a component.

The problem was that `mapFields` was getting called from the tracked version of `useFieldTransforms` with partial updates in the component. With the fix we did on `mapFields` this was setting any non set prop that was defined in the fields as undefined.

## Changes made

* Added a param to control which fields to map in `mapFields`.
* Updated use field transforms tracked to use this new param.
* Added tests to cover the new map fields prop.
* Added tests to cover use field transforms tracked.

## How to test

* Render the `Puck` component with the following config:

```tsx
export const config = {
  components: {
    HeadingBlock: {
      fields: {
        text: { type: "text", contentEditable: true },
        someNumber: { type: "number" },
      },
      defaultProps: {
        text: "Hello",
        someNumber: 0,
      },
      render: ({ text, someNumber }) => (
        <div style={{ padding: 64 }}>
          <h1>{text}</h1>
          <p>{someNumber}</p>
        </div>
      ),
    },
  },
};
```

* Update the text field
* Update the number field
* Observe the text field not being deleted from the canvas

### DIFF
--- a/packages/core/lib/field-transforms/use-field-transforms-tracked.tsx
+++ b/packages/core/lib/field-transforms/use-field-transforms-tracked.tsx
@@ -16,6 +16,7 @@ export function useFieldTransformsTracked<
   readOnly?: T["readOnly"],
   forceReadOnly?: boolean
 ): T["props"] {
+  const prevMappers = useRef<Mappers>(null);
   const prevProps = useRef<Record<string, any>>(null);
   const prevResult = useRef<Record<string, any>>(item.props);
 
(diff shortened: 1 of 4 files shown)


packages/core/lib/field-transforms/use-field-transforms-tracked.tsx
```diff
@@ -25,43 +26,57 @@ export function useFieldTransformsTracked<
...
       item.type === "root" ? config.root : config.components?.[item.type];
 
-    let changeIncludesSlot = false;
+    const componentFields = componentConfig?.fields ?? {};
 
-    for (const fieldName in item.props) {
-      const fieldType = componentConfig?.fields?.[fieldName]?.type;
+    // Transformed props depend on mappers, so a mappers change
+    // invalidates all of them (e.g when readOnly changes)
+    const fullRemap = prevMappers.current !== mappers;
```
FedericoBonel, line 36: This was needed because we had a stale closure issue where if the readonly updated or transforms we were not re running them

chrisvxd (approved):

Issue #1710, opened 2026-06-23
Add a configurable dwell delay before sortable reordering (ease nesting into containers)


levi-hrb:
## Description

When dragging a component in the editor — whether inserting a new one from the drawer or moving an existing one — the surrounding components immediately shift to "make space" the moment a collision is detected. For simple top-level reordering this feels great. But it makes dropping a component into a nested container painful: as you move the cursor toward a container's drop zone, the components around it are continuously reflowing, so the target you're aiming for keeps sliding out from under the pointer before you can release. 

The desired outcome is to be able to freely move a dragged component over the canvas without anything reflowing, and to have the surrounding components make space only once the user deliberately pauses over a position for a short, configurable moment. In other words: moving = nothing shifts (so you can aim into a container and drop), pausing = the layout opens up to accept a reorder. This would make nesting into parent containers far more predictable while keeping the current "make space" feedback for users who do want it.

## Considerations

This is closely related to the nesting/collision constraint described in #123 ("your mouse must be directly over a DropZone for a collision to be detected") — the constant reflow is what makes satisfying that constraint hard in practice.
The relevant logic lives in the dynamic collision detector introduced with the 0.18 DnD engine: createDynamicCollisionDetector (in core, under lib/dnd/collision/dynamic/), which is passed as the hardcoded collisionDetector into the sortable. It isn't currently injectable or configurable, so this behavior can't be achieved without patching internals.
Existing public DnD knobs don't address timing: collisionAxis on <DropZone> only constrains collision direction, and allow/disallow only restrict what may be dropped — neither affects when the reflow happens.
The only existing dnd option on <Puck> today is disableAutoScroll, so there's a natural place to add this.
Tested on Puck 0.20.2.

### Proposal 1 — Configurable dwell delay on <Puck dnd={...}>
Add an opt-in dwell delay that gates when the sortable commits a reorder, exposed through the existing dnd prop:

```
<Puck
  dnd={{
    // ms the pointer must stay (roughly) still before a sortable reorder commits.
    // 0 (default) preserves today's immediate "make space" behavior.
    reorderDelay: 500,
    // optional: pointer movement (px) that resets the dwell timer.
    reorderDelayMovementThreshold: 6,
  }}
/>
```

[...]

Issue #1486, opened 2026-01-07
Drag and drop: detach hit area from slots


FedericoBonel:
## Description

There are multiple use cases where a target slot needs to appear invisible or not be displayed on the page, while still remaining targetable. Alternatively, the target area may need to be displayed differently from the default block with a fixed minimum height.

With the current implementation this isn't really possible.

## Workaround

A possible workaround is to use a slot to wrap all the content, then disable the drag permissions for components that should remain static on the page. This allows the parent slot to wrap around the content (effectively becoming invisible), but still grow when targeted.

## Considerations

- This may be somewhat related to #1358.

## Proposals

Similar to overlay portals, we could provide a hook to register hit areas using a ref:

```tsx
import { useRegisterHitArea } from "@measured/puck";

const Example = ({ Content }) => {
  const ref = useRef(null);

  useRegisterHitArea(ref, "Content");

  return (
    <div ref={ref}>
      <h1>I'm in the hit area</h1>
      <Content />
    </div>
  );
};
```

In this scenario, the `Content` would simply render the content. It could collapse by default and then expand when hovered over.

### Considerations

- The sorting and reordering could be problematic or become hard to use, we should carefully study what the expected interaction would be for something like this.

```tsx
<div ref={ref}>
  <h1>I'm in the hit area, what happens with my position when I'm hovered over?</h1>
  <Content />
  <h1>I'm in the hit area too, what happens with my position when I'm hovered over?</h1>
  <NestedContent />
</div>
```

- Colliding hit areas could also be a problem if we use multiple nested slots.

PR #1754, opened 2026-07-19
fix: allow edge escape from components containing an active zone

edwardteach42:
## Description

Dropping a component next to a container that accepts it doesn't work. The most visible case is dragging a container while another container is on the page: the insert preview dives into the deepest zone under the pointer and never comes back out, so everything keeps nesting.

The cause is in `NestedDroppablePlugin.ts`. `getZoneId` returns null for any component candidate whose zone accepts the dragged type, which during a container drag is every container on the page. Null freezes zone targeting, so the parent zone can never be reached from a container's edge band. That band is exactly the region the `BUFFER` contraction in `getPointerCollisions` reserves for falling through to the parent, so the two mechanisms cancel each other out.

The freeze itself is worth keeping for interior hovers, since it stops the target from bouncing to the parent while hovering a component's chrome around its slot. This PR keeps that and only allows the pointer to escape to the parent zone within an edge margin of the component's box (2x `BUFFER`, clamped so small components keep a stable interior).

No public API changes.

## Changes made

- `findDeepestCandidate` lets a null zone resolve to the component's own
  zone (its parent) when the pointer is within an edge margin of the component's bounding box.
- `area` now follows the same resolution: a frozen zone highlights the
  hosting component, a resolved zone follows `areaId`. Previously these could disagree.

## How to test

- Add a slot-based Container component to a config:

```tsx
const config = {
  components: {
    Container: {
      fields: { content: { type: 'slot' } },
        <div style={{ padding: 10 }}>
          <Content />
        </div>
      ),
    },
  },
};
```

- Drop one Container onto the canvas, then drag a second one from the
  drawer and hover near the first one's edge. Before this change the preview gets stuck inside the first container and the drop nests. After it, the edge shows the sibling insert position and the drop lands next to it.
- Interior behavior is unchanged: hovering the middle of a container
  still nests into it, and reordering inside a container doesn't pop items out unless you drag them to the edge.

Related: #1710, #123, #1486 (not fully fixed by this, but this is the mechanism behind much of the pain described there).

### DIFF
--- a/packages/core/lib/dnd/NestedDroppablePlugin.ts
+++ b/packages/core/lib/dnd/NestedDroppablePlugin.ts
@@ -21,7 +21,7 @@ type NestedDroppablePluginOptions = {
       area: string | null;
       zone: string | null;
     },
-    manager: DragDropManager
+    manager: DragDropManager,
   ) => void;
 };
 
@@ -67,17 +67,17 @@ const BUFFER = 6;
 
 const getPointerCollisions = (
   position: GlobalPosition,
-  manager: DragDropManager
+  manager: DragDropManager,
 ) => {
   const candidates: Droppable[] = [];
 
   let elements = position.target.ownerDocument.elementsFromPoint(
     position.x,
-    position.y
+    position.y,
   );
 
   const previewFrame = elements.find((el) =>
-    el.getAttribute("data-puck-preview")
+    el.getAttribute("data-puck-preview"),
   );
 
   // Restrict to drawer element if pointer is over drawer. This is necessary if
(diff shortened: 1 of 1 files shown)


Anubhab2003 (commented):
Proceed for merging

FedericoBonel:
Hello @edwardteach42 and @Anubhab2003! Thanks for the contribution 🙏

We require an issue to be opened first, so I'll close this PR for now. Feel free to reopen it after adding a bug report with a minimal reproduction. This will help us determine whether the problem is on our end or within the integration, and whether this is the correct fix.

Thank you again!

PR #1759, opened 2026-07-23
fix: don't call generate-id at module import time

### LINKED ISSUE #1602: @puckeditor/core calling `createAppStore` at module top-level causing Cloudflare Workers and other edge runtimes error
@puckeditor/core crashes on Cloudflare Workers (and likely other edge runtimes such as Vercel Edge, Deno Deploy) because createAppStore is called at the module level inside chunk-UMTU6EA7.mjs. This triggers generateId() → uuid v4 → crypto.getRandomValues() during module initialization, before any request handler runs.
Cloudflare Workers explicitly disallow crypto.getRandomValues() in the global scope:

Asynchronous I/O (ex: fetch() or connect()), setting a timeout, and generating random values are not allowed within global scope.

This means it is impossible to import anything from @puckeditor/core — including the SSR-safe Render component — in a Cloudflare Worker, even if the editor Puck component is never rendered.

Steps to Reproduce

1. Create a Cloudflare Workers project (e.g. using Hono + Vite)
2. Install @puckeditor/core
3. Import Render from @puckeditor/core anywhere in your server entry file
4. Deploy or run locally with wrangler dev

```
Uncaught Error: Disallowed operation called within global scope.
Asynchronous I/O (ex: fetch() or connect()), setting a timeout, and 
generating random values are not allowed within global scope. To fix 
this error, perform this operation within a handler.
https://developers.cloudflare.com/workers/runtime-apis/handlers/

    at v4 (_worker.js:21915:27)
    at generateId (_worker.js:22357:86)
    at _worker.js:29547:23
    at createStoreImpl (_worker.js:23543:36)
    at createStore (_worker.js:23546:74)
    at createImpl (_worker.js:23570:19)
    at createAppStore (_worker.js:29543:74)
    at _worker.js:29728:56 in
  node_modules/@puckeditor/core/dist/chunk-UMTU6EA7.mjs

FedericoBonel:
Closes puckeditor/puck#1602

## Description

This PR fixes an issue where importing from `@puckeditor/core` would break in Cloudflare Workers and other edge runtimes.

The problem was that we were calling `createAppStore` at the module level. This invoked `generateId`, which then invoked `uuid` and used `crypto.getRandomValues`, which is not supported in edge runtimes. This is fine for the editor, but it is a problem when rendering at the edge.

## Changes made

* The `appStore` context is now initialized with a null value and lazily initialized when needed.
  * Previously, we always initialized it with a default store, but that called `generateId`, which triggered the chain described above. In the editor (`Puck`), the `appStoreContext` is initialized before being provided to the children at render time, so this change does not affect it. In `Render`, it is not provided and should never be used in the first place, so this change does not affect it either.

### DIFF
--- a/packages/core/__tests__/edge-runtime.spec.ts
+++ b/packages/core/__tests__/edge-runtime.spec.ts
@@ -1,0 +1,25 @@
+/** Tests package import side effects */
+class ResizeObserverMock {
+  observe() {}
+  unobserve() {}
+  disconnect() {}
+}
+(globalThis as any).ResizeObserver = ResizeObserverMock;
+
+describe("store edge runtime safety", () => {
+  it("does not call generateId at package import time", async () => {
+    jest.mock("../lib/generate-id", () => ({
+      generateId: jest.fn(() => {
+        return "test-id";
+      }),
+    }));
+
+    const { generateId } = await import("../lib/generate-id");
+
+    // The package import must NOT call generateId() because it uses crypto.randomUUID()
+    // under the hood, which is not available in edge runtimes.
+    await import("../index");
+
+    expect(generateId).not.toHaveBeenCalled();
+  });
+});
--- a/packages/core/components/Puck/__tests__/index.tsx
+++ b/packages/core/components/Puck/__tests__/index.tsx
@@ -87,7 +87,7 @@ describe("Puck", () => {
   const componentBRender = jest.fn(() => null);
   const rootRender = jest.fn(() => null);
 
-  const config: Config = {
+  const config: Config<{}> = {
     root: {
       render: ({ children }) => {
         rootRender();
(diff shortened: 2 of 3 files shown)


chrisvxd (commented):
This may appear to work for a single Puck instance, but you've essentially made the App Store global - the context is no longer doing anything.

This means that multiple instances of Puck on the same page will likely share an app store. Normally this manifests in the docs, but it's still working so I'm not totally sure why.

I think I might be missing how this works, or perhaps this is not the correct fix.

FedericoBonel:
Hm, I'm not sure I follow.

In the `Puck` component, the id was already generated at render time (which it still does):

[puck/packages/core/components/Puck/index.tsx at fede/puck-465-puckeditorcore-calling-createappstore-at-module-top-level · puckeditor/puck](https://github.com/puckeditor/puck/blob/fede/puck-465-puckeditorcore-calling-createappstore-at-module-top-level/packages/core/components/Puck/index.tsx#L295-L297)

The default would only apply where we don't provide the `appStore.Provider`, which I think is only in the `Render` component when rendering client side. When that happens the dropzone renderers use the hook to access the context but don't use the value, so we default it to preserve the same behavior we had before.

Does this make sense?

Maybe I'm still missing something

chrisvxd (approved):

PR #1770, opened 2026-07-29
feat: add optional drag handle to component action bar

### LINKED ISSUE #1477: Add a drag handle to the action bar
## Description

When using slots, there are cases where the slot fits its height and width to the nested content, making it difficult to select and drag when it contains nested children since it wraps around them tightly.  
Using dynamic padding in the editor can help, but it may cause the preview UI to feel disconnected from the actual rendered layout, leading users to design pages that look different when published.

To address this, we could consider adding a drag handle to the action bar so users can drag components by grabbing the handle.

## Considerations

- Users can override the action bar, so we should probably expose the handle as an additional prop, similar to how the `select parent` action is handled.

## Proposal

Add a drag handle icon to the action bar that allows users to drag and move the entire component.

dkrasniy:
Closes #1477

 ## Description

This PR adds an opt-in **drag handle** to a component's action bar. This is especially useful for slots that wrap tightly around nested content and are awkward to grab directly (#1477).

The handle is **off by default** and enabled via the existing `dnd` prop: `<Puck dnd={{ enableDragHandle: true }} />`.

Following the consideration in the issue, the handle is exposed to custom action bars as an additional `dragHandle` prop on the`actionBar` override — mirroring how `parentAction` (select parent) is passed — so consumers who override the action bar can place or omit it.

 **Non-breaking:** whole-component-body dragging still works exactly as before. 

Screenshot 2026-07-29 at 3 06 28 PM

 ## Changes made
- `dnd.enableDragHandle` — new opt-in flag, default `false`.
- `ActionBar.DragHandle` — new export. A grip button that forwards its ref.
- `overrides.actionBar` — now gets an optional `dragHandle` prop. Existing overrides are unaffected.
- `DraggableComponent` — renders the handle at the start of the action bar when `enableDragHandle && permissions.drag`.
- `useSensors` / `DragDropContext` — register both the body and the handle as drag activators.
- Added the `action-drag` dictionary key (`"Drag to reorder"`), docs, and tests.

## Known behavior

The action bar (and therefore the handle) is unmounted while the component it belongs to is being dragged. This is existing behavior from the `dragFinished` gate in `DraggableComponent`, not something this PR changes. The document-level `grabbing` cursor persists through the drag and the bar reappears at the dropped position. Making the bar track the drag would mean positioning it against dnd-kit's clone rather than the source element, which felt out of scope here.

  ## How to test
  1. In `apps/demo`, enable the handle:

     ```tsx
     <Puck dnd={{ enableDragHandle: true }} config={config} data={data} />
     ```

  2. Select a component — a grip handle appears at the left of the action bar. Grab it and drag to reorder the component. It should start dragging immediately, with no press-and-hold delay.
  3. Confirm existing behavior is intact: dragging the component body still reorders it (handle is additive, not a replacement),
  and still requires the short press-and-hold.
  4. Set `enableDragHandle: false` (or omit `dnd`) — no handle renders, behavior is unchanged from `main`.
  5. Set `permissions={{ drag: false }}` — the handle does not render.
  6. Override the action bar and place `dragHandle` yourself to confirm the prop is exposed:

     ```tsx
     <Puck
       dnd={{ enableDragHandle: true }}
       overrides={{
         actionBar: ({ label, children, parentAction, dragHandle }) => (
           <ActionBar label={label}>
             <ActionBar.Group>{dragHandle}</ActionBar.Group>
             <ActionBar.Group>{parentAction}</ActionBar.Group>
             <ActionBar.Group>{children}</ActionBar.Group>
           </ActionBar>
         ),
       }}
     />
     ```

## Summary by CodeRabbit

* **New Features**
  * Added an optional drag handle for reordering components.
  * Dragging can start from the handle while preserving body-dragging behavior.
  * Added accessibility labels, disabled-state handling, grab cursors, and customizable action-bar rendering.
  * Added the default “Drag to reorder” tooltip text.

* **Documentation**
  * Documented drag handles, configuration options, customization, accessibility, and usage in the API reference.

* **Tests**
  * Added coverage for rendering, labeling, refs, disabled behavior, and click handling.

### DIFF
--- a/apps/docs/pages/docs/api-reference/components.mdx
+++ b/apps/docs/pages/docs/api-reference/components.mdx
@@ -20,6 +20,7 @@ Puck provides several components to support different integration approaches.
 
 - [\<ActionBar\>](components/action-bar) - An action bar containing a series of actions, normally used with the [actionBar override](/docs/api-reference/overrides/action-bar).
 - [\<ActionBar.Action\>](components/action-bar-action) - An action for use within the ActionBar component.
+- [\<ActionBar.DragHandle\>](components/action-bar-drag-handle) - A drag handle for use within the ActionBar component.
 - [\<ActionBar.Group\>](components/action-bar-group) - A group of actions for use within the ActionBar component.
 - [\<ActionBar.Label\>](components/action-bar-label) - A label for use within the ActionBar component.
 - [\<ActionBar.Separator\>](components/action-bar-separator) - A separator to divide elements in the ActionBar component.
--- a/apps/docs/pages/docs/api-reference/components/_meta.js
+++ b/apps/docs/pages/docs/api-reference/components/_meta.js
@@ -1,6 +1,7 @@
 const menu = {
   "action-bar": {},
   "action-bar-action": {},
+  "action-bar-drag-handle": {},
   "action-bar-group": {},
   "action-bar-label": {},
   "action-bar-separator": {},
--- a/apps/docs/pages/docs/api-reference/components/action-bar-drag-handle.mdx
+++ b/apps/docs/pages/docs/api-reference/components/action-bar-drag-handle.mdx
@@ -1,0 +1,50 @@
+---
+title: <ActionBar.DragHandle>
+---
+
+import { PuckPreview } from "@/docs/components/Preview";
+import { ActionBar } from "@/puck";
+
+# \<ActionBar.DragHandle\>
+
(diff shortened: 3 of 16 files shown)


FedericoBonel:
Hey @dkrasniy! Thank you very much for the contribution 🙏

Played with this and it looks great!

I think we want to merge it, and want to do a proper code review, but before that we're keen to merge #1735 since it introduces some new dnd behavior that might play nicely with this.

In the meantime, if you could rebase from main it would be great as it would pick up the fix for the build failure you're facing.

FedericoBonel:
Awesome stuff @dkrasniy! Thanks again for the contribution, I have picked it up and continued it in #1810

PR #1771, opened 2026-07-30
Upgrade `@puckeditor/cloud-client` `@puckeditor/plugin-ai` and fix recipe builds

FedericoBonel:
This PR fixes the CI. Builds are currently failing because the lockfile is still pointing to older versions of `plugin-ai` and `cloud-client`. These versions did not include `withDynamicConfig` and other APIs we added.

The root cause was that running `yarn upgrade @puckeditor/cloud-client @puckeditor/plugin-ai` is not enough to upgrade with the wider version range. `^0` matches `0.7.0`, so the lockfile was never updated. 
The solution was to remove the entries for the Puck AI packages from the lockfile and let `yarn install` regenerate them. I guess this means we should verify whether `^0` is the version range we want here.

This was never surfaced in the `npm create puck-app` output because, in those cases, the lockfiles are created from scratch.

Upgrading surfaced another issue with the types in the new versions, so this PR fixes those issues as well.

### DIFF
--- a/recipes/react-router-ai/app/routes/puck-splat.tsx
+++ b/recipes/react-router-ai/app/routes/puck-splat.tsx
@@ -6,6 +6,7 @@ import { createAiPlugin, withDynamicConfig } from "@puckeditor/plugin-ai";
 
 import type { Route } from "./+types/puck-splat";
 import { config } from "../../puck.config";
+import type { UserData } from "../../puck.config";
 import { resolvePuckPath } from "~/lib/resolve-puck-path.server";
 import { getPage, savePage } from "~/lib/pages.server";
 import { PuckRender } from "~/components/puck-render";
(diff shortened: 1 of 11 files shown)


recipes/react-router-ai/app/routes/puck-splat.tsx
```diff
@@ -92,9 +93,7 @@ function Editor() {
         data={loaderData.data}
         onPublish={async (data) => {
           await fetcher.submit(
-            {
-              data,
-            },
+            { data: data as UserData },
```
FedericoBonel, line 96: `withDynamicConfig` returns a defaulted `Config` type, which then makes this onPublish receive a defaulted `Data` type.

`react-router` actions don't like that type, because it can include value types that the action doesn't like. This is actually a problem with `plugin-ai` and `react-router` I think, but open to do a patch for this somehow.

chrisvxd (approved):

PR #1772, opened 2026-07-31
Fede/puck 493 root slot content unmounts and remounts on every default

### LINKED ISSUE #1745: InlineTextField caret drops after one character when block is inside a root slot field
## Description

Any block whose config declares a `contentEditable: true` text field loses
its caret after exactly **one character** typed, when the block is placed
inside a root `slot` field. The character commits to the data model, but
the caret disappears and no further characters can be typed until the
field is re-clicked — at which point the same one-char-then-blur cycle
repeats.

Blocks in the classic top-level `content` zone, and blocks inside `array`
fields at the top level, are **unaffected**. Only nesting inside a **root
`slot` field** triggers the bug.

Tracing this in the source: `packages/core/lib/field-transforms/default-transforms/slot-transform.tsx`
returns a brand-new anonymous `Slot` function on every invocation, so
React (comparing component types by reference identity) unmounts and
remounts the entire subtree under the slot every time the theme's
`root.render` receives updated props. That remounts every
`<InlineTextField>` inside, resetting its `useState(false)` `isFocused`
flag to `false`; `contentEditable` flips back to `"false"` and the
browser drops the caret.

Believed to be the same root cause as still-open
[#1665](https://github.com/puckeditor/puck/issues/1665) (which describes
the more general "root slot subtree unmounts on any update" symptom).
This ticket adds a concrete `InlineTextField` reproduction and traces
the mechanism to a specific line.

Related: [#1369](https://github.com/puckeditor/puck/issues/1369) —
closed as non-reproducible; the repro below appears to match its title
exactly and I believe was misdiagnosed as a userland re-render.

## Environment

- Puck version: `0.22.0` (also confirmed on `0.22.2`)
- Browser version: Chrome 141, Safari 18 (behavior identical on both)
- Additional environment info: React 19.1, Vite 6, TypeScript 5.6, no plugins, no custom `fieldTransforms`

## Steps to reproduce

Minimal reproduction — plain Puck, one root `slot` field, one block with
a `contentEditable` text field:

```tsx
import type { Config, Data } from "@puckeditor/core";

const config: Config = {
  root: {
    fields: { main: { type: "slot" } },
    defaultProps: { main: [] },
    render: ({ main: Main }: any) => (
      <div style={{ padding: 24 }}>
        <Main />
      </div>
    ),
  },
  components: {
    Headline: {
      fields: { text: { type: "text", contentEditable: true } },
      defaultProps: { text: "Edit me" },
      render: ({ text }: { text: string }) => <h1>{text}</h1>,
    },
  },
};

const data: Data = {
  root: {
    props: {
      main: [{ type: "Headline", props: { id: "hl-1", text: "Edit me" } }],
    },
  },
  content: [],
};

export function Repro() {
  return <Puck config={config} data={data} iframe={{ enabled: false }} />;
}
```

1. Render `<Repro />`.
2. Click the "Edit me" text on the canvas so the inline editor activates.
3. Type any single character (e.g. `a`).
4. The character commits (data updates, sidebar reflects it) but the
   caret is gone; further keystrokes do nothing.
5. Move the `Headline` from `root.props.main` to a top-level `content: [...]`
   array in the same document — the bug disappears.

## What happens

Caret is dropped after exactly one character. Every subsequent focus
attempt yields the same behavior (one char, then blur).

## What I expect to happen

The caret stays in place and the user can type continuously, matching
the behavior of inline editing on any non-slot-nested block.

## Additional Media

_None attached — the repro above is self-contained. Happy to record a
screencap if useful._

FedericoBonel:
Closes puckeditor/puck#1745 and puckeditor/puck#1665

## Description

This PR fixes an issue where modifying any props would remount components that lived within root slots. This would also cause caret positions to be lost when `contentEditable` fields lived within root slots.

The reason was that the editor path used to render the root called the plain `useSlots` hook with the entire `root` data object. This function does not track props, and it recreated the transform functions on every change because `root` is a plain object that is recreated on every render. As a result, the slot render functions changed references between re-renders and were recreated, causing the components to remount.

## Changes made

* Refactored build mappers for performance and readability by replacing the `reduce` call with a plain `forEach`.
* Updated the preview to follow the same approach used to render component slots:
  1. The preview subscribes to the flattened root data, which defaults slots to `null`.
  2. The flattened root data is expanded and passed to `useTransformSlotsTracked`.
  3. `useTransformSlots` creates stable slot render functions because the slot props are equal between calls (`null`).
  4. Each `DropZonePure` renders an individual `DropZoneChild` for each `contentId` in `indexes`.
  5. Each `DropZoneChild` subscribes to its flattened props.

## How to test

* Render Puck with a root slot and some other components that log when they unmount:

```tsx
  components: {
    HeadingBlock: {
      fields: {
        title: { type: "text" },
      },
      render: ({ title }) => {
        useEffect(() => {
          return () => console.log(`Unmounting ${title}`);
        }, []);

        return (
          <div style={{ padding: 64 }}>
            <h1>{title}</h1>
          </div>
        );
      },
    },
  },
  root: {
    fields: {
      body: { type: "slot" },
      title: { type: "text" },
    },
    render: ({ body: Body, children }) => (
      <div>
        <Body />
        {children}
      </div>
    ),
  },
};

export default config;
```

* Add some components to the root slot and the default slot, then modify the components in the root slot.
* Observe that the root slot components do not remount.
* Modify the components in the default slot.
* Observe that the root slot components do not remount.
* Modify the root title.
* Observe that the root slot components do not remount.

### DIFF
--- a/packages/core/components/Puck/components/Preview/index.tsx
+++ b/packages/core/components/Puck/components/Preview/index.tsx
@@ -1,25 +1,13 @@
-import { useShallow } from "zustand/react/shallow";
-import { RefObject, useEffect, useRef, useMemo, memo } from "react";
+import { RefObject, useEffect, useRef, useMemo } from "react";
 
-import { rootDroppableId } from "../../../../lib/root-droppable-id";
 import { useAppStore } from "../../../../store";
 import { getClassNameFactory } from "../../../../lib";
-import {
-  ComponentData,
-  RootData,
-  DefaultRootRenderProps,
-} from "../../../../types";
 import { BubbledPointerEvent } from "../../../../lib/bubble-pointer-event";
-import { useFieldTransformsTracked } from "../../../../lib/field-transforms/use-field-transforms-tracked";
-import { getSlotTransform } from "../../../../lib/field-transforms/default-transforms/slot-transform";
-import { expandNode } from "../../../../lib/data/flatten-node";
-import { toComponent } from "../../../../lib/data/to-component";
 
-import { DropZoneEditPure, DropZonePure } from "../../../DropZone";
 import AutoFrame, { autoFrameContext } from "../../../AutoFrame";
 import { Render } from "../../../Render";
-import { useRichtextProps } from "../../../RichTextEditor/lib/use-richtext-props";
 
+import EditorPage from "./components/editor-page";
 import styles from "./styles.module.css";
 
 const getClassName = getClassNameFactory("PuckPreview", styles);
@@ -93,42 +81,8 @@ const usePreviewModeAttribute = (ref: RefObject<HTMLIFrameElement | null>) => {
   }, [previewMode, status, iframeEnabled]);
 };
 
-const slotFieldTransforms = getSlotTransform(DropZoneEditPure);
-
-type PageProps = DefaultRootRenderProps & { config: any };
-
-const Page = memo(({ config, ...pageProps }: PageProps) => {
-  const rootAsComponent = toComponent({ props: pageProps });
-
-  // Get the props with stable slot render functions
-  // (tracked only sees `null` for slot props across calls, so prev always === next)
-  const propsWithSlots = useFieldTransformsTracked(
-    config,
-    rootAsComponent,
-    slotFieldTransforms
-  );
-
-  const richtextProps = useRichtextProps(config.root?.fields ?? {}, pageProps);
-
-  return config.root?.render ? (
-    config.root?.render({
-      ...propsWithSlots,
-      ...richtextProps,
-      id: "puck-root",
-    })
-  ) : (
-    <>{propsWithSlots.children}</>
-  );
-});
-
-Page.displayName = "Page";
-
 export const Preview = ({ id = "puck-preview" }: { id?: string }) => {
   const dispatch = useAppStore((s) => s.dispatch);
-  // All the props, slot props are defaulted to null
-  const flatRootProps = useAppStore(
-    useShallow((s) => s.state.indexes.nodes.root?.flatData.props)
-  );
   const config = useAppStore((s) => s.config);
   const setStatus = useAppStore((s) => s.setStatus);
   const iframe = useAppStore((s) => s.iframe);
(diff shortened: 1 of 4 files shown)


packages/core/lib/field-transforms/build-mappers.ts
```diff
@@ -26,33 +26,32 @@ export function buildMappers<
   readOnly?: T["readOnly"],
   forceReadOnly?: boolean
 ) {
-  return Object.keys(transforms).reduce<Mappers>((acc, _fieldType) => {
+  const newMappers: Mappers = {};
```
chrisvxd, line 29: The build mappers commit should not be performance, as that'll be reported in the changelog and "remove repeated work from build mappers" means nothing to the end-user

packages/core/components/Puck/components/Preview/index.tsx
```diff
@@ -120,8 +140,14 @@ export const Preview = ({ id = "puck-preview" }: { id?: string }) => {
 
   const Frame = useMemo(() => overrides.iframe, [overrides]);
 
-  // DEPRECATED
-  const rootProps = root.props || root;
+  // Root as object with slots still defaulted to null
```
chrisvxd: You're dropping the deprecated behaviour (props directly on root rather than just on root.props) which is a breaking change.

We also retain the `// DEPRECATED` flag so we can tidy things up when we do a breaking release (v1)

chrisvxd (changes requested):
Couple of comments

packages/core/components/Puck/components/Preview/index.tsx
FedericoBonel: `flatData` already handles that because `walkAppState` handles the mapping. Will add a `//deprecated` comment on it though

packages/core/lib/field-transforms/build-mappers.ts
FedericoBonel: Got i! Will flag it as refactor

packages/core/components/Puck/components/Preview/index.tsx
FedericoBonel: Added the `//DEPRECATED` comment

packages/core/lib/field-transforms/build-mappers.ts
FedericoBonel: Done!

chrisvxd (approved):

PR #1775, opened 2026-08-01
fix: remove external Inter font import from no-external.css and runtime CSS

### LINKED ISSUE #1769: [Bug] no-external.css imports Inter from rsms.me (render-blocking - pagespeed insights)
## Description

`@puckeditor/core` ships CSS bundles that hard-code an external font import:

```css
@import "https://rsms.me/inter/inter.css";
```

This line is the **first line of both `dist/index.css` and `dist/no-external.css`**. The
`no-external.css` bundle is, by its name and documented purpose, supposed to be free of
external network dependencies — yet it pulls the Inter font from the third-party
`rsms.me` CDN.

Consuming the CSS in a bundler-based app (Next.js / webpack / Turbopack / Vite) means the
browser issues a cross-origin request to `rsms.me` for the Inter stylesheet. Because it is a
top-of-file `@import`, the request is **render-blocking**: a Lighthouse / PageSpeed Insights
audit flags `https://rsms.me/inter/inter.css` under the **"Render-blocking requests"**
section, hurting FCP/LCP. It also breaks strict CSP setups, offline/air-gapped deployments,
and privacy requirements, and defeats the entire point of the `no-external` variant.

## Environment

- Puck version: 0.21.1 (`@puckeditor/core`)
- Browser version: any (Chrome 138 desktop, reproduced)
- Additional environment info: Next.js 16 App Router storefront, Turbopack/webpack bundling, pnpm

## Steps to reproduce

1. Install `@puckeditor/core` and use it in an app that bundles CSS — e.g. render Puck
   content or use the package APIs from your blocks/components:

```tsx
import { 
   type Data, 
   resolveAllData 
} from "@puckeditor/core";
```

   The package's CSS is pulled in as part of using Puck, and its first line is:

```css
@import "https://rsms.me/inter/inter.css";
```

2. Run a production build and serve the app.

3. Run a Lighthouse / PageSpeed Insights performance audit on the page (or inspect the
   browser network tab).

## What happens

- Lighthouse flags `https://rsms.me/inter/inter.css` under the **"Render-blocking requests"**
  section of the performance report, adding to blocking time and hurting FCP/LCP.
- The browser makes an uncontrolled cross-origin request to `https://rsms.me/inter/inter.css`.
- Under a strict `Content-Security-Policy` the request is blocked and the console logs a CSP
  violation.
- The `no-external.css` bundle, which is meant to contain no external dependencies, still
  reaches out to `rsms.me`.

We currently work around this with a `pnpm` patch that strips the `@import` line from both
`dist/index.css` and `dist/no-external.css`.

## What I expect to happen

- `dist/no-external.css` should contain **zero** external `@import`/URL references — that is
  its contract.
- `dist/index.css` should not silently pull a font from a third-party CDN. The Inter font
  should either be bundled/self-hosted by the consumer, documented as an opt-in, or left to
  the app to provide — not imported from `rsms.me` by default.

## Additional Media

Image

jaimemartinez:
## Summary

Fixes #1769 — `no-external.css` imports Inter from rsms.me (render-blocking).

The `no-external.css` bundle is intended to be free of external network dependencies, but two issues caused the external Inter font `@import` to leak through:

1. **Runtime CSS**: The dynamically injected CSS (via `useInjectUiCss`) was generated from `bundle/index.ts`, which included `@import url("https://rsms.me/inter/inter.css")`. This meant even users of the `no-external` variant got a render-blocking cross-origin request.

2. **CSS contract violation**: The external `@import` being baked into the source `index.css` meant it could also appear in runtime CSS strings injected as `<style>` tags, where `@import` statements are non-functional anyway.

## Changes

| File | Change |
|------|--------|
| `bundle/fonts.css` | **[NEW]** Dedicated file for the external Inter font `@import` |
| `bundle/index.css` | Removed external `@import` — now identical to `no-external.css` |
| `bundle/index.ts` | Imports `fonts.css` before `index.css` (preserves backward compat for `dist/index.css`) |
| `tsup.config.ts` | Runtime CSS now generated from `bundle/no-external.ts` instead of `bundle/index.ts` |
| `bundle/__tests__/no-external-css.spec.ts` | **[NEW]** 10 unit tests enforcing the CSS external dependency contract |

## How it works

- `dist/index.css` (aka `puck.css`) still includes the Inter font import — **no breaking change** for existing users
- `dist/no-external.css` is guaranteed free of external URLs — its contract is now enforced by tests
- Runtime-injected CSS never includes external `@import` statements (they're non-functional inside `<style>` tags anyway)

## Testing

- ✅ 10 new unit tests — all passing
- ✅ Full test suite (42 suites, 225 tests) — all passing

## Summary by CodeRabbit

- **New Features**
  - Default styles now load without external font or stylesheet URLs, improving reliability in restricted or offline environments.
  - The complete style bundle continues to support the Inter font and associated styles when explicitly included.
- **Bug Fixes**
  - Corrected runtime style generation to use the no-external bundle by default.
  - Improved separation between standard, no-external, and iframe styling options.
- **Tests**
  - Added coverage to verify stylesheet imports and prevent unintended external resources.

### DIFF
--- a/packages/core/tsup.config.ts
+++ b/packages/core/tsup.config.ts
@@ -85,7 +85,7 @@ const runtimeCssPlugin: Plugin = {
       async () => {
         if (!cachedContents) {
           const [defaultStyles, interactionStyles] = await Promise.all([
-            readBundledCss("bundle/index.ts"),
+            readBundledCss("bundle/no-external.ts"),
             readBundledCss("bundle/iframe-styles.ts"),
           ]);
 
--- a/packages/core/bundle/__tests__/no-external-css.spec.ts
+++ b/packages/core/bundle/__tests__/no-external-css.spec.ts
@@ -1,15 +1,82 @@
+/**
+ * @jest-environment node
+ */
 import * as fs from "fs";
 import * as path from "path";
+import { build, type Plugin } from "esbuild";
 
 const bundleDir = path.resolve(__dirname, "..");
+const packageRoot = path.resolve(bundleDir, "..");
 
 const readBundleFile = (filename: string): string =>
   fs.readFileSync(path.join(bundleDir, filename), "utf-8");
 
 const EXTERNAL_URL_PATTERN = /https?:\/\//;
 
+// --- Helpers to build CSS the same way tsup.config.ts does ---
+
+const runtimeCssStubPlugin: Plugin = {
+  name: "runtime-css-stub",
+  setup(buildApi) {
+    buildApi.onResolve({ filter: /generated\/runtime-css$/ }, () => ({
+      path: "runtime-css-stub",
+      namespace: "runtime-css-stub",
+    }));
+
+    buildApi.onLoad({ filter: /.*/, namespace: "runtime-css-stub" }, () => ({
+      contents: `
+          export const defaultUiStyles = "";
+          export const iframeInteractionStyles = "";
+        `,
+      loader: "js" as const,
+    }));
+  },
+};
+
+/**
+ * Build an entry point with esbuild and return the emitted CSS string.
+ * Mirrors the `readBundledCss` function from tsup.config.ts.
+ */
+const buildCss = async (entryPoint: string): Promise<string> => {
+  const isCssEntry = entryPoint.endsWith(".css");
+  const result = await build({
+    absWorkingDir: packageRoot,
+    bundle: true,
+    format: "iife",
+    logLevel: "silent",
+    packages: "external",
+    platform: "browser",
+    plugins: [runtimeCssStubPlugin],
+    target: ["es2020"],
+    outdir: "out",
+    write: false,
+    ...(isCssEntry
+      ? {
+          stdin: {
+            contents: `import "./${entryPoint}";`,
+            loader: "ts" as const,
+            resolveDir: packageRoot,
+            sourcefile: "runtime-css-entry.ts",
+          },
+        }
+      : {
+          entryPoints: [entryPoint],
+        }),
+  });
+
+  return result.outputFiles
+    .filter((file) => file.path.endsWith(".css"))
+    .map((file) => file.text.trim())
+    .filter(Boolean)
+    .join("\n\n");
+};
+
+// ===========================================================================
+// Source file contract tests
+// ===========================================================================
+
 describe("CSS bundle external dependency contract", () => {
-  describe("no-external.css", () => {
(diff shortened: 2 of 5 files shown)


packages/core/tsup.config.ts
```diff
@@ -85,7 +85,7 @@ const runtimeCssPlugin: Plugin = {
       async () => {
         if (!cachedContents) {
           const [defaultStyles, interactionStyles] = await Promise.all([
-            readBundledCss("bundle/index.ts"),
+            readBundledCss("bundle/no-external.ts"),
```
chrisvxd, line 88: We want to retain the external font file in runtime CSS for most users.

When importing `no-external` explicitly, the runtime CSS shouldn't load at all.

chrisvxd (changes requested):

jaimemartinez:
Thanks @chrisvxd — understood, I'll put the runtime CSS back on `bundle/index.ts`.

Before I push, a question about what's actually left of this PR.

I built `main` and this branch with the revert applied, and compared the output:

| artifact | `main` | this branch + revert |
|---|---|---|
| `dist/index.css` | contains `rsms.me` | same |
| `dist/no-external.css` | clean | same |
| runtime CSS inlined in `dist/index.js` | contains `rsms.me` | same |

The shipped artifacts come out identical to `main`. Splitting `fonts.css` out is a no-op, since `bundle/index.ts` imports both files.

Digging into it, `bundle/no-external.css` never contained the external `@import` in source — not even at `v0.21.1` — and `dist/no-external.css` builds clean on `main` today. So the first ask in #1769 (*"`dist/no-external.css` should contain zero external references"*) already holds. The second (*"`dist/index.css` should not silently pull a font from a third-party CDN"*) is the one your comment rejects, and the reporter's repro — importing anything from the package root pulls `dist/index.css` — falls under that.

So how would you like to handle it?

1. **Reduce this PR to the regression guard**: keep the tests plus the `fonts.css` split so the no-external contract is enforced in CI, and close #1769 explaining that `no-external.css` already satisfies it and the font in `puck.css` is intentional.
2. **Give consumers an opt-out**: now that `fonts.css` is its own file, it could be exported as `@puckeditor/core/fonts.css` and dropped from `index.ts`, letting people self-host Inter. That's breaking for `puck.css` though, so only if that's a direction you want.
3. **Just close it** if you'd rather not carry the extra files.

Happy to go with whichever — option 1 is what I'd default to.

PR #1776, opened 2026-08-01
fix: render explicit viewport height at actual size (#883)

### LINKED ISSUE #883: Inconsistent behavior when setting custom viewport height
## Description

When setting a custom viewport width and height, the preview behaves inconsistently: the width stays within the preview's boundaries, while the height grows beyond them, making the preview scrollable. This inconsistency can cause issues for users trying to use viewports for aspect ratio-based designs.

## Steps to Reproduce

1. Set a custom viewport width and height matching your monitor resolution (e.g., 1920 × 1080).
2. Open the editor and observe the preview.
3. Use the preview controls to zoom out.

## What Happens

* The viewport width remains constrained within the preview boundaries.
* The viewport height exceeds the preview boundaries, causing scrolling.
* When zooming out, the height and width remain unchanged, leading to an inaccurate representation of the viewport size.

## Expected Behavior

The preview should maintain consistency in how it scales. There are two possible solutions:<br>A. The width should also expand to match the actual size, making both dimensions scrollable.<br>B. The aspect ratio should be preserved, scaling the height accordingly to accurately reflect the target resolution.

Option A is likely clearer since explicit width and height values are set. However, adding an option to enforce aspect ratios via the [CSS aspect-ratio property](<https://developer.mozilla.org/en-US/docs/Web/CSS/aspect-ratio>) could also be beneficial for users needing more flexibility.

**Video demonstrating the issue (height exceeding width boundaries):**:

https://github.com/user-attachments/assets/73324210-8819-42d4-a4a5-20e4d00407f7

jaimemartinez:
## Problem

Fixes #883.

When a custom viewport has both an explicit **width and height**, the preview scales inconsistently: the width is auto-fitted to the frame, but the height is rendered at its physical pixel value and overflows the preview. The same happens with manual zoom: zooming out shrinks the width but not the height, so the preview never reflects the target resolution.

## Fix (Option B)

`rootHeight` is the *unscaled* height of the canvas root, which is then multiplied by `transform: scale(zoom)`. Dividing it by `zoom` is correct for `height: "auto"` (the root must fill the frame at any zoom), but for an explicit height it pinned the rendered height while the width was scaled down.

An explicit height now keeps its real value, so it scales by the same factor as the width and the aspect ratio is preserved — e.g. a 1920×1080 viewport renders as 16:9 at any zoom level:

```ts
if (widthZoom < heightZoom) {
  rootHeight =
    uiViewport.height === "auto" ? viewportHeight / zoom : viewportHeight;
}
```

The existing fit-to-width auto-zoom is untouched, so there's no zoom-in behaviour, no horizontal overflow, and drag-and-drop is unaffected. Auto-height viewports are unchanged.

> Earlier revisions of this PR implemented Option A (render at actual size, both axes scrollable). That was switched to Option B per review feedback.

## Verification

Measured in the demo (frame 1484×753):

| Viewport | zoom | Rendered | Aspect | Overflow |
|---|---|---|---|---|
| 1920×1080 | 0.697 | 1339×753 | 1.778 | none |
| 4000×2000 | 0.371 | 1484×742 | 2.000 | none, both edges visible |
| 800×2000 | 0.377 | 301×753 | 0.400 | none |
| 1920×1080, manual zoom 50% / 25% | 0.5 / 0.25 | 960×540 / 480×270 | 1.778 | none |
| `auto` height (Small / Large / Full) | unchanged | unchanged | — | none |
| `auto` height, manual zoom 75% | 0.75 | 960×753 | — | "Constrain height" effect intact |

## Tests

Adds `get-zoom-config.spec.ts` (this helper was previously untested):

- explicit height, width is the constraint → fits width, `rootHeight` stays real (the fixed branch)
- explicit height, height is the constraint → fits height
- explicit height, both exceed the frame → most constrained axis wins
- explicit height within the frame → 1:1
- auto-fit ignores the previous zoom
- non-numeric width (`"100%"`) → no auto-zoom
- auto-height regression guards (fit width + stretch to frame, fits, `"100%"`)

### DIFF
--- a/packages/core/lib/__tests__/get-zoom-config.spec.ts
+++ b/packages/core/lib/__tests__/get-zoom-config.spec.ts
@@ -32,7 +32,11 @@ describe("getZoomConfig", () => {
 
   // Regression guard: auto-height viewports keep the existing auto-zoom-to-fit-width behaviour.
   it("auto-fits the width for an auto-height viewport wider than the frame", () => {
-    const { zoom, autoZoom, rootHeight } = getZoomConfig(vp(1600, "auto"), frame, 1);
+    const { zoom, autoZoom, rootHeight } = getZoomConfig(
+      vp(1600, "auto"),
+      frame,
+      1
+    );
     expect(zoom).toBeCloseTo(0.5); // 800 / 1600
     expect(autoZoom).toBeCloseTo(0.5);
     expect(rootHeight).toBeCloseTo(1200); // viewportHeight(600) / zoom(0.5)
--- a/packages/core/lib/__tests__/get-zoom-config.spec.ts
+++ b/packages/core/lib/__tests__/get-zoom-config.spec.ts
@@ -1,0 +1,48 @@
+import { getZoomConfig } from "../get-zoom-config";
+import type { AppState } from "../../types";
+
+// getZoomConfig reads the frame's rendered box; stub it to a fixed 800×600 preview area.
+jest.mock("css-box-model", () => ({
+  getBox: () => ({ contentBox: { width: 800, height: 600 } }),
+}));
+
+const frame = {} as HTMLElement;
+type Viewport = AppState["ui"]["viewports"]["current"];
+const vp = (width: Viewport["width"], height: Viewport["height"]): Viewport =>
+  ({ width, height } as Viewport);
+
+describe("getZoomConfig", () => {
+  // #883: an explicit height is a concrete target, so render at actual size (both axes scroll),
+  // instead of fitting the width while the height overflows.
+  it("renders an explicit height at 1:1 with both axes scrollable (#883, Option A)", () => {
+    expect(getZoomConfig(vp(1920, 1080), frame, 1)).toEqual({
+      autoZoom: 1,
+      zoom: 1,
+      rootHeight: 1080,
+    });
+  });
+
+  it("keeps an explicit height even when it is smaller than the frame", () => {
+    expect(getZoomConfig(vp(400, 300), frame, 1)).toEqual({
+      autoZoom: 1,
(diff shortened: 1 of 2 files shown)


FedericoBonel (changes requested):
Hey @jaimemartinez! Thank you very much for the contribution.

I haven't done a code review yet, but I have tried it functionally, and this breaks when you have previews wider than the viewport. For example, if I set a viewport width of 4000px (I'm using a 1920px-wide monitor), the preview doesn't let me scroll all the way to the left.

It also doesn't work well with drag and drop, since it doesn't scroll horizontally while dragging.

Looking back at this issue, I think the best solution would be to go with Option B, since that was what was already happening for width only. I think Option A introduces a "zoom in" behavior, which is a whole new feature rather than a bug fix and we got another issue tracking that.

FedericoBonel:
Closing since there haven't been any updates in the last two weeks.

Feel free to open the PR again if you can address the remaining comments and want us to review it again.

jaimemartinez:
@FedericoBonel thanks for testing, and sorry for the delay. You're right — I've switched to **Option B**.

The change is now a single line in `getZoomConfig`: for an explicit height, `rootHeight` keeps its real value instead of being divided by `zoom`, so the height scales by the same factor as the width. The existing fit-to-width behaviour is untouched, so there's no zoom-in, no horizontal overflow (verified 4000px on a 1484px frame — both edges visible, `scrollWidth === clientWidth`), and manual zoom now keeps the aspect ratio too. Auto-height viewports are unchanged and covered by regression tests.

Updated the PR description with the measurements. Happy to squash the earlier Option A commits if you'd prefer a clean history.

PR #1777, opened 2026-08-03
fix: don't reset caret position in custom and overriden fields

### LINKED ISSUE #1642: Cursor jumps to end in controlled custom input fields

FedericoBonel:
Closes puckeditor/puck#1642

Requires the changes from puckeditor/puck#1757.

## Description

This PR fixes an issue where controlled custom and overridden fields lost their caret position on every edit.

This happened because the value of a custom or overridden field was persisted to the field store only through a subscription to `appStore`. The problem with this approach was that `appStore` was updated only after the data had been resolved, making it an asynchronous operation.

As a result, the value received while typing in the field was always behind the value being displayed. This caused strange visual effects with long-running resolvers and made the field lose its caret position.

## Changes made

* Custom and overridden fields now reuse the same logic as text and text area fields. They keep a local copy of the value and pass it to the override or custom field render function. This ensures that the fields always receive the most recently edited value for display while synchronizing with the store value on blur.
  * They now reuse the `useLocalValue` utility.
  * `useLocalValue` was updated to accept values other than strings.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved synchronization for custom and overridden field values, including nested objects and arrays.
  * Prevented stale field renders and ensured change handlers run in the correct order.
  * Improved handling of empty and nullish values in text and default fields.
  * Disabled fields no longer track or update local state unnecessarily.

### DIFF
--- a/packages/core/components/AutoField/lib/use-local-value.ts
+++ b/packages/core/components/AutoField/lib/use-local-value.ts
@@ -2,26 +2,69 @@ import { useCallback, useEffect, useState } from "react";
 import { useDeepField } from "./use-deep-field";
 import { useIsFocused } from "./use-is-focused";
 
-export const useLocalValue = (path: string, onChange: (val: any) => void) => {
-  const value = useDeepField(path);
+type UseLocalValueOptions = {
+  /**
+   * Set false to skip tracking store updates. When false, returns a stable
+   * `undefined` value and the untouched `onChange`, so callers won't re-render
+   * when the field store value changes.
+   */
+  tracked?: boolean;
+  /**
+   * Returned in place of a nullish value.
+   */
+  fallback?: any;
+};
+
+/**
+ * Track a field's value in local state, seeded and kept in sync from the field
+ * store.
+ *
+ * The store is a shared channel: the subscription in Fields writes every
+ * committed change into it. Because `resolveData` is async, a commit can land
+ * carrying the value from a keystroke the user has already typed past, which
+ * would overwrite what they're typing and reset the caret. Local state is
+ * private, so holding the value here and ignoring the store while the field is
+ * focused keeps the edit intact. On blur it re-syncs from the store.
+ */
+export const useLocalValue = (
+  path: string,
+  onChange: (value: any, ...rest: any[]) => void,
+  { tracked = true, fallback }: UseLocalValueOptions = {}
+) => {
+  const value = useDeepField(path, tracked);
   const isFocused = useIsFocused(path);
 
-  const [localValue, setLocalValue] = useState(value?.toString());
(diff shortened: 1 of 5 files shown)


packages/core/components/AutoField/lib/use-local-value.ts
```diff
@@ -2,16 +2,45 @@ import { useCallback, useEffect, useState } from "react";
...
 
-export const useLocalValue = (path: string, onChange: (val: any) => void) => {
-  const value = useDeepField(path);
+type UseLocalValueOptions = {
+  /**
+   * Set false to skip tracking entirely. Returns the field store value as a
+   * stable `undefined` and the untouched `onChange`, so a field that doesn't
+   * read its value from the store pays nothing.
+   */
+  enabled?: boolean;
```
chrisvxd: I find `enabled` a bit of a weird API name for this. Can you be more explicit? like `tracked`?

chrisvxd (changes requested):

packages/core/components/AutoField/lib/use-local-value.ts
FedericoBonel: Done! The refactor I have gets rid of this attribute completely since I consider it a smell, but that might land later on, so we can do this for now.

FedericoBonel:
I did address your comment, but forgot to re-request your review! My bad

chrisvxd (approved):

PR #1783, opened 2026-08-06
Improve drag and drop for granular changes follow ons

### LINKED ISSUE #1286: Limit max height of element while dragging

### LINKED ISSUE #1358: Improve drag-and-drop for smaller slots

FedericoBonel:
This PR builds on top of puckeditor/puck#1735 and addresses a few bugs, the PR comments, and re-organizes the architecture while keeping the same behavior. It also rebases from main and resolves the conflicts we had.

\---- Original PR description:

Closes puckeditor/puck#1358, closes puckeditor/puck#1286

## Description

This PR changes the default drag-and-drop behavior when dragging between slots to use a static placeholder with a line preview. When dragging within the same slot, the existing behavior remains.

This is configurable via the `dnd.behavior` API, which can be `auto` (the default, changing based on reparenting), `fluid` (the original) or `static`.

## How to test

* Try dragging items between slots
* Set `?dndBehavior="fluid"` on the demo URL to restore the legacy behavior
* Set `?dndBehavior="static"` on the demo URL to enable static behavior for everything

In code, set the `dnd.behavior` property to change behavior:

```tsx
<Puck dnd={{ behaviour: "static" }} />
```

## Remaining considerations

1. I wasn't totally sure of the parameter name `behavior`, so if you have any suggestions, I'm all ears.

## Summary by CodeRabbit

* **New Features**
  * Added configurable drag-and-drop animation modes: automatic, fluid, and static.
  * Improved drag previews with insertion indicators, ghost previews, and smoother movement across zones.
  * Added polished drop and layout animations, including reduced-motion support.
  * Enhanced dragging across varied layouts, directions, virtualized content, and embedded frames.
* **Documentation**
  * Documented the available drag-and-drop behavior options and animation modes.
* **Tests**
  * Added coverage for drag placement, animations, layout handling, and configuration behavior.

### DIFF
--- a/apps/docs/pages/docs/api-reference/components/puck.mdx
+++ b/apps/docs/pages/docs/api-reference/components/puck.mdx
@@ -149,7 +149,7 @@ Configure drag-and-drop behavior.
 
 Control animation behavior while dragging. Defaults to `"auto"`.
 
-- `"auto"`: Use fluid behavior within a slot, then switch to a static when dragging an item between slots.
+- `"auto"` (default): Use fluid behavior within a slot, switching to a static line placeholder when dragging an item between slots or inserting a new component.
 - `"fluid"`: Move the item during a drag, animating other items out of the way.
 - `"static"`: Freeze the item in-place during a drag, showing a line placeholder at the drop destination.
 
--- a/apps/docs/pages/docs/api-reference/components/puck.mdx
+++ b/apps/docs/pages/docs/api-reference/components/puck.mdx
@@ -149,7 +149,7 @@ Configure drag-and-drop behavior.
 
 Control animation behavior while dragging. Defaults to `"auto"`.
 
-- `"auto"` (default): Use fluid behavior within a slot, switching to a static line placeholder when dragging an item between slots or inserting a new component.
+- `"auto"` (default): Use fluid behavior within a slot, then switch to static behavior when dragging an item between slots or inserting a new component.
 - `"fluid"`: Move the item during a drag, animating other items out of the way.
 - `"static"`: Freeze the item in-place during a drag, showing a line placeholder at the drop destination.
 
--- a/apps/docs/pages/docs/api-reference/components/puck.mdx
+++ b/apps/docs/pages/docs/api-reference/components/puck.mdx
@@ -139,10 +139,19 @@ Configure drag-and-drop behavior.
 
 #### dnd params
 
-| Param                                       | Example                    | Type    | Status |
-| ------------------------------------------- | -------------------------- | ------- | ------ |
-| [`disableAutoScroll`](#disableautoscroll)   | `disableAutoScroll: true`  | boolean | -      |
-| [`disableOutlineDrag`](#disableoutlinedrag) | `disableOutlineDrag: true` | boolean | -      |
+| Param                                       | Example                    | Type                            | Status |
+| ------------------------------------------- | -------------------------- | ------------------------------- | ------ |
+| [`behavior`](#behavior)                     | `behavior: "static"`       | `"auto" \| "fluid" \| "static"` | -      |
+| [`disableAutoScroll`](#disableautoscroll)   | `disableAutoScroll: true`  | boolean                         | -      |
+| [`disableOutlineDrag`](#disableoutlinedrag) | `disableOutlineDrag: true` | boolean                         | -      |
+
+##### `behavior`
+
+Control animation behavior while dragging. Defaults to `"auto"`.
+
+- `"auto"`: Use fluid behavior within a slot, then switch to a static when dragging an item between slots.
+- `"fluid"`: Move the item during a drag, animating other items out of the way.
+- `"static"`: Freeze the item in-place during a drag, showing a line placeholder at the drop destination.
 
 ##### `disableAutoScroll`
 
--- a/packages/core/components/DragDropContext/index.tsx
+++ b/packages/core/components/DragDropContext/index.tsx
@@ -25,6 +25,11 @@ import {
   ZoneStoreProvider,
 } from "../DropZone/context";
 import { createNestedDroppablePlugin } from "../../lib/dnd/NestedDroppablePlugin";
+import { prepareCommitFlip } from "../../lib/dnd/flip-commit";
+import {
+  resolveDndMode,
+  resolveOriginPreviewIndex,
+} from "../../lib/dnd/resolve-dnd-mode";
 import { insertComponent } from "../../lib/insert-component";
 import { moveComponent } from "../../lib/move-component";
 import { useDebouncedCallback } from "use-debounce";
@@ -41,6 +46,8 @@ import {
 import { useSensors } from "../../lib/dnd/use-sensors";
 import { getFrame } from "../../lib/get-frame";
 import { effect } from "@dnd-kit/state";
+import type { DndBehavior } from "../../types";
+import { useLinePlaceholder } from "./use-line-placeholder";
 
 const DEBUG = false;
 
@@ -83,6 +90,7 @@ const AREA_CHANGE_DEBOUNCE_MS = 100;
 type DragDropContextProps = {
   children: ReactNode;
   disableAutoScroll?: boolean;
+  behavior?: DndBehavior;
 };
 
 /**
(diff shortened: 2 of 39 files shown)


packages/core/components/DropZone/LinePlaceholder.tsx
```diff
@@ -1,40 +1,43 @@
...
+import { getComponentSelector } from "../../lib/dom-selectors";
+
 import styles from "./styles.module.css";
 
 const getClassName = getClassNameFactory("DropZone", styles);
-const LINE_PLACEHOLDER_SIZE = 4;
 
-const clamp = (value: number, min: number, max: number) =>
-  Math.max(min, Math.min(max, value));
+const LINE_SIZE = "var(--puck-border-width-focus, 2px)";
```
chrisvxd: Hmm `--puck-border-width-focus` is probably not the right theming variable - the intention is different.

I also think 2px might also be too small for the default here? Did you want it to be a little tighter?

packages/core/components/DragDropContext/index.tsx
```diff
@@ -608,15 +608,24 @@ const DragDropContextClient = ({
...
+                  targetZone,
+                  manager
+                );
+
+                if (linePreviewIndex !== null) {
+                  targetIndex = linePreviewIndex;
+                } else {
+                  // No gap resolved: keep the fluid `targetIndex` and drop the
+                  // line flag so onDragEnd does not offset it a second time.
+                  isLinePlaceholder = false;
```
chrisvxd: Will this revert to fluid mode? That will be confusing if the user has explicitly set `static`

apps/docs/pages/docs/api-reference/components/puck.mdx
```diff
@@ -149,7 +149,7 @@ Configure drag-and-drop behavior.
 
 Control animation behavior while dragging. Defaults to `"auto"`.
 
-- `"auto"`: Use fluid behavior within a slot, then switch to a static when dragging an item between slots.
+- `"auto"` (default): Use fluid behavior within a slot, switching to a static line placeholder when dragging an item between slots or inserting a new component.
```
chrisvxd: Hmm I now find this a bit receptive, as it's redefining `static` behavior (but doesn't redefine `fluid` behavior).

I think there was a typo in the original one ("switch to a static when dragging"). Would it be better if we just fixed that?

```suggestion
- `"auto"` (default): Use fluid behavior within a slot, then switch to static behavior when dragging an item between slots or inserting a new component.
```

packages/core/components/DropZone/styles.module.css
```diff
@@ -43,9 +43,10 @@
 .DropZone-linePlaceholder {
   background: var(
     --puck-slot-component-color-placeholder,
-    var(--puck-color-azure-06)
+    var(--puck-color-insertion-indicator)
```
chrisvxd: I know it wasn't added here, but I find "puck-color-insertion-indicator" quite an awkward theming variable. I think "puck-color-static-drag-indicator" might be clearer

chrisvxd (changes requested):

packages/core/components/DropZone/LinePlaceholder.tsx
FedericoBonel: Yeah. was just thinking on unifying it with the outline line we got, which uses this. We could introduce a new variable for both.

packages/core/components/DragDropContext/index.tsx
FedericoBonel: It's a defensive condition. In reality this won't ever be reached, the only case where this happens is if the `window` is undefined, which would be very strange. In that case, you could insert in the wrong position if you keep the line, so this disables it then. Again, mad condition to reach, but happy to rever this

packages/core/components/DropZone/styles.module.css
FedericoBonel: We also use this for the outline. Which might be fine, if we set that relationship between line === static.

chrisvxd: We could also use `line-placeholder`?

packages/core/components/DragDropContext/index.tsx
chrisvxd: Fine if it's defensive, maybe just comment that it should never be reached

packages/core/components/DropZone/styles.module.css
FedericoBonel: Yeah we can! I don't remember now why I decided on the one above, I think I might have done it because it should be "semantic", and decided to go with that. But line placeholder works too, it just ties it to always be a line color.

packages/core/components/DragDropContext/index.tsx
FedericoBonel: Yeah sounds good! I can't think of a situation where window is undefined and you're still able to use Puck, so will do that

packages/core/components/DropZone/styles.module.css
FedericoBonel: Updated it

packages/core/components/DropZone/LinePlaceholder.tsx
FedericoBonel: Added a new `--puck-line-placeholder-width` variable to theming for this and the outline.

apps/docs/pages/docs/api-reference/components/puck.mdx
FedericoBonel: Done!

packages/core/components/DragDropContext/index.tsx
FedericoBonel: Ended up reverting this, it won't be reached, and in the very fringe situation it happens, it will show a line preview as expected.

packages/core/components/DragDropContext/use-drop-animation.ts
```diff
@@ -0,0 +1,47 @@
...
+import type { ZoneStore } from "../DropZone/context";
+
+/**
+ * Builds the "glide" drop animation callback for components in the canvas/drawer.
+ *
+ * @param zoneStore - The store for the drop zone where the item is being dropped.
+ * @param id - The ID of the item being dropped. Provide it if using within created components.
+ * @returns A function that can be used as the drop animation callback.
+ */
+export const useDropAnimation = (
```
chrisvxd, line 16: Hm didn't notice this before, but I've noticed some pretty jittery animations in Safari Version 26.5 (21624.2.5.11.4) during the drop + resize.

https://github.com/user-attachments/assets/36faabd9-507c-435e-9d85-e6d5f8486be9

Can you find a way to improve this?

chrisvxd (changes requested):
Unfortunately I found one more issue.

chrisvxd (approved):

PR #1793, opened 2026-08-11
Support field transforms within render

### LINKED ISSUE #1780: Support field transforms within `Render`
## Description

Currently field transforms are only used within `Puck`, but it would be useful if we could provide it to the `Render` component.

This would make it possible to apply field transformations at render time after a user (or AI) has created a page. 

`resolveData` doesn't work in a lot of these cases since it pollutes the data payload and it expects JSON returned from it, whereas field transforms could return components and be computed every time.

## Proposals

Add the [`fieldTransforms` API](https://puckeditor.com/docs/api-reference/field-transforms) to the Render component:

```tsx
export function Page() {
  return (
    <Render
      fieldTransforms={{
        text: ({ value }) => <Heading body={value} />, // Convert all text values into `Heading` components
      }}
    />
  );
}
```

### LINKED ISSUE #1794: Inline fields don't work within root
## Description

Setting `contentEditable: true` doesn't allow for inline editing.
The reason this is happening is because the root editor is rendered  using only slot and user transforms.

## Environment

- Puck version: 0.23.0

FedericoBonel:
Closes #1780, Closes #1794

## Description

This PR adds support for the [`fieldTransforms` API](https://puckeditor.com/docs/api-reference/field-transforms) to the `Render` component. Field transforms were previously only available inside `Puck`; now the same API can be applied at render time, after a page has been created.

This makes it possible to transform stored field values into components or computed values on every render, without polluting the saved data. Unlike `resolveData`, transforms can return React components (not just JSON) and re-run whenever the transforms or data change.

Transforms run through the existing `useSlots` walk (the same one that resolves slots), so props are transformed in a single pass at every depth: root props, component props, and slot children. Both render paths are supported: the client `Render` (which shares state through context) and the RSC `Render` from the `rsc` bundle (which prop-drills, so nothing pulls the page to the client). 

Inside the editor, the interactive preview and the edit canvas run the same transforms, so the preview stays consistent with production output.

## Changes made

- `Render` and the RSC `Render` (`ServerRender`) accept an optional `fieldTransforms` prop and thread it through `useSlots`, `SlotRender`, and the dropzones, so it applies to root props, component props, and slot children at every depth.
  - `useSlots` now merges the built-in slot transform with the caller's transforms (`{ ...slotTransform, ...fieldTransforms }`), so user transforms compose with slots and can override them. 
    - It also forces `isReadOnly`, since nothing is editable under `Render`. 
    - This hook used to be used for edit mode root slots, but it doesn't use it any longer. So I'm defaulting this hook to be the `render` path for transforming slots.
- The render rich text hook (`useRichtextProps`) is now transform-aware: it skips any field whose type has a user transform. Without this, the richtext pass would descend into a value a transform had already replaced, produce corrupt props, and crash React.
- Fixed a pre-existing bug where a root with both a richtext field and a slot field dropped its resolved slot props, because the richtext pass was fed the raw props instead of the slot-resolved ones.
- Split the editor preview into an interactive path (`Render` with transforms) and an edit path.
- Fixed a pre-existing bug where field transforms were not applied to root fields.

## How to test

- Render a page with a `text` transform that converts every text value into a component with red text, and confirm it renders as that component instead of a raw string:

```tsx
import { Render } from "@puckeditor/core";

export function Page() {
  return (
    <Render
      config={config}
      data={data}
      fieldTransforms={{
        text: ({ value }) => <div style={{ color: "red" }}>{ value }</div>,
      }}
    />
  );
}
```

- Import `Render` from the `rsc` bundle and render it inside a Next.js Server Component with the same `fieldTransforms`; confirm the transforms run on the server and nothing is pulled to the client.
- In the editor, pass `fieldTransforms` to `Puck` and toggle between edit and interactive preview; confirm root and component props render identically in both modes.

## Summary by CodeRabbit

- **New Features**
  - Field transforms now apply consistently during editing and rendered output.
  - Added optional `fieldTransforms` support to `Render`, including nested content, slots, drop zones, and rich-text fields.
  - Transforms can replace values with rendered components and receive read-only context.

- **Bug Fixes**
  - Prevented rich-text processing from overriding values replaced by field transforms.

- **Documentation**
  - Updated API and extension guides with configuration details, examples, and `Render` read-only behavior.

### DIFF
--- a/apps/docs/pages/docs/api-reference/components/render.mdx
+++ b/apps/docs/pages/docs/api-reference/components/render.mdx
@@ -80,7 +80,7 @@ export function Example() {
 
 Specify transforms to modify field values before being passed to your components. Implements the [Field Transforms API](/docs/api-reference/field-transforms).
 
-Transforms may return components, unlike [`resolveData`](/docs/api-reference/configuration/component-config#resolvedatadata-params), which must return JSON and writes back to your [`Data`](/docs/api-reference/data-model/data) payload.
+Unlike [`resolveData`](/docs/api-reference/configuration/component-config#resolvedatadata-params), transforms may return components and other unserializable data.
 
 ```tsx {4} copy
 export function Example() {
--- a/apps/docs/pages/docs/api-reference/components/render.mdx
+++ b/apps/docs/pages/docs/api-reference/components/render.mdx
@@ -16,11 +16,12 @@ export function Example() {
 
 ## Props
 
-| Param                   | Example                      | Type                                        | Status   |
-| ----------------------- | ---------------------------- | ------------------------------------------- | -------- |
-| [`config`](#config)     | `config: { components: {} }` | [Config](/docs/api-reference/config)        | Required |
-| [`data`](#data)         | `data: {}`                   | [Data](/docs/api-reference/data-model/data) | Required |
-| [`metadata`](#metadata) | `metadata: {}`               | [Metadata](/docs/api-reference/metadata)    | -        |
+| Param                                 | Example                                  | Type                                                    | Status   |
+| ------------------------------------- | ---------------------------------------- | ------------------------------------------------------- | -------- |
+| [`config`](#config)                   | `config: { components: {} }`             | [Config](/docs/api-reference/config)                    | Required |
+| [`data`](#data)                       | `data: {}`                               | [Data](/docs/api-reference/data-model/data)             | Required |
+| [`fieldTransforms`](#fieldtransforms) | `fieldTransforms: {text: () => <div />}` | [FieldTransforms](/docs/api-reference/field-transforms) | -        |
+| [`metadata`](#metadata)               | `metadata: {}`                           | [Metadata](/docs/api-reference/metadata)                | -        |
 
 ## Required props
 
@@ -68,6 +69,25 @@ export function Example() {
           },
         ],
         root: {},
+      }}
+      // ...
+    />
+  );
+}
+```
+
+### `fieldTransforms`
+
+Specify transforms to modify field values before being passed to your components. Implements the [Field Transforms API](/docs/api-reference/field-transforms).
+
+Transforms may return components, unlike [`resolveData`](/docs/api-reference/configuration/component-config#resolvedatadata-params), which must return JSON and writes back to your [`Data`](/docs/api-reference/data-model/data) payload.
+
+```tsx {4} copy
+export function Example() {
+  return (
+    <Render
+      fieldTransforms={{
+        text: ({ value }) => <div>{value}</div>, // Wrap all text field values in a div
       }}
       // ...
     />
--- a/apps/docs/pages/docs/api-reference/components/render.mdx
+++ b/apps/docs/pages/docs/api-reference/components/render.mdx
@@ -87,7 +87,7 @@ export function Example() {
   return (
     <Render
       fieldTransforms={{
-        text: ({ value }) => <div>{value}</div>, // Wrap all text field values in a div
+        text: ({ value }) => <span>{value}</span>, // Wrap all text field values in a span
       }}
       // ...
     />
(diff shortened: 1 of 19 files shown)


apps/docs/pages/docs/api-reference/components/render.mdx
```diff
@@ -75,6 +76,25 @@ export function Example() {
 }
 ```
 
+### `fieldTransforms`
+
+Specify transforms to modify field values before being passed to your components. Implements the [Field Transforms API](/docs/api-reference/field-transforms).
+
+Transforms may return components, unlike [`resolveData`](/docs/api-reference/configuration/component-config#resolvedatadata-params), which must return JSON and writes back to your [`Data`](/docs/api-reference/data-model/data) payload.
```
chrisvxd: I think explanation of resolveData here feels a bit of a distraction

```suggestion
Unlike [`resolveData`](/docs/api-reference/configuration/component-config#resolvedatadata-params), transforms may return components and other unserializable data.
```

apps/docs/pages/docs/extending-puck/field-transforms.mdx
```diff
@@ -22,6 +16,22 @@ const fieldTransforms = {
...
+
+```tsx
+const fieldTransforms = {
+  text: ({ value }) => <div style={{ color: "blue" }}>{value}</div>, // Make all text fields blue
+};
+
+const Example = () => <Render fieldTransforms={fieldTransforms} />;
+```
+
+Transforms can return components, which makes them a good fit for non-serializable values that should not be baked into your [`Data`](/docs/api-reference/data-model/data) payload.
```
chrisvxd:
```suggestion
Transforms can return components, which makes them a good fit for unserializable values that should not be baked into your [`Data`](/docs/api-reference/data-model/data) payload.
```

apps/docs/pages/docs/extending-puck/field-transforms.mdx
```diff
@@ -22,6 +16,22 @@ const fieldTransforms = {
 const Example = () => <Puck fieldTransforms={fieldTransforms} />;
 ```
 
+## Transforming at render time
+
+Transforms can also be provided to [`<Render>`](/docs/api-reference/components/render) to apply them to a finished page.
```
chrisvxd: > to apply them to a finished page.

I find this phrasing a bit awkward

packages/core/components/Render/__tests__/field-transforms.spec.tsx
```diff
@@ -296,7 +296,10 @@ describe("Render fieldTransforms", () => {
               type: "List",
               props: {
```
chrisvxd, line 297: what is "scaped/unscaped HTML" in this commit message? Do you mean escaped?

packages/core/components/Puck/components/Preview/components/edit-page.tsx
```diff
@@ -1,4 +1,4 @@
-import { memo, useMemo } from "react";
+import { useMemo } from "react";
```
chrisvxd, line 1: I think "fix: apply inline and plugin field transforms to root fields" should be simply "fix: apply field transforms to root fields"

chrisvxd (changes requested):
excellent work, only a few comments on language and phrasing

apps/docs/pages/docs/extending-puck/field-transforms.mdx
FedericoBonel: Changed it to an active voice!

packages/core/components/Render/__tests__/field-transforms.spec.tsx
FedericoBonel: Yes, typo!

packages/core/components/Puck/components/Preview/components/edit-page.tsx
FedericoBonel: Changed it!

packages/core/components/Render/__tests__/field-transforms.spec.tsx
FedericoBonel: Changed it!

FedericoBonel:
Addressed your comments @chrisvxd!

PR #1796, opened 2026-08-12
Refactor edit/render field tranforms

FedericoBonel:
NOTE: Requires changes from #1793

## Description

This PR refactors the codebase to centralize field transformation in the editor and render paths.

While working on #1793, I realized there were a lot of inconsistencies and redundancies between the `root` render path and the component render path, so I decided to try to centralize the field transformation logic in a single hook.

I also realized the render path needed some cleanup, since `useSlots` and `useRichtextTransforms` were always called in the same order, and the goal of the double call is to apply all the required transforms for the render path. Since user transforms are now also available in `Render`, it made sense to extract that into a central hook that combines the logic and centralizes it in one place for parity with the edit path.

## Changes made

- Added a new `useEditorProps` hook that implements the logic for orchestrating editor field transforms for a single component.
- Added a new `useRenderProps` hook that implements the logic for orchestrating render field transforms for a single component.
- Removed the unnecessary `WithPuckProps` from `useRichtextProps`.

## Summary by CodeRabbit

* **Improvements**
  * Improved consistency when rendering components, slots, rich text, and root content across editor, preview, and server-rendered experiences.
  * Consolidated field transformations so editing and rendering apply component properties more reliably.
  * Preserved read-only and loading behavior while simplifying property handling.
  * Maintained existing rich-text discovery and slot rendering behavior.


packages/core/lib/field-transforms/use-editor-props.tsx
```diff
@@ -0,0 +1,72 @@
+import { useMemo } from "react";
+import { useShallow } from "zustand/react/shallow";
+
+import { DropZoneEditPure } from "../../components/DropZone";
+import { ContextSlotRender } from "../../components/SlotRender";
```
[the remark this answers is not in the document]
FedericoBonel: This was an existing problem. We need to refactor `DropZone` soon, it's a big file with a lot of merged responsibilities.

packages/core/components/RichTextEditor/lib/use-richtext-props.tsx
```diff
@@ -5,7 +5,6 @@ import {
   FieldTransforms,
```
chrisvxd, line 5: The second commit message isn't escaped so just shows "chore: remove unnecesary  from" (also typo in unnecessary)

packages/core/lib/props/render/use-render-props.tsx
```diff
@@ -0,0 +1,63 @@
+// This file should avoid importing any non RSC friendly code (state, effects, context, etc.), as it is used in both RSC and CSR contexts.
```
chrisvxd, line 1: These do various bits beyond just field transforms so I think they could probably move up a directory to `lib`

chrisvxd (changes requested):
Good stuff!

packages/core/lib/props/render/use-render-props.tsx
FedericoBonel: Moved it to `./lib/props/editor` and `./lib/props/render`

packages/core/components/RichTextEditor/lib/use-richtext-props.tsx
FedericoBonel: changed it!

FedericoBonel:
Addressed @chrisvxd!

PR #1801, opened 2026-08-14
Forward additional props to slot `as` component

### LINKED ISSUE #1441: Include additional props when using the `as` prop in Slots

FedericoBonel:
Closes puckeditor/puck#1441

## Description

Slots (and `Dropzone`/`puck.renderDropZone`) accept an `as` prop to render a custom element or component, but previously there was no way to pass props to the component. This forwards additional props not used by puck onto the as element/component, so you can pass e.g. library specific props:

```
render: ({ SlotComponent, height }) => (
  <SlotComponent as={Box} sx={{ height: (theme) => theme.units * height }} />
);
```

## Changes made

* Slot render components, `<DropZone>`, and `puck.renderDropZone` now forward any additional props to the element or component provided via `as` (defaulting to a `div`).
* The forwarded props are fully typed: `SlotComponent`, `renderDropZone`, and `<DropZone>` are generic over `as`, so `as="button"` only accepts button attributes, a custom component only accepts its own props, and unknown props are compile errors.
  * The `SlotProps` type was refactored from `Omit<DropZoneProps<T>, "zone">` to `DropZoneSharedProps` and `DropZoneOwnProps`.
    * This was done because using `Omit` erases the generic inference (`T`), collapsing `as` to the default `div` and letting any prop through.
  * The public `DropZone` component stays a div-typed `forwardRef` internally and is exported publicly with a cast type, since `forwardRef` can't use generics.
* Added `getDropZoneProps` (`lib/props/shared`) as the single formatting point that splits incoming public props into `props` (consumed by Puck), `forwardableProps` (spread onto `as`), and `nonForwardableProps` (dropped with a warning).
  * The benefit of doing this is that we avoid future props leakage bugs by forcing future devs to check this list and update it if needed.
  * This should be used anywhere `DropZoneProps` are read.
    * `DropZoneEdit`, `DropZoneRender`, and `SlotRender`, the three components where props reach the DOM, all use it instead of hand-rolled destructured lists.
  * Props are grouped by reference in a single pass, never cloned or merged, so downstream identity and memoization are unaffected.
  * The key maps are `Record<Key, true>` objects tied to the `ConsumedPropKey` and `NonForwardableProps` types, so adding a prop to the types without updating the runtime check (or vice versa) is a compile error.
  * Component-specific consumed props (`SlotRender`'s `config` and `metadata`) are passed as `extraConsumedKeys` so they're never forwarded to the underlying `as` component.
* Marked the `DropZone` export and `puck.renderDropZone` as `@deprecated` in favor of slots. Since that's what the docs clarify and what we want people to use.
* Added tests.
* Documented the forwarding behavior in the slot docs, including the table of props that are never forwarded and why.

## How to test

Add extra props alongside `as` in a slot render function and confirm they reach the element and that internal props don't leak:

```tsx
  <Content as="details" open data-custom="foo" />
);
```

## Summary by CodeRabbit

* **New Features**
  * Slot and drop zone props can now be forwarded to custom elements or components specified through `as`.
  * Added typed, component-specific props for slots and drop zones.
  * Internal props, `children`, and unsafe HTML content are excluded from forwarding.
* **Documentation**
  * Expanded Slot documentation with forwarding behavior and examples.
  * Marked the legacy DropZone and `renderDropZone` APIs as deprecated in favor of Slots.
  * Added migration guidance for replacing `renderDropZone` with Slots.

### DIFF
--- a/apps/docs/pages/docs/api-reference/fields/slot.mdx
+++ b/apps/docs/pages/docs/api-reference/fields/slot.mdx
@@ -231,6 +231,31 @@ const config = {
 };
 ```
 
+Any additional props are forwarded to the element or component provided via `as`.
+
+```tsx copy {5}
+const config = {
+  components: {
+    Example: {
+      render: ({ content: Content }) => {
+        return <Content as="details" open />;
+      },
+    },
+  },
+};
+```
+
+The following props are **not** forwarded to the `as` element, because Puck either consumes them internally or they are incompatible with slots:
+
+| Prop                      | Reason                              |
+| ------------------------- | ----------------------------------- |
+| `zone`                    | Identifies the drop zone internally |
+| `content`                 | Injected by Puck internally         |
+| `config`                  | Injected by Puck internally         |
+| `metadata`                | Injected by Puck internally         |
+| `dangerouslySetInnerHTML` | Isn't compatible with Puck slots    |
+| `children`                | Isn't compatible with Puck slots    |
+
 ### `className`
 
 Provide a className to the rendered element. The default styles will still be applied.
--- a/packages/core/lib/props/shared/get-drop-zone-props.ts
+++ b/packages/core/lib/props/shared/get-drop-zone-props.ts
@@ -1,0 +1,91 @@
+import type {
+  ConsumedPropKey,
+  NonForwardableProps,
+} from "../../../components/DropZone/types";
+
+type NonForwardableKey = keyof NonForwardableProps;
+
+type DropZonePropGroups<Props> = {
+  /** Props consumed by Puck internally. */
+  props: Pick<Props, Extract<keyof Props, ConsumedPropKey>>;
+  /** Props that should be forwarded to the element or component provided via `as`. */
+  forwardableProps: Omit<Props, ConsumedPropKey | NonForwardableKey>;
+  /** Props that shouldn't be forwarded to the component provided via `as` (e.g., `children` and `dangerouslySetInnerHTML`). */
+  nonForwardableProps: Pick<Props, Extract<keyof Props, NonForwardableKey>>;
+};
+
+// Key maps with all the consumed and non-forwardable keys -----------------------------------------------
+
+// Note to future maintainers: When adding new keys to the DropZone props, add them to the appropriate map below.
+//
+// `Record<Key, true>` makes these compile errors if a key
+// is added to (or removed from) the source types without updating them here, keeping
+// the runtime in sync with the type public interface.
+
+const CONSUMED_PROP_KEYS: Record<ConsumedPropKey, true> = {
+  zone: true,
+  allow: true,
+  disallow: true,
+  style: true,
+  minEmptyHeight: true,
+  className: true,
+  collisionAxis: true,
+  ref: true,
+  as: true,
+  content: true,
+  config: true,
+  metadata: true,
+};
+
+const NON_FORWARDABLE_PROP_KEYS: Record<NonForwardableKey, true> = {
+  children: true,
+  dangerouslySetInnerHTML: true,
+};
+
+const CONSUMED_KEY_SET: ReadonlySet<string> = new Set(
+  Object.keys(CONSUMED_PROP_KEYS)
+);
+const NON_FORWARDABLE_KEY_SET: ReadonlySet<string> = new Set(
+  Object.keys(NON_FORWARDABLE_PROP_KEYS)
+);
+
+/**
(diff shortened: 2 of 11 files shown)


packages/core/types/Props.tsx
```diff
@@ -1,9 +1,15 @@
...
 import { WithChildren, WithPuckProps } from "./Utils";
 
 export type PuckContext = {
-  renderDropZone: (props: DropZoneProps) => React.ReactNode;
+  /**
+   * @deprecated Use {@link https://puckeditor.com/docs/api-reference/fields/slot Slots} instead.
+   */
+  renderDropZone: <ComponentType extends ElementType = "div">(
+    props: DropZoneProps<ComponentType>
+  ) => React.ReactNode;
```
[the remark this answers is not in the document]
FedericoBonel: This was already existing in main, I think we need a follow up instead of fixing it here.

packages/core/lib/props/shared/get-drop-zone-props.ts
```diff
@@ -0,0 +1,91 @@
+import type {
```
chrisvxd, line 1: This is a weirdly deep directory. Just move it into `/lib`.

apps/docs/pages/docs/api-reference/fields/slot.mdx
```diff
@@ -231,6 +231,35 @@ const config = {
...
+| `zone`                    | Identifies the drop zone internally             |
+| `allow`                   | Controls which component types are allowed      |
+| `disallow`                | Controls which component types are disallowed   |
+| `collisionAxis`           | Configures drag collision detection             |
+| `minEmptyHeight`          | Sets the minimum height of an empty zone        |
+| `content`                 | Injected by Puck with the slot's component data |
+| `config`                  | Injected by Puck internally                     |
+| `metadata`                | Injected by Puck internally                     |
+| `dangerouslySetInnerHTML` | Isn't compatible with Puck slots                |
+| `children`                | Isn't compatible with Puck slots                |
```
chrisvxd: This table is going to be awkward to maintain. I'd at least remove the APIs that can be seen on the slot prop API, like allow and disallow

apps/docs/pages/docs/api-reference/fields/slot.mdx
```diff
@@ -231,6 +231,35 @@ const config = {
 };
 ```
 
+Any additional props are forwarded to the element or component provided via `as`, typed against it.
```
chrisvxd:
```suggestion
Any additional props are forwarded to the element or component provided via `as`.
```

chrisvxd (changes requested):
looks good, few comments

packages/core/lib/props/shared/get-drop-zone-props.ts
FedericoBonel: Happy to do it. The main reason I added it here is because I introduced `lib/props` in this PR: https://github.com/puckeditor/puck/pull/1796

Before rolling it back, let me explain my reasoning.

This would keep all prop transformation/processing in a single shared place. I currently find myself having a hard time finding these utilities, since some are in `/lib/data` and some are in `/lib`, and they normally have signatures and names that make it hard to know what they're used for without looking at the implementation.

So I was thinking of slowly organizing the `lib` folder a bit to make the intent clearer. I'll eventually add some JSDoc comments to them as well, so you don't have to read the implementation before using them.

I've also found agents constantly re-implementing them and, when asked, refactoring and apologizing. If I clean things up a bit, I can then add some guidance to the `AGENTS.md` file so they avoid this and organize their implementations better in the future (also making it easier for us to review).

What do you think?

apps/docs/pages/docs/api-reference/fields/slot.mdx
FedericoBonel: Done!

chrisvxd (approved):

PR #1808, opened 2026-08-17
fix: add accessible names to icon-only editor controls

### LINKED ISSUE #1807: Three icon-only editor controls have no accessible name
## Description

Three controls in the Puck editor render only an icon, with no text content and no `aria-label`. Screen readers announce them as an unlabelled "button" or "combo box", so it is not possible to tell what they do.

| Control | File |
|---|---|
| Detach external item (`<Unlock />`) | `packages/core/components/ExternalInput/index.tsx` |
| Add array item (`<Plus />`) | `packages/core/components/AutoField/fields/ArrayField/index.tsx` |
| Zoom level `<select>` | `packages/core/components/ViewportControls/index.tsx` |

This looks like an oversight rather than a decision — the neighbouring controls in each of those files are already labelled via `useMessage`, and `IconButton` sets `aria-label={title}`. These three seem to have been missed.

## Environment

- Puck version: 0.23.0 (`master` at 668cfb79)
- Browser version: any
- Additional environment info: n/a — visible in source

## Steps to reproduce

1. Open the Puck editor with a config containing an array field and an external field.
2. Navigate the fields panel with a screen reader (VoiceOver, NVDA) or inspect the accessibility tree in devtools.
3. Focus the add-item button, the detach button, or the zoom select.

## Expected behaviour

Each control is announced with a meaningful name, e.g. "Add item, button".

## Actual behaviour

Each is announced only by role — "button", "combo box" — with no name.

---

Happy to send a PR. The fix is additive: three new `defaultDictionary` keys following the existing `{area}-{name}-{variant}` convention, plus `aria-label` wired through `useMessage` in each component. No visual or behavioural change.

For disclosure: I found these with a static accessibility linter I maintain. I have deliberately left out the more debatable findings (the modal backdrop, the resize handle) since those are design calls rather than clear defects.

SomilKSharma:
Closes #1807

## Description

Three editor controls render only an icon, with no text content and no `aria-label`, so assistive technology announces them by role with no name — "button", "combo box".

This wires each through `useMessage`, matching how the neighbouring controls in the same files are already labelled.

## Changes made

| Control | File | New key |
|---|---|---|
| Detach external item (`<Unlock />`) | `ExternalInput/index.tsx` | `field-external-detach` |
| Add array item (`<Plus />`) | `AutoField/fields/ArrayField/index.tsx` | `field-arrayitem-add` |
| Zoom level `<select>` | `ViewportControls/index.tsx` | `viewport-zoom-level` |

Keys follow the documented `{area}-{name}-{variant}` convention and sit beside their existing siblings (`field-arrayitem-delete`, `field-external-selectdata`, `viewport-zoom-auto`), so they are overridable through the `dictionary` prop like every other string.

Additive only — 9 insertions, no deletions. No visual or behavioural change, and no new dependency.

## Testing

- `tsc --noEmit` on `packages/core`: 12 errors before this change, 12 after, none in the touched files.

I could not run the full suite locally — this machine is on Node 20.12.2 and the install requires `^20.19.0 || ^22.13.0 || >=24` — so CI is the real check here.

---

For disclosure: I found these with a static accessibility linter I maintain. I have deliberately left out its more debatable findings — the modal backdrop click handler and the resize handle — since those are design calls rather than clear defects, and I did not want to bundle judgement calls with an unambiguous fix.

## Summary by CodeRabbit

* **Accessibility**
  * Added localized accessible labels for adding array items, detaching external values, and selecting viewport zoom levels.
* **Localization**
  * Added default messages for the new interface labels to support translation and consistent wording.


packages/core/lib/dictionary.ts
```diff
@@ -39,6 +39,8 @@ export const defaultDictionary = {
   "field-arrayitem-summary": "Item #{index}",
   "field-arrayitem-duplicate": "Duplicate",
   "field-arrayitem-delete": "Delete",
+  "field-arrayitem-add": "Add item",
```
FedericoBonel, line 42: Can we remove the "item" segment to follow the same format as the other item messages?

FedericoBonel (changes requested):
Hey @SomilKSharma! Thanks for the contribution.

Left one comment!

FedericoBonel:
Also, we need to add the new strings to the dictionary docs in:
- [Array field](https://puckeditor.com/docs/api-reference/dictionary#array-field)
- [Viewport controls](https://puckeditor.com/docs/api-reference/dictionary#viewport-controls)

FedericoBonel:
Closing this as stale since there hasn't been any updates in the last 2 weeks. 

Feel free to push the required changes and I'll re-open the PR!

PR #1810, opened 2026-08-18
Add a drag handle to the action bar

### LINKED ISSUE #1477: Add a drag handle to the action bar

FedericoBonel:
Closes #1477

## Description

This PR builds on top of the initial feature introduced in #1770 and adds a few improvements:

- Rebases to latest changes from main
- Uses line previews when dragging from the action bar handle
- Keeps the action bar mounted while dragging from the handle
- Only starts drag from the handle after a few pixels of movement to prevent accidental drags.

## Changes made

- Added a `dnd.enableDragHandle` Puck prop to enable drag handles in the action bar, defaults to `false`.
- Added a new `ActionBar.DragHandle` component and exposed a `dragHandle` through the `actionBar` override to override the action bar while also being able to use the handle.
- Respected the component's drag permission when rendering the drag handle.
- Updated the drag sensors so both the component body and drag handle can initiate a drag.
- Added a `5px` mouse distance constraint when dragging from the handle to prevent accidental drags.
- Added `draggingFromHandle` to the `ZoneStoreContext` to track whether the current drag originated from the handle.
- Updated `resolveDndMode` so handle drags use static line previews when behavior is "auto", while explicit "fluid" and "static" behaviors remain unchanged.
- Kept the action bar mounted during handle drags and synchronized it with the dragged component so it follows the drag ghost through the drag and drop animation.
- Added the `action-drag` dictionary token.
- Added documentation for `dnd.enableDragHandle`, `ActionBar.DragHandle`, and the `dragHandle` override.

### Basic behavior

- Navigate to the demo.
- Click on a component.
- Drag it from the action bar.
- Verify that it works as expected.

### Nested components with parents that don't have any padding

- Add a flex container inside a grid.
- Add a card inside that flex container.
- Click on the card, then select the parent using the parent option in the action bar.
- Use the handle to drag the flex container.

## Summary by CodeRabbit

* **New Features**
  * Added an action-bar drag handle for repositioning components.
  * Enabled custom drag handles with clear static drop indicators.
  * Added accessible labels, disabled states, and a “Drag to reorder” tooltip.
  * Improved drop-target highlighting with customizable theme colors.

* **Documentation**
  * Documented drag-handle components, configuration, behavior, and customization options.

* **Tests**
  * Added coverage for drag-handle rendering, interaction, accessibility, and drag behavior.

### DIFF
--- a/apps/demo/app/styles.css
+++ b/apps/demo/app/styles.css
@@ -61,6 +61,11 @@ body {
       var(--puck-color-azure-06) 22%,
       transparent
     );
+    --puck-color-drop-target-bg: color-mix(
+      in srgb,
+      var(--puck-color-azure-05) 22%,
+      transparent
+    );
     --puck-color-selection-border: var(--puck-color-azure-06);
 
     --puck-color-bg-disabled: #252527;
--- a/apps/docs/pages/docs/api-reference/components/puck.mdx
+++ b/apps/docs/pages/docs/api-reference/components/puck.mdx
@@ -144,12 +144,13 @@ Configure drag-and-drop behavior.
 | [`behavior`](#behavior)                     | `behavior: "static"`       | `"auto" \| "fluid" \| "static"` | -      |
 | [`disableAutoScroll`](#disableautoscroll)   | `disableAutoScroll: true`  | boolean                         | -      |
 | [`disableOutlineDrag`](#disableoutlinedrag) | `disableOutlineDrag: true` | boolean                         | -      |
+| [`enableDragHandle`](#enabledraghandle)     | `enableDragHandle: true`   | boolean                         | -      |
 
 ##### `behavior`
 
 Control animation behavior while dragging. Defaults to `"auto"`.
 
-- `"auto"` (default): Use fluid behavior within a slot, then switch to static behavior when dragging an item between slots or inserting a new component.
+- `"auto"` (default): Use fluid behavior within a slot, then switch to static behavior when dragging an item between slots, inserting a new component, or dragging from the [drag handle](#enabledraghandle).
 - `"fluid"`: Move the item during a drag, animating other items out of the way.
 - `"static"`: Freeze the item in-place during a drag, showing a line placeholder at the drop destination.
 
@@ -160,6 +161,10 @@ Disable auto-scroll when the user drags an item near the edge of the preview are
 ##### `disableOutlineDrag`
 
 Disable drag-and-drop in the [outline](/docs/api-reference/components/puck-outline). Items can still be selected and expanded.
+
+##### `enableDragHandle`
+
+Show a drag handle in the component action bar. Defaults to `false`.
 
 ### `fieldTransforms`
 
--- a/apps/docs/pages/docs/api-reference/theming/computed-values.mdx
+++ b/apps/docs/pages/docs/api-reference/theming/computed-values.mdx
@@ -189,6 +189,7 @@ These tables document the default computed value each component token resolves t
 | Token                                         | Computed default value                                      |
 | --------------------------------------------- | ----------------------------------------------------------- |
 | `--puck-slot-color-bg`                        | <Color>color-mix(in srgb, #cfdff0 30%, transparent)</Color> |
+| `--puck-slot-color-target-bg`                 | <Color>color-mix(in srgb, #f3f6fb 30%, transparent)</Color> |
 | `--puck-slot-color-border`                    | <Color>#abc7e5</Color>                                      |
 | `--puck-slot-border-width`                    | `2px`                                                       |
 | `--puck-slot-border-style`                    | `dashed`                                                    |
--- a/packages/core/components/ActionBar/index.tsx
+++ b/packages/core/components/ActionBar/index.tsx
@@ -1,4 +1,5 @@
 import { forwardRef, ReactNode, SyntheticEvent } from "react";
+import { GripVertical } from "lucide-react";
 import getClassNameFactory from "../../lib/get-class-name-factory";
 import styles from "./styles.module.css";
 const getClassName = getClassNameFactory("ActionBar", styles);
@@ -52,6 +53,29 @@ export const Action = forwardRef<
 
 Action.displayName = "Action";
 
+export const DragHandle = forwardRef<
+  HTMLButtonElement,
+  {
+    label?: string;
+    disabled?: boolean;
+  }
+>(({ label, disabled, ...rest }, ref) => (
+  <button
+    type="button"
+    {...rest}
+    ref={ref}
+    className={getActionClassName({ disabled, grab: !disabled })}
+    onClick={(e) => e.stopPropagation()}
+    title={label}
+    tabIndex={0}
+    disabled={disabled}
+  >
+    <GripVertical className={getActionClassName("icon")} />
+  </button>
+));
+
+DragHandle.displayName = "DragHandle";
+
 export const Group = ({ children }: { children: ReactNode }) => (
   <div className={getClassName("group")}>{children}</div>
 );
@@ -66,3 +90,4 @@ ActionBar.Action = Action;
 ActionBar.Label = Label;
 ActionBar.Group = Group;
 ActionBar.Separator = Separator;
+ActionBar.DragHandle = DragHandle;
--- a/packages/core/components/DragDropContext/index.tsx
+++ b/packages/core/components/DragDropContext/index.tsx
@@ -1,5 +1,3 @@
-import { DragDropProvider } from "@dnd-kit/react";
-import { useAppStore, useAppStoreApi } from "../../store";
 import {
   createContext,
   Dispatch,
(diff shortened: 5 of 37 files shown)


chrisvxd (changes requested):
Haven't done a code review, but this breaks the nested hit area collision, especially noticeable with small slot targets.

For example, go to the buttons near the bottom of the demo page, drag with the handle, and try to move the button back into the flex. You'll notice it's hard to get the hit area to activate, because the NestedDroppablePlugin is disabling to reduce layout shift.

However, I'm wondering if the NestedDroppablePlugin should be disabled when using non-fluid (aka auto or static) dragging altogether, with all areas enabled by default, since layout shift is no longer an issue.

I suspect that would fix the issue above.

FedericoBonel:
@chrisvxd! Ah nice! I didn't know that was what that did 🙏.

Will try it out, I also noticed that limitation and couldn't really understand why it was having issues, specifically with the action bar drag and not with the component drag, was going to try and triage it, but that might be it, thank you.

packages/core/components/ActionBar/index.tsx
```diff
@@ -52,6 +53,29 @@ export const Action = forwardRef<
...
+>(({ label, disabled, ...rest }, ref) => (
+  <button
+    type="button"
+    {...rest}
+    ref={ref}
+    className={getActionClassName({ disabled, grab: !disabled })}
+    onClick={(e) => e.stopPropagation()}
+    title={label}
+    tabIndex={0}
+    disabled={disabled}
```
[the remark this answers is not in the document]
FedericoBonel: This is an existing issue with any dropzone element, accessibility is currently a problem across the board for dnd.

Worth a follow up ticket, but probably not part of this PR

FedericoBonel:
@chrisvxd, this should be fixed now. The problem was that the dynamic collision algorithm measured collisions from the dragged components instead of the actionbar. 

Added a flag to the algorithm to signal when we should measure from the cursor instead, and when that is the case I repositioned the component shape around the cursor for detection.

apps/docs/pages/docs/api-reference/components/puck.mdx
```diff
@@ -164,7 +164,7 @@ Disable drag-and-drop in the [outline](/docs/api-reference/components/puck-outli
 
 ##### `enableDragHandle`
 
-Show a drag handle in the component action bar. When [`behavior`](#behavior) is `"auto"`, drags started from the handle show a static line placeholder instead of animating sibling items.
+Show a drag handle in the component action bar. Defaults to `false`. When [`behavior`](#behavior) is `"auto"`, drags started from the handle show a static line placeholder instead of animating sibling items.
```
chrisvxd: I don't know if we need the verbosity, as the behavior is documented separately.

```suggestion
Show a drag handle in the component action bar. Defaults to `false`.
```

apps/demo/app/styles.css
```diff
@@ -61,6 +61,11 @@ body {
       var(--puck-color-azure-06) 22%,
       transparent
     );
+    --puck-color-drop-target-bg: color-mix(
+      in srgb,
+      var(--puck-color-azure-08) 22%,
+      transparent
+    );
```
chrisvxd, line 68: It should go darker, not lighter. And try to stick with the actual values without introducing color-mix.

apps/docs/pages/docs/api-reference/theming/computed-values.mdx
```diff
@@ -189,6 +189,7 @@ These tables document the default computed value each component token resolves t
 | Token                                         | Computed default value                                      |
 | --------------------------------------------- | ----------------------------------------------------------- |
 | `--puck-slot-color-bg`                        | <Color>color-mix(in srgb, #cfdff0 30%, transparent)</Color> |
+| `--puck-slot-color-target-bg`                 | <Color>color-mix(in srgb, #f3f6fb 30%, transparent)</Color> |
```
chrisvxd, line 192: This is a theming API change. Don't forget to `feat` it.

packages/core/components/DragDropContext/index.tsx
```diff
@@ -315,7 +325,14 @@ const DragDropContextClient = ({
...
   ]);
 
-  const sensors = useSensors();
+  // Allow dragging from the component body or a registered handle (action bar).
+  const sensors = useSensors({
+    activatorElements: (source) =>
+      source.handle && source.handle !== source.element
+        ? [source.element, source.handle]
+        : [source.element],
+    mouse: [new PointerActivationConstraints.Distance({ value: 5 })],
```
chrisvxd: Doesn't this override the default sensors for the regular drag? I'm not sure we should change that behaviour.

packages/core/components/DropZone/index.tsx
```diff
@@ -436,6 +442,8 @@ export const DropZoneEdit = forwardRef<HTMLDivElement, DropZoneProps>(
     const isZoneEmpty = contentIdsWithPreview.length === injectedPreviewCount;
     const isDropEnabled = isEnabled && isZoneEmpty;
 
+    // const isTargettingArea = isDragging && isEnabled && hoveringOverArea;
```
chrisvxd: Rogue comment

packages/core/lib/dnd/use-sensors.ts
```diff
@@ -17,29 +18,50 @@ const otherDefault: ActivationConstraints = [
...
     touch: touchDefault,
     other: otherDefault,
   }
 ) => {
   const [sensors] = useState(() => [
     PointerSensor.configure({
+      activatorElements,
       activationConstraints(event, source) {
-        const { pointerType, target } = event;
+        onPointerDown?.(event);
```
chrisvxd: 🤔

packages/core/components/DragDropContext/index.tsx
```diff
@@ -12,40 +10,50 @@ import {
...
+import { useAppStore, useAppStoreApi } from "../../store";
 import { getItem } from "../../lib/data/get-item";
-import {
-  DropZoneContext,
-  Preview,
-  RootVirtualizerHandle,
-  ZoneStore,
-  ZoneStoreProvider,
-} from "../DropZone/context";
 import { createNestedDroppablePlugin } from "../../lib/dnd/NestedDroppablePlugin";
```
chrisvxd, line 28: This plugin causes a delay when switching area, because it was originally designed for the fluid drag mode to avoid layout shift.

With auto/static mode (or the drag handle), there is no layout shift, and so the delay is no longer necessary.

I think we need to either disable that delay, or disable the plugin entirely, in those scenarios.

chrisvxd (changes requested):
Works great, some more thoughts

packages/core/components/DropZone/index.tsx
FedericoBonel: Good catch!

packages/core/components/DragDropContext/index.tsx
FedericoBonel: Right! I actually tried it out originally as you suggested, but I found the delay + buffer area we add around dropzones within this plugin to be quite useful.

For example, if you have a grid that contains a flex as the first item, both without any padding, and you add another component as the second item in the grid, it's quite hard to move the second component to the first position because there's no target area around it (the flex and grid dropzones don't have any spacing between them) so it keeps falling into the nested flex instead of the first position. You can always move the flex instead of the second item, but it's a bit counter intuitive. With the plugin's delay + buffer, you have an easier time targeting it.

packages/core/lib/dnd/use-sensors.ts
FedericoBonel: Ah sorry, left over of a previous approach I was taking. Good catch.

packages/core/components/DragDropContext/index.tsx
FedericoBonel: Interestingly we actually were only applying mouse activators if we used a handle drag but I agree we might want to make this more explicit

apps/docs/pages/docs/api-reference/components/puck.mdx
FedericoBonel: Ah yes, I actually hadn't change the docs from the original PR, reworking now.

apps/demo/app/styles.css
FedericoBonel: Ah nice one! Made it darker here, but did you also mean that we should make it darker in the default `@puckeditor/core` theme? or lighter for that one is fine?

Also, I decided to use a color-mix, for opacity, if I don't add that, then the whole dropzone turns into a solid color. We also are already doing that for the resting background and the overlay.

apps/docs/pages/docs/api-reference/components/puck.mdx
FedericoBonel: Used your suggestion!

apps/docs/pages/docs/api-reference/theming/computed-values.mdx
FedericoBonel: You're right! Will do 🙏

packages/core/components/DragDropContext/index.tsx
FedericoBonel: Changed it up so it now each pointer accept a callback too. That way we can set it up dynamically based on the source and the event

FedericoBonel:
This should be ready for another review @chrisvxd!

Basically:

- Refactored `useDragHandle` into two hooks, since I realized I was doing a lot of duplicated, complicated, and brittle dragging state-tracking logic that we should have been handling directly in the drag-and-drop context:
  - `useDragHandle`: Same as before, but it now only positions the overlay with the drag handle on top of the dragged component and "registers" the drag handle element (adds a `data-puck-drag-handle` attribute to the ref) so that we can detect clicks on it. It is also zone-store independent.
  - `useDragHandleListener`: Tracks clicks on drag handles and returns a function to check whether we're currently dragging from the handle. We then use this to update the zone store state directly in the dnd-kit events. The hook docs explain a bit about why we can't use the event from dnd-kit.
- Refactored `useSensors` to also allow callbacks besides plain arrays, so that we can dynamically resolve them based on the source and event triggering the sensor call.
- Made the demo dark mode render slots darker when targeting them. I need you to confirm whether you meant for the slot to go darker only in the demo dark mode or in general as well.

chrisvxd (commented):
Works great, I'm just not sure about the position of the drag handle. I wonder if it would be better adjacent to the label, with the whole label being draggable

Image

The only issue is that it's then a bit messy when the parent selector is there

Image

What do you think?

FedericoBonel:
I don't dislike it! Will do it like that, the whole label being draggable should also help with UX.

PR #1813, opened 2026-08-21
Expose inline text field component

### LINKED ISSUE #1252: Expose the `InlineTextField` component
## Description

Currently, the only way to use inline text editing is through the `contentEditable` property, which is simple but doesn’t cover all use cases, such as enabling inline text editing on a nested key inside a custom object field.

For example:

```tsx
const componentConfig = {
  fields: {
    customObject: {
      type: "custom",
        // There's no way to support inline text editing out of the box for this field
        return (
          <div>
            <input
              value={value}
              onChange={(e) =>
                onChange({ ...value, inline: e.target.value })
              }
            />
          </div>
        );
      },
    },
  },
};
```

While it’s currently possible to work around this by implementing custom inline editing logic, that approach can lead to inconsistencies with Puck's implementation. It’s also difficult to trigger data resolvers when an inline field changes, since we don’t have access to `resolveComponentData`, and `InlineTextField` always handles that internally.

Exposing the `InlineTextField` component would make this less cumbersome, and consistent with the rest of the Puck logic.

## Considerations

* If puckeditor/puck#1249  gets implemented and `resolveComponentData` is exposed, we should consider whether we still want to always run data resolvers internally.

FedericoBonel:
Closes puckeditor/puck#1252

## Description

This PR exposes the inline text field component to allow users to create new inline editable fields using field transforms.

## Changes made

* Added new docs for the component
* Exposed the component from the core bundle

## Summary by CodeRabbit

* **New Features**
  * Added the `InlineTextField` component to the public component library.
  * Supports inline text editing through field transforms, with configurable line-break behavior.
* **Documentation**
  * Added comprehensive `InlineTextField` documentation, including an interactive example, props, usage guidance, and related resources.
  * Added the component to the API reference navigation.

## Summary by CodeRabbit

* **New Features**
  * Added `InlineTextField` to the public component library.
  * Supports inline editing through field transforms, with configurable line-break behavior.

* **Documentation**
  * Added comprehensive usage documentation, interactive examples, prop details, and related links.
  * Added `InlineTextField` to the component reference and navigation menu.

### DIFF
--- a/apps/docs/pages/docs/api-reference/components/inline-text-field.mdx
+++ b/apps/docs/pages/docs/api-reference/components/inline-text-field.mdx
@@ -1,0 +1,129 @@
+---
+title: <InlineTextField>
+---
+
+import { PuckPreview } from "@/docs/components/Preview";
+import { Puck, InlineTextField } from "@/puck";
+import { Callout } from "nextra/components";
+
+# \<InlineTextField\>
+
+Render inline-editable text from a [field transform](/docs/api-reference/field-transforms).
+
+This is the same component that powers the [`contentEditable`](/docs/api-reference/fields/text#contenteditable) option internally. Use it to create new [inline-editable fields](/docs/extending-puck/ui-overrides#introducing-new-field-types).
+
+<PuckPreview
+  label="Example"
+  config={{
+    components: {
+      Example: {
+        fields: {
+          heading: { type: "heading" },
+        },
+        render: ({ heading }) => <h2>{heading}</h2>,
+      },
+    },
+  }}
+  data={{
+    content: [
+      {
+        type: "Example",
+        props: {
+          id: "example",
+          heading: "Edit me inline",
+        },
+      },
+    ],
+    root: { props: {} },
+  }}
+  overrides={{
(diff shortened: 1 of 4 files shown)


apps/docs/pages/docs/api-reference/components/inline-text-field.mdx
```diff
@@ -0,0 +1,157 @@
...
+    subtitle: (
+      <InlineTextField
+        value={value.subtitle}
+        componentId={componentId}
+        propPath={`${propPath}.subtitle`}
+        isReadOnly={isReadOnly}
+      />
+    ),
+  }),
+};
```
chrisvxd, line 71: Two examples is unnecessary. Reduce to one.

chrisvxd (changes requested):

apps/docs/pages/docs/api-reference/components/inline-text-field.mdx
FedericoBonel: Left it as 1!

chrisvxd (approved):

PR #1815, opened 2026-08-27
Clarify that field transforms that use components need to be rendered instead of passed

### LINKED ISSUE #1251: Hooks can't be used inside fieldTransforms
## Description

Using any hooks inside field transforms throws an error. This includes copying and pasting the example of how to use field transforms from the docs, and also using the `PuckAPI`.

This might be happening because of the use of `useMemo` when initializing fieldTransforms [here](https://github.com/puckeditor/puck/blob/51863fb8912aa64b61c48cceac4f1930ce5799e0/packages/core/components/Puck/index.tsx#L297).

## Environment

- Puck version: 0.20.0

## Steps to reproduce

1. Copy and paste the ["Making it interactive"](https://puckeditor.com/docs/extending-puck/field-transforms#making-it-interactive) example from the `fieldTransforms` docs, which uses two hooks (`useEffect` and `useRef`):

```tsx
import { registerOverlayPortal } from "@measured/puck";

const EditableText = ({ value }) => {
  const ref = useRef(null);

  useEffect(() => {
    if (ref.current) {
      // Register the element as an overlay portal
      registerOverlayPortal(ref.current);
    }
  }, [ref.current]);

  return (
    // Mark the element as editable for inline text editing
    <p ref={ref} contentEditable>
      {value}
    </p>
  );
};

const fieldTransforms = {
  text: EditableText,
};

const Example = () => <Puck fieldTransforms={fieldTransforms} />;
```

2. Add a config with a component that uses the transformed field (in this case, a `text` field):

```tsx
const Example = () => (
  <Puck
    config={{
      components: {
        Heading: {
          fields: { title: { type: "text" } },
          defaultProps: { title: "Something" },
          render: ({ title }) => {
            return <div>{title}</div>;
          },
        },
      },
    }}
    fieldTransforms={fieldTransforms}
  />
);
```

3. Navigate to the editor  
4. Drag and drop the component that uses the transformed field

## What happens

An error is thrown:

```bash
Do not call Hooks inside useEffect(...), useMemo(...), or other built-in Hooks. 
You can only call Hooks at the top level of your React function. 
For more information, see https://react.dev/link/rules-of-hooks
```

## What I expect to happen

Hooks should be usable inside field transforms without breaking the application.

## Additional Media

Image

FedericoBonel:
Closes #1251

## Description

This PR addresses parts of the docs that were mistakenly encouraging users to follow anti-patterns in field transforms.

The docs were telling people to pass component references directly to transforms instead of rendering the component from the transform. This is an anti-pattern because transforms are called internally as plain functions within `useMemo`. For plain values, that's fine, but for components, that's not correct because it merges the render context (state tracking, hook usage, DOM access, etc.) of the user-defined components with the internal Puck ones. Basically, React doesn't "see" two components. It just sees the component code as a function call within our internal rendering flow.

That's wrong for a number of different reasons, performance being one of them, but it also makes React throw when components passed directly to transforms use hooks, since those hooks get bubbled up to the parent and can't be used within `useMemo`.

Since we can't know whether the user will return a value or a component, and therefore can't know whether we need to call the transform as JSX or as a plain function, I decided to fix the docs to clarify how to render components instead of providing an additional API that would just wrap what React already handles.

This is not an issue with things like component config or override render functions because we always call those as JSX, since we know they will always be components.

## Changes made

- Documented that transforms are called as functions, so when using components, users should render the component instead of passing it directly.
- Fixed the snippet that originally surfaced this problem to use the correct pattern.

## Summary by CodeRabbit

* **Documentation**
  * Updated field transform examples to demonstrate uppercase text transformations.
  * Clarified that transforms are invoked as functions, including guidance for interactive components using hooks.
  * Improved rich text plugin examples by simplifying inline editing implementation and correcting explanatory text.

### DIFF
--- a/apps/docs/pages/docs/api-reference/field-transforms.mdx
+++ b/apps/docs/pages/docs/api-reference/field-transforms.mdx
@@ -7,8 +7,10 @@ title: FieldTransforms
 Transform the data for each [field type](/docs/api-reference/fields) before rendering in the editor.
 
 ```tsx copy
+const UppercaseText = ({ value }) => <p>{value.toUpperCase()}</p>;
+
 const fieldTransforms = {
-  text: ({ value }) => <p>{value}</p>,
+  text: (props) => <UppercaseText {...props} />,
   // ...
 };
 ```
--- a/apps/docs/pages/docs/extending-puck/field-transforms.mdx
+++ b/apps/docs/pages/docs/extending-puck/field-transforms.mdx
@@ -16,7 +16,7 @@ Specify a transforms object for the fields you want to modify before rendering:
 
 ```tsx
 const fieldTransforms = {
-  text: ({ value }) => <div>Value: {value}</div>, // Wrap all text field props in divs
+  text: ({ value }) => value.toUpperCase(), // Transform text fields to uppercase
 };
 
 const Example = () => <Puck fieldTransforms={fieldTransforms} />;
@@ -48,11 +48,13 @@ const EditableText = ({ value }) => {
 };
 
 const fieldTransforms = {
-  text: EditableText,
+  text: (props) => <EditableText {...props} />,
 };
 
 const Example = () => <Puck fieldTransforms={fieldTransforms} />;
 ```
+
+Because Puck calls transforms as functions, render components that use hooks from the transform instead of passing them directly.
 
 ## Define new fields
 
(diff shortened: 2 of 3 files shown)


chrisvxd (commented):
Looks good but I'm just wondering again if there's a way we can support the original examples with a change to the field transform behaviour 🤔

FedericoBonel:
I know what you mean, I had a thought, but the problem is that we don't know what the user will return from the function, so we can't know if we need to use the transform as a component (either through JSX or React.createElement) or not. <br><br>So in order for us to know we would have to call the transform as a function, but if we call the transform we then merge render contexts and that opens a full can of worms with other issues that have not surfaced but could.

chrisvxd (approved):

PR #1819, opened 2026-08-29
feat: add inline field placeholders

### LINKED ISSUE #1795: Add inline placeholders
## Description

Currently, we have a `placeholder` option for `textarea`, `text`, and `number`. However, applying this option doesn't have any effect on inline editing, it only shows in the sidebar.

To achieve something similar, you can apply a default prop or a `render` default value.

However, it would be beneficial to also show the placeholder inline in the editor only. This would make it easier to guide users without polluting the data payload.

## Proposals

Show the placeholder prop inline when using `contentEditable: true`.

## Considerations

- We should be careful with the placeholder font color. Applying a hardcoded grey might not be feasible in this case. It could be better to simply apply a lower `opacity` and inherit the font color that's currently being used in the document.

ujwalbholan:
Closes #1795

## Description

Adds inline placeholder display for `text` and `textarea` fields when `contentEditable` is enabled. Placeholders now appear directly in the editor, using a reduced opacity to distinguish them from actual content. The placeholder is purely visual and is not saved to the component data payload.

**Approach:**
- Passes `field.placeholder` to `InlineTextField` component from the text and textarea transforms
- Uses `data-placeholder` attribute and CSS `::before` pseudo-element with `attr()` to render the placeholder text
- Applies reduced opacity (0.5) to visually indicate the placeholder is not actual content
- Inherits the current text color from the component for style consistency

## Changes made

- **InlineTextField component** (`packages/core/components/InlineTextField/index.tsx`):
  - Added optional `placeholder?: string` prop to component signature
  - Set `data-placeholder` attribute on the editable span

- **Inline text transform** (`packages/core/lib/field-transforms/default-transforms/inline-text-transform.tsx`):
  - Pass `placeholder={field.placeholder}` to `InlineTextField` in both `text` and `textarea` transforms

- **Inline text styles** (`packages/core/components/InlineTextField/styles.module.css`):
  - Added `.InlineTextField:empty::before` rule to display placeholder text with reduced opacity when the field is empty

- **InlineTextField tests** (`packages/core/components/InlineTextField/__tests__/index.spec.tsx`):
  - Updated test to verify `data-placeholder` attribute is correctly set when value is empty
  - Confirmed placeholder does not appear in the actual text content

## How to test
  - ``` yarn workspace @puckeditor/core test --runInBand components/InlineTextField/__tests__/index.spec.tsx```

##visual evidences
https://github.com/user-attachments/assets/e802e2a0-57b5-4d70-9212-355c82bc6c76

1. **Update the demo Card component** temporarily to include placeholders and empty default values:
   ```tsx
   // In apps/demo/config/blocks/Card/index.tsx
   title: {
     type: "text",
     contentEditable: true,
     placeholder: "Enter a title",
   },
   description: {
     type: "textarea",
     contentEditable: true,
     placeholder: "Enter a description",
   },
   defaultProps: {
     title: "",
     description: "",
     icon: "Feather",
     mode: "flat",
   },

<!-- This is an auto-generated comment: release notes by coderabbit.ai -->
## Summary by CodeRabbit

* **New Features**
  * Inline text fields now display configured placeholder text when empty.
  * Placeholder text uses subdued styling and does not interfere with interaction.
  * Demo Card and Hero fields now provide guidance such as “Enter a title” when blank.

* **Bug Fixes**
  * Empty inline fields no longer retain browser-inserted filler content.
  * Placeholder behavior is consistent across text and textarea fields.
<!-- end of auto-generated comment: release notes by coderabbit.ai -->

### DIFF
--- a/packages/core/components/InlineTextField/__tests__/index.spec.tsx
+++ b/packages/core/components/InlineTextField/__tests__/index.spec.tsx
@@ -30,19 +30,21 @@ jest.mock("../../../lib/overlay-portal", () => ({
 jest.mock("../styles.module.css");
 
 describe("InlineTextField", () => {
-  it("renders empty when value is null instead of the literal string 'null'", () => {
+  it("renders the placeholder as an attribute when value is empty", () => {
     const { container } = render(
       <InlineTextField
         propPath="subtitle"
         componentId="comp-1"
         value={null}
         isReadOnly={false}
+        placeholder="Enter a subtitle"
       />
     );
 
     const span = container.querySelector("span");
     expect(span).not.toBeNull();
     expect(span?.innerText ?? span?.textContent ?? "").toBe("");
+    expect(span).toHaveAttribute("data-placeholder", "Enter a subtitle");
   });
 
   it("renders empty when value is undefined instead of the literal string 'undefined'", () => {
@@ -52,6 +54,7 @@ describe("InlineTextField", () => {
         componentId="comp-1"
         value={undefined}
         isReadOnly={false}
+        placeholder="Enter a subtitle"
       />
     );
 
@@ -67,6 +70,7 @@ describe("InlineTextField", () => {
         componentId="comp-1"
         value="Hello world"
         isReadOnly={false}
+        placeholder="Enter a subtitle"
       />
     );
 
--- a/packages/core/components/InlineTextField/__tests__/index.spec.tsx
+++ b/packages/core/components/InlineTextField/__tests__/index.spec.tsx
@@ -77,4 +77,42 @@ describe("InlineTextField", () => {
     const span = container.querySelector("span");
     expect(span?.textContent).toBe("Hello world");
   });
+
+  it("normalizes empty state and shows placeholder even when filler <br> is retained", () => {
+    const { container, rerender } = render(
+      <InlineTextField
+        propPath="subtitle"
+        componentId="comp-1"
+        value="Initial"
+        isReadOnly={false}
+        placeholder="Enter a subtitle"
+      />
+    );
+
+    const span = container.querySelector("span");
+    expect(span).not.toBeNull();
+
+    // Simulate browser retaining a <br> filler when field is emptied by user
+    if (span) {
+      span.innerHTML = "<br>";
+    }
+
+    // Re-render with empty value to trigger normalization
+    rerender(
+      <InlineTextField
+        propPath="subtitle"
+        componentId="comp-1"
+        value=""
+        isReadOnly={false}
+        placeholder="Enter a subtitle"
+      />
+    );
+
+    // After normalization, the <br> should be removed and element should be empty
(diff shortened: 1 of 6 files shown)


packages/core/components/InlineTextField/__tests__/index.spec.tsx
```diff
@@ -30,19 +30,21 @@ jest.mock("../../../lib/overlay-portal", () => ({
...
         isReadOnly={false}
+        placeholder="Enter a subtitle"
       />
     );
 
     const span = container.querySelector("span");
     expect(span).not.toBeNull();
     expect(span?.innerText ?? span?.textContent ?? "").toBe("");
+    expect(span).toHaveAttribute("data-placeholder", "Enter a subtitle");
   });
```
FedericoBonel, line 46: Can you leave this test as it was? It's actually testing other behavior and is making sure that if you don't have any placeholder you don't render falsey values.

You can add an additional test for the new placeholder behavior.

packages/core/components/InlineTextField/index.tsx
```diff
@@ -47,7 +49,10 @@ const InlineTextFieldInternal = ({
       // args, so passing null/undefined renders the literal text "null"/"undefined".
       const safeValue = value ?? "";
 
-      if (safeValue !== ref.current.innerText) {
+      // Normalize empty contenteditable spans to remove filler <br> nodes that
+      // browsers insert for editability (which prevent :empty from matching).
+      // Always replaceChildren when empty, not just when innerText differs.
+      if (!safeValue || safeValue !== ref.current.innerText) {
```
FedericoBonel:
```suggestion
      // Replace when the user has edited the value or when the value is empty 
      // to remove any elements inserted by the browser that would break placeholder styling (e.g. `<br>`).
      if (!safeValue || safeValue !== ref.current.innerText) {
```

FedericoBonel (changes requested):
Hey @ujwalbholan! Thanks for the review.

Left a couple comments.

packages/core/components/InlineTextField/__tests__/index.spec.tsx
ujwalbholan: Makes sense — I split them so the null/undefined check stays as its own regression, and the placeholder behavior has its own dedicated test.

packages/core/components/InlineTextField/index.tsx
ujwalbholan: I narrowed it to the actual behavior: we replace the content whenever the value changes or is empty so browser-added filler nodes like <br> don’t break the placeholder styling.

ujwalbholan:
@FedericoBonel Thanks, I’ve made the requested changes and kept the tests separated by behavior. I also added the browser filler-node regression case. Could you take another look when you have a moment?

FedericoBonel (approved):

PR #1832, opened 2026-09-07
fix: Selected Item Gets Unselected When Switching Viewports

### LINKED ISSUE #1170: Selected Item Gets Unselected When Switching Viewports
## Description

When switching between viewports using the provided viewport buttons in the Puck editor, the currently selected item in the outline/component tree gets unselected. This breaks the user's workflow as they lose track of which component they were working on when switching between different viewport sizes.

## Environment

- Puck version: 0.19.1
- Browser version: Chrome 135 (desktop)
- Additional environment info: TypeScript 5.x, React 18, Remix application

## Steps to reproduce

1. Load the Puck editor with some content and multiple components
2. Select any component in the outline or on the canvas (component should appear as selected with visual indicators)
3. Click on any viewport button (e.g., switch from desktop to mobile viewport)
4. Observe the outline and selected component state

## What happens

When switching viewports, the selected component gets unselected. The component that was previously highlighted/selected in the outline loses its selected state, and no component appears as selected anymore. The user has to re-select the component they were working on.

## What I expect to happen

The selected component should remain selected when switching between viewports. The selection state should persist across viewport changes, maintaining the user's context and allowing them to continue editing the same component across different viewport sizes.

## Additional Media

Video demonstration of the issue will be attached showing the component getting unselected when viewport buttons are clicked.

https://github.com/user-attachments/assets/b63de649-fbb2-4f32-a31a-0780019b6728

In the video I also click outside of the `preview` to demonstrate that any click outside of the preview deselects the item, which can shed some light about what's happening. Maybe the viewport buttons are not preventing that from happening.

DamianKocjan:
Closes #1170

## Changes made

Stop event propagation on viewport button click.

## How to test

Open puck editor, select item and select viewport using viewport buttons.

### DIFF
--- a/packages/core/components/Puck/components/Canvas/index.tsx
+++ b/packages/core/components/Puck/components/Canvas/index.tsx
@@ -247,7 +247,8 @@ export const Canvas = () => {
 
         if (
           !el.hasAttribute("data-puck-component") &&
-          !el.hasAttribute("data-puck-dropzone")
+          !el.hasAttribute("data-puck-dropzone") &&
+          !el.closest("[data-puck-viewport-controls]")
         ) {
           dispatch({
             type: "setUi",
--- a/packages/core/components/Puck/components/Preview/index.tsx
+++ b/packages/core/components/Puck/components/Preview/index.tsx
@@ -121,7 +121,8 @@ export const Preview = ({ id = "puck-preview" }: { id?: string }) => {
 
         if (
           !el.hasAttribute("data-puck-component") &&
-          !el.hasAttribute("data-puck-dropzone")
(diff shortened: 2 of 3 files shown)


Itrulia:
This unfortunately doesn't properly fix the issue. For example any element that has a ref registered with `registerOverlayPortal` causes it to lose the selection.

The only real way to solve this and to solve this properly is to remember what element last has been selected and only swap when the user selects another.

DamianKocjan:
@Itrulia Thanks for reporting that.

DamianKocjan:
@Itrulia Can you check if that issue still occurs after that fix?

PR #1835, opened 2026-09-09
chore: migrate to pnpm

xaviemirmon:
## Description

Migrates from `yarn` to `pnpm` 11.


package.json
```diff
@@ -6,22 +6,22 @@
...
-    "release:canary": "yarn release:prepare && node scripts/get-unstable-version canary | xargs yarn version:auto $1",
+    "release": "pnpm release:prepare && pnpm changelog && pnpm release-commit",
+    "release:prepare": "git fetch --tags && conventional-recommended-bump -p angular | xargs pnpm version:auto $1",
+    "release:canary": "pnpm release:prepare && node scripts/get-unstable-version canary | xargs pnpm version:auto $1",
     "release-commit": "git add -u && git commit -m \"release: v${npm_package_version}\"",
     "smoke": "node scripts/e2e/smoke.mjs",
     "test": "turbo run test",
     "version": "lerna version --force-publish -y --no-push --no-changelog --no-git-tag-version $npm_package_version",
-    "version:auto": "yarn version --no-git-tag-version --new-version $1",
+    "version:auto": "npm version --no-git-tag-version $1",
```
FedericoBonel: Is the --new-version not needed with pnpm?

FedericoBonel (approved):
Looks good! So I'm approving. Tested it locally and it seemed to work for development, building, and publishing.

FedericoBonel:
Ah actually the reason the CI is failing is because pnpm install also runs prepare scripts, which then generates dist files and the format check checks those files. So maybe we need to narrow the formatter to those files

package.json
xaviemirmon: Good catch — `--new-version` was a yarn v1 flag; both `npm version` and `pnpm version` take the new version as a positional argument, so it isn't needed here.

I've switched this to `pnpm version` for consistency with the migration. One caveat I ran into: `pnpm version` refuses to run on a dirty working tree (`ERR_PNPM_UNCLEAN_WORKING_TREE`) even with `--no-git-tag-version`, unlike `npm version`. Since `release:canary` calls `version:auto` twice (the second bump runs on a tree already dirtied by the first), I added `--no-git-checks` to preserve that behavior.

Final: `pnpm version --no-git-tag-version --no-git-checks $1` (c6a88748).

chrisvxd (approved):

Issue #522, opened 2024-06-27
Adding Card Previews on Hover


Sarfraz-droid:
Hi, are there any plans for adding Popover previews for Drawer Items on hover? If not, is there any way we can do it manually??

<img width="286" alt="Screenshot 2024-06-27 at 9 41 46 AM" src="https://github.com/measuredco/puck/assets/73013838/65ee67b2-6d1c-4a1d-912f-e7d69401f2cb">

IMythx:
i suppose u can override the component item via plugin or the override prop

chrisvxd:
What @IMythx  said - it's possible to [override](https://puckeditor.com/docs/extending-puck/custom-interfaces#ui-overrides) the [component item](https://puckeditor.com/docs/api-reference/overrides/component-item) to achieve something like this.

Issue #495, opened 2024-06-04
Feature Request: Option to call fetchList() in "external" types as you type instead of on clicking Search


Arsenalist:
Have a way invoke fetchList() so it works like an autocomplete instead of for waiting the user to hit the search button.

[Discord thread.](https://discord.com/channels/1153376562259951687/1246623927392931942)

Issue #450, opened 2024-04-17
Issue with Dispatch Function Returning Null in usePuck with Custom Button Overrides


arehmanbod:
Hi there,
Firstly, I'd like to commend the team for developing such an awesome library.
I've been working on implementing custom buttons to toggle the left and right sidebars using override.header with usePuck, following the documentation provided.
`const App = () => {
  const { dispatch, appState } = usePuck();
  return (<Puck
      dispatch={dispatch}
      ui={{
        leftSideBarVisible: true,
        viewports: {
          controlsVisible: true,
        },
      }}
      config={config}
      data={initialData}
      onPublish={save}
      overrides={{
        header: ({ children }) => {
          return (
            <div>
              <button
                onClick={() =>{
                  console.log("CALLING", dispatch, appState);
                  dispatch({
                    type: "setUi",
                    ui: {
                      rightSideBarVisible: !appState.ui.rightSideBarVisible,
                    },
                  });
                  console.log("Ending", dispatch, appState);
                }}
              >
                Right Sidebar
              </button>
              SIMPLE COMPONENTS
              <button
                onClick={() =>
                  dispatch({
                    type: "set",
                    state: {
                      data: {},
                      ui: {
                        leftSideBarVisible: !appState.ui.leftSideBarVisible,
                      },
                    },
                  })
                }
              >
                left Sidebar
              </button>
            </div>
          );
        },
    />);
}`
The problem I'm facing is that dispatch is returning as () => null, while appState is functioning correctly. Additionally, I've tried using renderHeader, but it's deprecated and only works when closing the sidebar, not when opening it.
I've searched through existing issues but haven't found a solution that meets my expectations.
Any guidance on resolving this issue would be greatly appreciated.

Thank you!
![image](https://github.com/measuredco/puck/assets/165709745/e2504b06-c0f5-4399-8a2d-0ed6d2bb926a)

PR #1841, opened 2026-09-15
build: remediate Dependabot alerts (patch runtime ws, prune redundant overrides)

xaviemirmon:
## Summary

Triage and remediation of the 51 open Dependabot alerts (#448–#522), plus an audit of the existing pnpm `overrides`.

Policy applied: **fix anything that reaches a *published* package's runtime; dismiss what is confined to private (unpublished) example/docs apps or in-repo dev/build tooling** (it cannot reach consumers of the library).

### Code changes (`pnpm-workspace.yaml` + `pnpm-lock.yaml`)
- **Fix `ws` → 8.21.3** (GHSA-96hv-2xvq-fx4p). This is the **only** advisory whose vulnerable package lands in a published package's production tree: `@puckeditor/core` pulls `ws` via `happy-dom` and `@tiptap/html` (both in core `dependencies`). Pinned through `overrides` to keep the lockfile delta to `ws` alone (a plain `pnpm update` also re-shuffled an unrelated `react` peer). Verified: all 290 `@puckeditor/core` tests pass.
- **Removed 3 redundant security overrides** — `sharp`, `undici`, `zod-validation-error`. An audit (`pnpm audit` with each pin removed) confirmed the transitive graph now resolves them to non-vulnerable versions on its own, so the pins were inert (`sharp`/`undici` produced zero lockfile change on removal). The other 11 pins still suppress real advisories and were kept.

### Dismissed alerts (50, on GitHub)
None of the remaining vulnerable packages ship in a published `@puckeditor/*` / `create-puck-app` package:
- **`not_used` (5)** — no patch exists and the vulnerable code path is not reachable:
  - `image-size` #493/#494 — resolves to 0.5.5; the advisory's JXL/HEIF/ICNS parser code doesn't exist in that version (private recipes).
  - `extract-zip` #495/#508 — dev-only via puppeteer; only unpacks the Chromium archive from Google's CDN (trusted input).
  - `dompurify` #457 — vulnerable `IN_PLACE` mode (on attacker-supplied DOM nodes) is not used; mermaid sanitizes HTML strings.
- **`tolerable_risk` (45)** — confined to the private react-router example recipes, the private docs site (mermaid/dompurify/js-yaml), or dev/build/release tooling (lerna, nx, eslint, puppeteer, tsup, sigstore). Not shipped to consumers.

### Result
`50 dismissed`, `1 open` (#450 `ws`) — which this branch's lockfile fixes, so it flips to `fixed` on the next Dependabot scan.

## Verification
- `pnpm install --frozen-lockfile` passes; package `prepare` builds succeed.
- `pnpm --filter @puckeditor/core test` → 54 suites / 290 tests pass.
- `pnpm --filter @puckeditor/core why ws --prod` → only `ws@8.21.3`.
- `pnpm audit` confirms `sharp`/`undici`/`zod-validation-error` remain clean without their pins.


FedericoBonel (approved):
Looks good I believe! Did you test SSR just to be safe? Shouldn't affect it but you never know with tiptap.

xaviemirmon:
Yup! I checked `/rsc` on the demo and it rendered the HTML as expected.

